Issue #89, opened 2026-02-15
Add pandas DataFrame integration to wingfoil-python


0-jake-0:
## Overview

Add lightweight pandas integration to enable converting stream of structs (dicts) into pandas DataFrames for analysis workflows.

## Motivation

- `.collect()` already accumulates stream values into Python lists
- KDB+ adapter demonstrates dict-based structured data patterns
- Natural next step: convert collected structured data to DataFrames for analysis
- No pandas support currently exists

## Proposed Solution

Add pure Python helper module (`pandas_helpers.py`) with two functions:

### API

```python
# Option 1: Two-step (explicit)
stream = ticker(1.0).map(lambda i: {"price": 100 + i, "qty": 10}).collect()
stream.run(cycles=10)
df = to_dataframe(stream.peek_value())

# Option 2: Convenience function
df = stream_to_dataframe(
    ticker(1.0).map(lambda i: {"price": 100 + i}),
    cycles=10
)
```

### Type Handling

- List of dicts → DataFrame with dict keys as columns (primary use case)
- List of objects with `__dict__` → Extract attributes as columns
- List of primitives → Single-column DataFrame
- Empty list → Empty DataFrame

### New Files

1. `wingfoil-python/python/wingfoil/pandas_helpers.py` - Core helpers
2. `wingfoil-python/tests/test_pandas.py` - Test coverage
3. `wingfoil-python/examples/dataframe.py` - Usage examples

### Modified Files

1. `wingfoil-python/python/wingfoil/__init__.py` - Export new functions
2. `wingfoil-python/examples/kdb.py` - Add DataFrame conversion example (optional)

## Benefits

- No Rust/PyO3 changes required
- Pandas as optional dependency
- Reuses existing `.collect()` infrastructure
- Simple and pythonic
- Easy to extend

## Verification

```bash
cd wingfoil-python
maturin develop
pytest tests/test_pandas.py -v
python examples/dataframe.py
```

See plan file: `.claude/plans/goofy-inventing-platypus.md`

PR #97, opened 2026-02-16
Feat: Add Pandas integration for stream conversion (#89)

ZohaibHassan16:
**Overview:** Implements `to_dataframe` and `stream_tod_dataframe` to allow direct conversion of stream results (dicts, objects, or primitives) into pandas DataFrame.

**Changes**
- **Logic**: New `pandas_helpers.py` module for data transformation
- **API**: Exported helpers in `__init__.py` and added `pandas` dependency.
- **Testing:** Added `tests/test_pandas.py` (4/4 passing).
- **Usage**: Added `examples/dataframe.py`

### DIFF
--- a/wingfoil-python/src/lib.rs
+++ b/wingfoil-python/src/lib.rs
@@ -4,7 +4,6 @@ mod py_kdb;
 mod py_stream;
 mod types;
 
-// Removed unused 'Graph' import
 use ::wingfoil::{Dep, Node, NodeOperators};
 use py_element::*;
 use py_stream::*;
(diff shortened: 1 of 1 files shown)


0-jake-0:
Take a look at collect here...

https://github.com/wingfoil-io/wingfoil/blob/55d69ec66f8c081b641a1b2f3c5dd0349b648222/wingfoil/src/nodes/mod.rs#L549C1-L558C6

we just need to implement similar pattern in PyStream, that collects into dataframe.

ZohaibHassan16:
I think running fmt commands may have caused this ?? lemme restore them

PR #153, opened 2026-03-28
zmq service discovery

0-jake-0:
Publishers register by name with lightweight seed nodes; subscribers discover by name without hardcoded addresses. Data flows peer-to-peer over existing PUB/SUB sockets — seeds handle only discovery traffic.

- Add zmq_seed.rs: REP-based seed node with start_seed/SeedHandle, SeedRequest/SeedResponse wire protocol (bincode), register_with_seeds, query_seeds; AtomicBool shutdown with 10ms poll cycle
- Extend ZeroMqPub trait: zmq_pub_named / zmq_pub_named_on register with seeds in start() after binding
- Add zmq_sub_discover: queries seeds then delegates to zmq_sub
- Re-export start_seed and SeedHandle from zmq module
- Add zmq_discover example
- 17 tests: 4 seed unit tests + 6 discovery integration tests + 7 existing zmq tests (all passing)

### DIFF
--- a/wingfoil/Cargo.toml
+++ b/wingfoil/Cargo.toml
@@ -153,3 +153,8 @@ required-features = ["zmq-beta"]
 name = "zmq_sub"
 path = "examples/messaging/zmq_sub.rs"
 required-features = ["zmq-beta"]
+
+[[example]]
+name = "zmq_discover"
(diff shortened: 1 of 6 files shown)


0-jake-0:
going to go different approach - etcd

PR #171, opened 2026-04-01
  Replace .unwrap() calls with proper Result error handling

### LINKED ISSUE #11: Replace .unwrap() calls with proper Result error handling
## Current State
The codebase contains `.unwrap()` calls that could panic in production.

## Goals
- Audit all `.unwrap()` usage in `/wingfoil/src/`
- Replace with proper error handling using `?` operator and `Result`
- Document any remaining `.unwrap()` calls with safety justification

## Acceptance Criteria
- [ ] Catalog all unwrap locations (use `rg "\.unwrap\(\)"`)
- [ ] Replace unwraps in critical paths (graph execution, I/O)
- [ ] Add tests for error paths
- [ ] Update CLAUDE.md if error handling patterns change

## Related
- #14 (closed) - Improved error handling interface

MavenRain:
Audit all unwrap/expect usage in wingfoil/src/ production code and replace with Result propagation via ?, ok_or_else, or explicit error handling.  Zero unwrap/expect calls remain in production code.

Closes #11

### DIFF
--- a/wingfoil/examples/dynamic/demux/main.rs
+++ b/wingfoil/examples/dynamic/demux/main.rs
@@ -42,7 +42,8 @@ fn inst_key(event: &InstEvent) -> Instrument {
 
 type PriceBook = BTreeMap<Instrument, Price>;
 
-pub fn build(period: Duration) -> (Rc<dyn Stream<PriceBook>>, Rc<dyn Node>) {
+#[allow(clippy::type_complexity)]
+pub fn build(period: Duration) -> anyhow::Result<(Rc<dyn Stream<PriceBook>>, Rc<dyn Node>)> {
     let src = self::source::market_data(period);
     let price_events = src.inst_price.map(|(i, p)| InstEvent::Price(i, p));
     let del_events = src.del_instrument.map(InstEvent::Delete);
@@ -55,7 +56,7 @@ pub fn build(period: Duration) -> (Rc<dyn Stream<PriceBook>>, Rc<dyn Node>) {
         };
         (key, de)
     });
-    let overflow_node = overflow.panic();
+    let overflow_node = overflow.panic()?;
     let processed: Vec<Rc<dyn Stream<Burst<InstEvent>>>> = slots
         .into_iter()
         .map(|slot| {
@@ -90,7 +91,7 @@ pub fn build(period: Duration) -> (Rc<dyn Stream<PriceBook>>, Rc<dyn Node>) {
             },
         );
 
-    (price_book, overflow_node)
+    Ok((price_book, overflow_node))
 }
 
 fn main() -> anyhow::Result<()> {
@@ -99,7 +100,7 @@ fn main() -> anyhow::Result<()> {
 }
 
 pub fn run(period: Duration, cycles: u32) -> anyhow::Result<()> {
-    let (price_book, overflow_node) = build(period);
+    let (price_book, overflow_node) = build(period)?;
     Graph::new(
         vec![
             price_book.logged("price book (demux)", Info).as_node(),
@@ -156,9 +157,9 @@ mod tests {
 
     /// Demux has no recycle cycles, so RunFor::Cycles(20) maps cleanly to n=1..20.
     #[test]
-    fn price_book_accumulation() {
+    fn price_book_accumulation() -> anyhow::Result<()> {
         let period = std::time::Duration::from_secs(1);
-        let (price_book, overflow_node) = build(period);
+        let (price_book, overflow_node) = build(period)?;
         let assertion = price_book.accumulate().finally(|states, _| {
             assert_eq!(states, expected_book_states());
             Ok(())
@@ -169,6 +170,5 @@ mod tests {
             RunFor::Cycles(20),
         )
         .run()
-        .unwrap();
     }
 }
--- a/wingfoil/examples/dynamic/dynamic-manual/main.rs
+++ b/wingfoil/examples/dynamic/dynamic-manual/main.rs
@@ -69,7 +69,7 @@ impl MutableNode for PriceAggregator {
                 self.inst_price
                     .filter_value(move |(inst, _)| inst == &inst_key),
             );
-            state.add_upstream(processed.clone().as_node(), true, true);
+            state.add_upstream(processed.clone().as_node(), true, true)?;
             self.per_instrument.insert(instrument, processed);
         }
         if state.ticked(self.del_instrument.clone().as_node()) {
--- a/wingfoil/examples/order_book/main.rs
+++ b/wingfoil/examples/order_book/main.rs
@@ -18,9 +18,9 @@ struct Message {
 }
 
 #[doc(hidden)]
-pub fn main() {
+pub fn main() -> anyhow::Result<()> {
     env_logger::init();
-    set_current_dir();
+    set_current_dir()?;
     let book = RefCell::new(lobster::OrderBook::default());
     // map from seconds from midnight to NanoTime time
     let get_time = |msg: &Message| NanoTime::new((msg.seconds * 1e9) as u64);
@@ -29,7 +29,7 @@ pub fn main() {
         .split();
     let prices_export = prices
         .filter_value(|price| !price.is_none())
-        .map(|price| price.unwrap())
+        .map(|price| price.unwrap_or_default())
         .distinct()
         .csv_write("prices.csv");
     let fills_export = fills.csv_write("fills.csv");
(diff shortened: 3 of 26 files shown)


0-jake-0 (commented):
Thanks for tackling this — replacing panicking `.unwrap()` calls with proper `Result` propagation is exactly the right direction for a library, and a lot of the changes here are clean and correct. The `?` propagation through `csv_read`/`csv_write`, the `graph.rs` internal fixes, and the `tokio_runtime` refactor are all solid improvements.

**Build is currently failing:** https://github.com/wingfoil-io/wingfoil/actions/runs/23856456268/job/69591191505 — worth getting that green before re-review.

---

### Needs work

A few changes swap one problem for another — silent failure is harder to debug than a panic:

1. **`TimeQueue::next_time` returns `NanoTime::MAX` on empty queue** — this is a programming error call site, not a recoverable runtime condition. The sentinel value will silently schedule callbacks far in the future. Suggest returning `Option<NanoTime>` instead.

2. **`Overflow::stream()` / `Overflow::panic()` returning `Option`** — in the test helper, if these return `None` the overflow guard is silently dropped, meaning overflow can occur with no assertion. This should be an error, not a silent no-op.

3. **CSV `filter_map` silently skips malformed rows** — this is a meaningful behavioral change: callers currently relying on fail-fast will silently get partial results. If skip-and-continue is the desired behaviour it should be an explicit opt-in (e.g. a separate `csv_read_lenient` function or a parameter), not a silent default swap.

4. **`FeedbackSink::send` logs and drops when called before setup** — the value is queued but the callback is never scheduled, so it's silently lost forever. The original panic was correct here; if `send` can't return `Result`, a `debug_assert!` or panic is safer than a log line.

5. **`bencher.rs` `builder.take()` wrapped in `if let Some`** — calling `start()` twice now silently no-ops. The original `.unwrap()` was correctly catching that misuse.

6. **`#[must_use]` removed from `csv_read`, `csv_write`, etc.** — these should keep `#[must_use]`; it's more important on `Result`-returning functions than it was before.

---

### Notes (not on critical path)

Listing these here for my own reference to break out into a separate issue — none of these are introduced by this PR and none are blockers:

- `TimeQueue::pop()` callers use `if let Some(v) = queue.pop() { self.value = v; }` which silently retains the old value on `None` — logic hole worth a dedicated fix.
- `graph_node.rs` `peek_ref` returns a silent default when the stream is uninitialized rather than surfacing the error.
- Some `start()` implementations call `add_callback`/`always_callback` without `?`, silently discarding errors.
- Errors from inner graphs in spawned threads may still be swallowed in some teardown paths.

MavenRain:
Thanks for the thorough review, Jake.  I pushed a follow-up commit (87d7800) that addresses the core items:

1. **TimeQueue::next_time** — now returns `Option<NanoTime>` instead of the `NanoTime::MAX` sentinel.
2. **Overflow::stream()/panic()** — changed from `Option` to `anyhow::Result` with descriptive error messages.
3. **CSV silent skipping** — removed the `filter_map` that was silently dropping malformed rows.  However, the current approach eagerly collects (see below).
4. **FeedbackSink::send** — reverted to `expect()` since this is a programming error, not a recoverable condition.
5. **bencher.rs** — restored `expect("Bencher::start() called twice")` with a clear message.
6. **#[must_use]** — restored on all csv_read/csv_write functions.

Still outstanding:
- The csv_iterator eager collection is a behavioral regression, as you noted in your follow-up comment.  I'll rework this to keep the iterator lazy and propagate deserialize errors per-record through `cycle()`'s `anyhow::Result<bool>` return.
- The `SimpleIteratorStream` peek/next pattern — you're right that `next()` returning `None` after `peek()` returned `Some` is an invariant violation, not a runtime error.  I'll change the `ok_or_else` to `expect("BUG: peek returned Some but next returned None")`.
- Will rebase against main once the above are sorted.

MavenRain:
Replying to [your discussion comment](https://github.com/wingfoil-io/wingfoil/pull/171#issuecomment-4187193676) — agreed on all three points:

1. **csv_iterator** — You're right, the eager collect is a regression for large backtesting files.  I'll rework it to keep the iterator lazy and surface deserialize errors through cycle()'s Result return type rather than collecting upfront.

2. **bencher.rs** — Already restored the `expect()` in the follow-up commit (87d7800).  Agree it's the right tool for a double-call bug.

3. **SimpleIteratorStream** — Good call.  I'll change it to `expect("BUG: peek returned Some but next returned None")`.

Will rebase against main once the above are sorted.  Thanks for the kind words and the detailed feedback, Jake.

0-jake-0:
Hey @MavenRain — great progress on the follow-up commit! You've cleanly addressed all six items from the first review: `TimeQueue::next_time` returning `Option`, `Overflow` returning `Result`, the CSV `filter_map` removal, the `FeedbackSink`/`bencher` `expect()` restorations, and the `#[must_use]` attributes. The error messages throughout are clear and contextual — really nice work.

Two items you mentioned are still in the pipeline:

**1. `csv_iterator` eager collection** — the `.collect::<anyhow::Result<Vec<_>>>()?` is still there. As discussed, this loads the entire CSV into memory at construction time, which is a regression for large backtesting files. Still needs the lazy-iterator-with-per-record-error-propagation rework you described.

**2. `SimpleIteratorStream` peek/next** — still uses `ok_or_else(|| anyhow!(...))`. Since `next()` returning `None` after `peek()` returned `Some` is an invariant violation (not a runtime error), this should be `expect("BUG: peek returned Some but next returned None")` as you mentioned.

A couple of new things I noticed in the latest commit:

**3. `time.rs` `NanoTime → NaiveDateTime` fallback to `UNIX_EPOCH`** (line 120) — silently mapping an invalid timestamp to 1970-01-01 could cause confusing downstream behaviour. This is the same category as the silent-failure patterns from round 1. Suggest `expect("BUG: NanoTime contains invalid timestamp")` here, or at minimum log a warning.

**4. `bencher.rs` graph error handling** (line 77) — the `if let Err(e) = graph.run() { error!(...) }` pattern means the thread exits cleanly without setting `Signal::End`, so the caller could hang in `wait()`. The original panic at least surfaced through the `JoinHandle`. Consider either re-panicking after logging, or setting the signal in a scope guard so it fires regardless of how the thread exits.

Once those are sorted, just needs a rebase against main (`0c5dc07`) — there's been some iceoryx2 work merged since, but it shouldn't overlap with your changes much. Let me know if you hit anything tricky!

---

MavenRain:
I rebased onto current main (6b872b0) and pushed the round-2 fixes:         
  
   1. csv_iterator is now lazy — yields anyhow::Result<ValueAt<T>> so        
   per-record deserialize errors propagate through the stream instead of 
   being collected upfront. csv_read returns anyhow::Result for the eager    
   File::open failure.
   2. SimpleIteratorStream peek/next mismatch is now expect("BUG: peek 
   returned Some but next returned None") — invariant violation, not a       
   runtime error.
   3. time.rs NanoTime → NaiveDateTime invalid-timestamp fallback is now     
   expect("BUG: NanoTime contains invalid timestamp") rather than silently   
   mapping to UNIX_EPOCH.
   4. bencher.rs thread now uses a SignalEndOnDrop RAII guard so Signal::End 
   fires on every exit path (graph error, panic, kill), preventing callers   
   from spinning in step().

0-jake-0:
Hey @MavenRain — first off, thanks so much for sticking with this through three rounds of review. The shape of this PR is genuinely great work, and the round-2 fixes (lazy CSV iterator, `SignalEndOnDrop`, BUG-style `expect()` for invariants) all landed cleanly. Really appreciate the perseverance.

I reproduced the CI build failure locally — it's all call sites behind feature flags that aren't on by default, so a local `cargo build` wouldn't catch them. CI's `cargo clippy --workspace --all-targets --all-features` does. **If you can patch these, I'm happy to merge and clean up any remaining nits myself afterwards.**

### Hard build error (`--features full`)

**`wingfoil/src/adapters/fix/mod.rs:1049`** — `ready_notifier()` now returns `Result`:
```rust
chan_sender.set_notifier(state.ready_notifier()?);
//                                              ^ add `?`
```

### Unresolved-method errors (`csv` feature)

**`wingfoil/src/adapters/csv/write.rs:83` and `:94`** — `csv_read` now returns `Result`, so the method-chain breaks. Either:
```rust
csv_read("…", get_time, false)
    .unwrap()
    .csv_write("…")
```
or thread `?` through the test if you make it return `anyhow::Result<()>`.

### `unused_must_use` warnings (= clippy errors with `-D warnings`)

Same fix as the others — append `?` (these all sit inside `MutableNode::cycle`/`start`/`setup` which already return `Result`):

- `wingfoil/src/adapters/fix/mod.rs:659, 749` — `state.always_callback()?;`
- `wingfoil/src/adapters/iceoryx2/read.rs:380, 408, 568, 595` — `state.always_callback()?;`
- `wingfoil/src/adapters/iceoryx2/integration_tests.rs:60` — `state.always_callback()?;`

### Quick way to verify locally

```bash
sudo apt-get install -y protobuf-compiler   # for etcd-client
cargo clippy --workspace --all-targets --all-features --exclude wingfoil-python -- -D warnings
```

That's it — once those compile, I'll pull it in. Thanks again for the careful, systematic work on this one.

---

0-jake-0:
Also @MavenRain — quick apology: the build failures you hit were a documentation gap on our side, not anything you missed. Our `CLAUDE.md` / `CONTRIBUTING.md` were telling contributors to run only the default-features clippy, which silently skips exactly the feature-gated code (`fix`, `iceoryx2-beta`, `csv`) that broke for you. That was on us.

Cleaned up in #270 — there are now `cargo lint` / `cargo lint-all` aliases that mirror CI exactly, with the `protoc` prerequisite documented. Sorry for the round-trip on this one — really appreciate you sticking with it.

---

PR #402, opened 2026-06-09
fix(adapters): iceoryx2 spin real-time guard + fluvio test marker

0-jake-0:
Part 1 of 4 from an adapter compliance review against the `new-adapter` skill. These are the two **high-severity** behavioral findings.

## 1. iceoryx2 spin mode does not reject non-real-time runs

`Iceoryx2ReceiverNode::start()` called `state.always_callback()` in `Spin` mode without checking the run mode. A `HistoricalFrom` (backtest) run would therefore busy-spin against a live IPC socket instead of failing fast. The skill's spin-loop pattern (step 7) requires a `run_mode() != RealTime` bail guard, which the FIX spin source already has.

Added the guard at both `Spin` `start()` sites in `iceoryx2/read.rs` (typed + slice subscribers):

```rust
Iceoryx2Mode::Spin => {
    if state.run_mode() != crate::RunMode::RealTime {
        anyhow::bail!("iceoryx2 spin mode only supports real-time");
    }
    state.always_callback();
    ...
```

## 2. fluvio Python tests silently skip

`test_fluvio.py` gated its live-cluster suites with `@unittest.skipUnless(FLUVIO_AVAILABLE, ...)` driven by a TCP probe — the exact "never silently skip" anti-pattern the skill forbids (step 13.f). With no `requires_fluvio` marker registered, the default `pytest` run could be falsely green against a down cluster.

- Replaced both `skipUnless` decorators with `@pytest.mark.requires_fluvio`.
- Removed the `fluvio_available()` TCP probe and the now-unused `socket`/`urllib` imports.
- Registered the `requires_fluvio` marker in `pyproject.toml` and added it to the default `addopts` deselect list, so live tests fail loudly under `-m requires_fluvio` and are skipped by default.

The construction / unreachable unit suites are unchanged — they still run by default with no marker.

## Verification

- `cargo check -p wingfoil --features iceoryx2` passes.
- `cargo fmt --all` clean.
- `python -m py_compile tests/test_fluvio.py` passes; no stale probe references remain; marker registered in both `markers` and `addopts`.

https://claude.ai/code/session_01JLqvwSW8V4xPpvxV1vwXuX

---
_Generated by [Claude Code](https://claude.ai/code/session_01JLqvwSW8V4xPpvxV1vwXuX)_

### DIFF
--- a/wingfoil-python/pyproject.toml
+++ b/wingfoil-python/pyproject.toml
@@ -51,6 +51,7 @@ markers = [
     "requires_iceoryx2: needs wingfoil built with --features iceoryx2",
     "requires_web: needs a wingfoil web server (runs in-process)",
     "requires_kafka: needs a Kafka-compatible broker on localhost:9092",
+    "requires_fluvio: needs a Fluvio cluster on localhost:9003",
     "requires_aeron: needs an Aeron media driver (and wingfoil built with --features aeron)",
 ]
-addopts = "-m 'not requires_etcd and not requires_kdb and not requires_otel and not requires_iceoryx2 and not requires_web and not requires_kafka and not requires_aeron'"
+addopts = "-m 'not requires_etcd and not requires_kdb and not requires_otel and not requires_iceoryx2 and not requires_web and not requires_kafka and not requires_fluvio and not requires_aeron'"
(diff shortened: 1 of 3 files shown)


0-jake-0:
Closing as superseded by the `next` → `main` cutover, which deletes the legacy engine this PR targets.

The iceoryx2 non-real-time guard is a design property of the `next` adapter rather than a patch: sources take a `GraphBuilder` + `RunMode` and reject `RunMode::HistoricalFrom` **at wiring time** — see `crates/wingfoil/src/adapters/iceoryx2/mod.rs:84`. More broadly, the inconsistent-realtime-guard problem this PR chipped at is tracked repo-wide as #455, and every `next` adapter now consults `RunMode`.

---

PR #415, opened 2026-07-08
Add kdb_sub real-time tickerplant subscription; simplify kdb read signatures

0-jake-0:
## Summary

Adds a real-time KDB+ read path and simplifies the historical read signatures.
Scope is **kdb only** — the PostgreSQL adapter lives in #411.

## `kdb_sub` — real-time tickerplant subscription

The live counterpart to `kdb_read` (bounded historical replay). It subscribes to a
q **tickerplant** via `.u.sub[`table;syms]` and streams published rows on-graph as
they arrive, decoded by the same `KdbDeserialize` impls `kdb_read` uses.

Unlike the postgres `LISTEN`/`NOTIFY` model — where the notification is only a
wake-up and rows are re-queried — a tickerplant is genuinely push-based: the
`(`upd; table; data)` message carries the rows themselves, so there is **no
re-query and no cursor**. The `upd` payload (a table or a bare list of column
vectors) is normalised by a new `upd_payload_rows` helper in `read.rs` and fed
through the existing `Rows`/`from_kdb_row` decode path.

- Requires `RunMode::RealTime` (bails otherwise); `kdb_read` stays historical.
- Captures column names from the synchronous `.u.sub` reply schema.
- Ignores non-`upd` control messages (e.g. end-of-day `.u.end`).
- Tails from subscription time; does **not** replay the tickerplant log / RDB buffer.
- Unit test covers the `RunMode::RealTime` guard (no live server needed).

## Read-signature simplification

Factors the query-closure generic `F` out of `kdb_read` and `kdb_read_cached`: it is
now an argument-position `impl FnMut(...)`, so callers write `kdb_read::<Trade>(…)`
instead of `kdb_read::<Trade, _>(…)`, matching the `csv_read` style. All kdb call
sites, examples, and docs updated.

## Testing note

The subscribe→receive→decode loop is **not** exercised against a live tickerplant
(none available in CI here); the wire path is covered by reuse of the existing,
tested `Rows`/`from_kdb_row` machinery, and the run-mode guard is unit-tested. A
`kdb-integration-test` against a real `tick.q` (plus an example and Python binding,
matching how `postgres_sub` shipped) is a sensible follow-up.

## Verification

- `cargo clippy` (kdb + integration-test features, `-D warnings`) — clean
- `cargo fmt --all --check` — clean
- 241 kdb lib unit tests pass (incl. the new guard test)
- `wingfoil-python` compiles

https://claude.ai/code/session_014Cj5Jt9YcxAqSCPKuAVuDk

---
_Generated by [Claude Code](https://claude.ai/code/session_014Cj5Jt9YcxAqSCPKuAVuDk)_

### DIFF
--- a/wingfoil-python/src/py_kdb.rs
+++ b/wingfoil-python/src/py_kdb.rs
@@ -143,7 +143,7 @@ pub fn py_kdb_read(
     chunk_size: u64,
 ) -> PyStream {
     let conn = KdbConnection::new(host, port);
-    let stream: Rc<dyn Stream<Burst<PyKdbRow>>> = kdb_read::<PyKdbRow, _>(
+    let stream: Rc<dyn Stream<Burst<PyKdbRow>>> = kdb_read::<PyKdbRow>(
         conn,
         std::time::Duration::from_secs(chunk_size),
         move |(t0, t1), _date, _iter| {
--- a/wingfoil/examples/README.md
+++ b/wingfoil/examples/README.md
@@ -46,7 +46,7 @@ struct Price {
     mid: f64,
 }
 
-kdb_read::<Price, _>(
+kdb_read::<Price>(
     conn,
     Duration::from_secs(10),
     |(t0, t1), _date, _iter| {
--- a/wingfoil/examples/kdb/read/README.md
+++ b/wingfoil/examples/kdb/read/README.md
@@ -49,7 +49,7 @@ impl KdbDeserialize for Price {
     }
 }
 
-kdb_read::<Price, _>(
+kdb_read::<Price>(
     conn,
     Duration::from_secs(10),
     |(t0, t1), _date, _iter| {
--- a/wingfoil/examples/kdb/read/main.rs
+++ b/wingfoil/examples/kdb/read/main.rs
@@ -35,7 +35,7 @@ fn main() -> Result<()> {
     let conn = KdbConnection::new("localhost", 5000);
     let chunk = Duration::from_secs(10);
 
-    kdb_read::<Price, _>(
+    kdb_read::<Price>(
         conn,
         chunk,
         |(t0, t1), _date, _iter| {
--- a/wingfoil/examples/kdb/read_cached/README.md
+++ b/wingfoil/examples/kdb/read_cached/README.md
@@ -59,7 +59,7 @@ impl KdbDeserialize for Price {
 // Use u64::MAX for an unbounded cache.
 let config = CacheConfig::new("/tmp/wingfoil-kdb-cache", 10 * 1024 * 1024);
 
-kdb_read_cached::<Price, _>(
+kdb_read_cached::<Price>(
     conn,
     Duration::from_secs(10),
     config,
--- a/wingfoil/examples/kdb/read_cached/main.rs
+++ b/wingfoil/examples/kdb/read_cached/main.rs
@@ -32,7 +32,7 @@ impl KdbDeserialize for Price {
 }
 
 fn run(conn: KdbConnection, config: CacheConfig) -> Result<()> {
-    kdb_read_cached::<Price, _>(
+    kdb_read_cached::<Price>(
         conn,
         Duration::from_secs(10),
         config,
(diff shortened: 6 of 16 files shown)


0-jake-0 (commented):
## Review — `kdb_sub` + read-signature cleanup

Reviewed the full diff plus the surrounding machinery (`produce_async`/`to_message_stream`, `read.rs` `Rows`/`KdbExt`, and module visibility). Overall this is clean, well-documented, and correctly scoped to kdb-only. No blocking issues. A few observations below.

### What's good

- **Read-signature simplification is a safe, mechanical win.** Moving `F` from a generic param to argument-position `impl FnMut(...)` is purely a call-site ergonomics change; every call site, example, and doc is updated consistently (`kdb_read::<T, _>` → `kdb_read::<T>`), and the `Send + 'static` bounds are preserved verbatim, so no behavioural change.
- **The real-time claim checks out against the engine.** I confirmed in `async_io.rs::to_message_stream` that `RunMode::RealTime` yields `Message::RealtimeValue(value)` and stamps graph time with `NanoTime::now()`, discarding the `NanoTime` returned by `from_kdb_row`. So the doc statement ("the timestamp column ... is carried through but not used to order the real-time stream") is accurate, and yielding the row's embedded timestamp is harmless.
- **Reuse of the tested decode path is the right call.** `upd_payload_rows` → `Rows` → `from_kdb_row` funnels tickerplant rows through the same machinery `kdb_read` uses; `Rows::from_column_list` is a minimal, sensible addition. Visibility (`pub(super)` in `read`, consumed from sibling `sub`) resolves correctly.
- **The run-mode guard test is valid** — the `Err` returned by the producer closure surfaces via `Message::Error` and fails `run()`, matching the existing `produce_async_mid_stream_error` pattern.

### Observations (non-blocking)

1. **Silent column-name capture** — `sub.rs:100-104`. If the `.u.sub` reply doesn't parse as expected, `columns` falls back to `unwrap_or_default()` (empty `Vec`). That's fine for the positional decoders the crate documents (they ignore `columns`), but a `KdbDeserialize` impl that looks columns up *by name* would silently misbehave with no diagnostic. Consider a `log::warn!` on the fallback so a schema-parse failure isn't invisible.

2. **Single-row `upd` payloads sent as atoms** — `read.rs` `upd_payload_rows`/`from_column_list`. Standard `tick.q` enlists columns so each is a vector, and this code assumes that (`K::len()` on the first column → row count; `Row::get` → `element_at(index)`). A feed handler that publishes a single row as *atoms* rather than 1-element vectors would give `n_rows` based on `K::len()` of an atom and then error in `element_at` (non-list → `InvalidOperation`), terminating the stream. This is a known kdb+tick nuance and matches the PR's "not integration-tested against a live tickerplant" caveat — worth a one-line note in the doc comment as a known limitation.

3. **No table-name filter on `upd` messages** — `sub.rs`. Only `element_at(0) == `upd`` is checked; `element_at(1)` (the table name) is ignored. Correct for the single-table, single-connection subscription here, since the TP only pushes the subscribed table. It's a latent assumption if the connection is ever reused for multiple subscriptions — a defensive `element_at(1) == table` check would harden it, but it's optional given current usage.

4. **Doc nesting regression** — `kdb/CLAUDE.md`. The new `kdb_sub()` bullet pushed the `KdbDeserialize trait` line down a level so it now reads as a *sub-bullet of `kdb_sub`* rather than a top-level component. Minor, but it misrepresents the structure — de-indent that line back to the top level.

5. **Nit: redundant re-bindings** — `sub.rs:66-68` (`let connection = connection;` etc.). These don't change capture semantics for the following `async move` and could be dropped (only `let run_mode = ctx.run_mode;` extracts anything). Harmless; flag only if `clippy::redundant_locals` complains in CI.

### Scope note

No example, Python binding, or `kdb-integration-test` for `kdb_sub` — the PR explicitly defers these as follow-ups (mirroring how `postgres_sub` shipped). Reasonable given no tickerplant in CI, but it does mean the subscribe→receive→decode wire path is unexercised end-to-end; the follow-up integration test is the real confidence gate here.


---

0-jake-0:
Superseded by #417, which contains this branch's `kdb_sub` + read-signature work plus follow-ups (kdb_read out-of-window row fix, integration tests for `kdb_sub` and the fix, and docs). Closing in favour of #417.

---

PR #475, opened 2026-07-22
feat(next-macros): add stateless single-input ops to the compiled path

0-jake-0:
## What

The interpreted engine + fluent API support ~30 ops, but the `graph!` /
`compiled()` / `nested()` path supported only 13. This PR brings the
**simpler stateless / single-input** ops into the compiled path:

| op | source | notes |
|----|--------|-------|
| `distinct` | `ops.rs` `Distinct<T>` | `State = Option<T>` |
| `difference` | `Difference<T>` | `State = Option<T>` |
| `limit` | `Limit<T>` | `Cfg = u32`, `State = u32` |
| `inspect` | `Inspect<A, F>` | closure config (like `map`) |
| `map_filter` | `MapFilter<A, B, F>` | closure returns `(B, bool)` |
| `with_time` | `WithTime<T>` | non-Default slot seed — see below |
| `ticked_at` | `TickedAt<T>` | emits `NanoTime` |
| `ticked_at_elapsed` | `TickedAtElapsed<T>` | emits `NanoTime` |

Deferred to a follow-up PR (out of scope, as agreed): scheduling/timer ops
(`throttle`, `window`, `buffer`) and the `try_*` / 3-ary variants
(`try_map`, `try_join`, `join3`, `try_join3`, etc.).

## How

For each op, the four macro pieces are added, mirroring an existing template
(`Map` for the stateless closure ops, `RollingSum`/`Fold` for the
stateful-single-input ops):

1. an `OpKind` variant,
2. an `info()` `OpInfo` row with **every field spelled out** (no `..base`
   struct-update — an omitted dispatch flag is a compile error, not a silent
   wrong default),
3. a parse arm in `apply_call`,
4. the op's `CfgInit` / `StateInit` / `ValueSeed`.

### `with_time` slot seeding (the one correctness subtlety)

`interp.rs` seeds the `with_time` slot with `(NanoTime::ZERO, src.clone())`,
**not** `Default::default()`. `NanoTime::ZERO == NanoTime::default()`, so the
time component always matches — but the value component drifts when the
upstream is a `fold` with a **non-default init**: interp shows
`(ZERO, init)` while a `Default`-seeded compiled slot would show
`(ZERO, T::default())` on a pre-first-tick read.

Fixed by adding `ValueSeed::WithTimeUpstream`, which seeds the compiled slot
identically from the upstream's already-declared `__v_<up>` local (the graph
def is now threaded into `node_decl` to resolve it). For an `Input` upstream
in the `nested` path — where `__v_<input>` is not in scope at setup — it falls
back to `Default`; the island's *observable* pre-tick value is the outer
graph's `Default` composite-slot seed regardless.

All other new ops seed via `register_op1` with `Out::default()`, so they use
`ValueSeed::Default`.

## Tests — 3-engine parity (interpreted == compiled == nested)

New `tests/compiled_stateless_ops.rs` runs a source-island micro-graph per op
through all three expansions and asserts they agree exactly (plus exact
expected values). Includes the seeding/timing case that motivated
`WithTimeUpstream`: a `with_time`-over-`fold(100, …)` read via a **passive
`sample` edge before it first ticks** — verified to fail the interpreted-vs-
compiled assertion if the slot is Default-seeded. The existing catalog parity
tests (values **and** tick times) still pass, and the `unknown_combinator`
trybuild fixture was updated for the expanded combinator list.

## Verification

- `cargo fmt --all` — clean
- `cargo test -p wingfoil-next -p wingfoil-next-macros` — all pass (default and `--features async`)
- `cargo clippy -p wingfoil-next -p wingfoil-next-macros --all-targets -- -D warnings` — clean (also with `--features async`)

---

### DIFF
--- a/wingfoil-next-macros/src/lib.rs
+++ b/wingfoil-next-macros/src/lib.rs
@@ -93,7 +93,10 @@
 //! `.filter(&cond)`, `.fold(init, f)`, `.sample(&trigger)`, `.merge(&other)`,
 //! `.join(&other, f)`, `.delay(duration)`; statistics (on `f64` streams)
 //! `.ewma(decay)` / `.ewma_per_tick(alpha)` / `.ewma_half_life(dur)`,
-//! `.rolling_sum(window)`, `.rolling_mean(window)`; sugar `.count()` and
+//! `.rolling_sum(window)`, `.rolling_mean(window)`; stateless / single-input
+//! `.distinct()`, `.difference()`, `.limit(n)`, `.inspect(f)`,
+//! `.map_filter(f)`, `.with_time()`, `.ticked_at()`, `.ticked_at_elapsed()`;
+//! sugar `.count()` and
 //! `.accumulate()`; and bounded repetition `.map_n(N, f)` (chain `map` N
 //! times) / `.fan(N, |s| <sub-chain>)` (N parallel copies of a sub-chain,
 //! merged). `map_n` / `fan` counts must be integer literals so the unrolled
@@ -144,6 +147,15 @@ enum OpKind {
     Ewma,
     RollingSum,
     RollingMean,
+    // Stateless / single-input ops brought into the compiled path.
+    Distinct,
+    Difference,
+    Limit,
+    Inspect,
+    MapFilter,
+    WithTime,
+    TickedAt,
+    TickedAtElapsed,
 }
 
 /// The shape of an op's `Op::In` tuple — how the cycle call assembles its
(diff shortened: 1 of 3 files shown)


0-jake-0:
Superseded by #496. That change deleted the `graph!`/`compiled()` `OpKind`/`OpInfo` table and made every `#[op]`-annotated op reach `compiled()`/`nested()` through a generic forwarder fallback (benchmarked at ~1.01× a hand-written table row). Every op this PR added a table row for — `distinct`, `difference`, `limit`, `inspect`, `map_filter`, `with_time`, `ticked_at`(`_elapsed`) — carries `#[op]` and now reaches the compiled path for free, so these rows no longer apply. Closing as superseded.

---

PR #477, opened 2026-07-22
feat(next-macros): add stateful, timer, and fallible ops to the compiled path

0-jake-0:
## What

PR2 of the compiled-path completion (follows #475, but branched fresh from
`next` — not stacked). Brings the remaining stateful / scheduling-timer and
multi-input / fallible partial(macro) ops into the `graph!` / `compiled()` /
`nested()` path:

| op | source | shape |
|----|--------|-------|
| `throttle` | `Throttle<T>` | `Cfg = NanoTime`, `State = Option<NanoTime>` |
| `window` | `Window<T>` | `Cfg = NanoTime`, `State = WindowState<T>`, has `start` |
| `buffer` | `Buffer<T>` | `Cfg = usize`, `State = Vec<T>` |
| `join3` | `Join3<…>` | 3 active inputs, closure config |
| `try_map` | `TryMap<…>` | fallible, 1 input |
| `try_join` | `TryJoin<…>` | fallible, 2 active inputs |
| `try_join3` | `TryJoin3<…>` | fallible, 3 active inputs |

Left for a follow-up (per the plan): the `rolling_min`/`max`/`var`/`std`/`median`
macro rows (mechanical, analogous to the existing `rolling_sum`/`mean` rows).

## How

For each op, the four macro pieces are added, every `OpInfo` field spelled out
(no `..base`). Added an `Inputs::Three` shape (`(&a, &b, &c)`) for the 3-ary
joins.

### Key design points

- **The timer ops are input-activated (`Activation::NONE`), not
  self-scheduling.** `throttle` / `window` / `buffer` read `ctx.time()` /
  `ctx.is_last_cycle()` but never call `ctx.schedule()`, so there is no dynamic
  wake for the straight-line compiled schedule to represent — they emit
  cleanly. **No op resisted compiled emission.**
- **`window` keeps a `start` hook** (`state_in_start`) that seeds its first
  boundary; wired through `node_start` in both the compiled and nested paths.
- **All seven ops seed their value slot `Out::default()`** (matching
  `interp.rs` `new_slot`), so `ValueSeed::Default` is correct — no PR1-style
  custom seed needed.
- **`is_last_cycle` flush is an outer-run boundary.** `window` / `buffer` flush
  pending contents on the final cycle via `Ctx::is_last_cycle`. That signal is
  intentionally **not** propagated into a nested island (`Ctx::nested`
  hard-codes it `false` — a pre-existing, documented island limitation), so an
  island omits the run-end flush.
- **`try_*` need no special handling** beyond mirroring their infallible
  counterparts: the emitted cycle already threads `?`, so a returned `Err`
  aborts the run on every engine.

## Tests — `compiled_stateful_ops.rs`, 3-engine parity

- `interpreted == compiled == nested` for every op.
- **Timing asserted via count-as-time proxies:** every source is
  `ticker(10ns).count()`, so count `N` occurs at engine time `(N-1)·10ns` — a
  `window` flushing `[1,2,3]` proves the 30ns boundary fired at t=30, a
  `throttle` passing `4` proves the gate reopened at t=30. Drift in any
  boundary/gate computation regroups these values, so the exact-value
  assertions are timing assertions.
- **`is_last_cycle` flush:** the "clean" timer tests are sized to end on a
  natural flush boundary (where `is_last_cycle` is a no-op and all three agree);
  dedicated `window_last_cycle_flush` / `buffer_last_cycle_flush` tests assert
  the run-end flush on `interpreted == compiled` **and** assert `nested`
  deliberately omits it.
- **Fallible propagation:** a `try_map` returning `Err` aborts interpreted,
  compiled, and nested alike.

Updated the `unknown_combinator` trybuild `.stderr` for the expanded list.

## Merge note

Additively edits the same `OpKind` enum / `info()` / parse-arm / combinator-list
regions as #475; any textual conflict is trivial to resolve at merge time.

## Verification

- `cargo fmt --all` — clean
- `cargo test -p wingfoil-next -p wingfoil-next-macros` — all pass (default **and** `--features async`)
- `cargo clippy -p wingfoil-next -p wingfoil-next-macros --all-targets -- -D warnings` — clean (default and `--features async`)

---

### DIFF
--- a/wingfoil-next/src/interp.rs
+++ b/wingfoil-next/src/interp.rs
@@ -744,7 +744,7 @@ impl Builder {
         let idx = self.nodes.len();
         let src_slot = self.slot(src);
         let out = self.new_slot(T::default());
-        let cs = Self::cell(NanoTime::from(interval), None::<NanoTime>);
+        let cs = Self::cell(interval, None::<NanoTime>);
         self.push_node(
             vec![src.idx],
             Throttle::<T>::ACTIVATION,
@@ -782,7 +782,7 @@ impl Builder {
         let idx = self.nodes.len();
         let src_slot = self.slot(src);
         let out = self.new_slot(Vec::<T>::new());
-        let cs = Self::cell(NanoTime::from(interval), WindowState::<T>::default());
+        let cs = Self::cell(interval, WindowState::<T>::default());
         let cs2 = cs.clone();
         self.push_node(
             vec![src.idx],
--- a/wingfoil-next/src/ops.rs
+++ b/wingfoil-next/src/ops.rs
@@ -257,27 +257,30 @@ impl<T: Clone + 'static> Op for Limit<T> {
 }
 
 /// Rate-limits: emits the first value, then suppresses until at least
-/// `interval` has passed since the last emit. `Cfg` = interval, `State` =
-/// last emit time.
+/// `interval` has passed since the last emit. `Cfg` = the interval as passed
+/// at the call site (`Duration`, per the uniform arg-is-the-config
+/// convention; converted to engine time in `cycle`), `State` = last emit time.
 pub struct Throttle<T>(PhantomData<T>);
 
+#[op(build = throttle, no_builder)]
 impl<T: Clone + 'static> Op for Throttle<T> {
-    type Cfg = NanoTime;
+    type Cfg = Duration;
     type State = Option<NanoTime>;
     type In<'a> = (&'a T,);
     type Out = T;
     const ACTIVATION: Activation = Activation::NONE;
 
     fn cycle(
-        cfg: &mut NanoTime,
+        cfg: &mut Duration,
         state: &mut Option<NanoTime>,
         input: (&T,),
         ctx: &mut Ctx<'_>,
     ) -> Result<Tick<T>> {
+        let interval = NanoTime::from(*cfg);
         let now = ctx.time();
         let emit = match *state {
             None => true,
-            Some(last) => now - last >= *cfg,
+            Some(last) => now - last >= interval,
         };
         Ok(if emit {
             *state = Some(now);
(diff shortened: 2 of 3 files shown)


0-jake-0:
## Reworked onto the `#[op]` mechanism (#496 superseded the original approach)

The original commit added `OpKind`/`OpInfo` **table rows** to the `graph!` macro. `next` merged #496, which **deleted that entire table** — ops now reach `compiled()`/`nested()` purely by carrying `#[op(build = name)]`, which emits the naming-convention forwarders (`__wf_op_<name>_*`) + the `__WF_OP_<NAME>_ACTIVATION` const that the macro's generic fallback dispatches through. Multi-input works via #496's argument-classification convention (a `&stream` arg is an edge).

This branch has been **reset to `origin/next`** (dropping the dead table-row commit, which targeted deleted code) and re-authored as a single commit under the new mechanism.

### What changed
Annotated the remaining single-graph catalog ops that lacked `#[op]`:

- **`throttle`, `window`** — single-input timer ops (`ACTIVATION::NONE`; they read `ctx.time()`/`is_last_cycle()` but never self-schedule). Their `Cfg` was `NanoTime`, but the call site passes a `Duration` and `compiled()` re-emits that arg **as** the `Cfg`, so `Cfg` had to become `Duration` (converted internally, matching the `Ticker`/`Delay` convention). `window` also exercises `#[op]`'s `start`-hook forwarding. Both keep their hand-written interpreted `Builder` methods (`no_builder`) — `window`'s needs the `start` hook, which `register_op1` can't express.
- **`join3`, `try_join`, `try_join3`** — multi-input edges, dispatched through the same convention `custom_op.rs` already proves (`no_builder`; their interpreted `Builder` methods are the hand-written `trimap`/`try_bimap`/`try_trimap`).

`buffer` and `try_map` already carried `#[op]` and already reached `compiled()` — no change.

### Tests
New `compiled_stateful_ops.rs`: `interpreted() == compiled() == nested()` for each newly-reachable op, asserting exact values (and tick times via `.ticked_at()` for `throttle`/`window`). The `window` run ends on a natural flush boundary so the run-end `is_last_cycle` flush — deliberately not propagated into a nested island — is a no-op and all three engines agree. Plus an error-propagation test: a `try_join` returning `Err` aborts interpreted and compiled identically.

### No unsupported shapes
Every op in scope was expressible under the current `#[op]` + fallback — no macro extension was needed. (`throttle`/`window` needed the `Cfg = Duration` fix in the op itself, not a macro change.)

### Verification
`cargo fmt --all`; `cargo test -p wingfoil-next -p wingfoil-next-macros` (all green, incl. 8 new tests); `cargo clippy -p wingfoil-next -p wingfoil-next-macros --all-targets -- -D warnings` (clean). No `.unwrap()` outside `#[cfg(test)]`.

---

PR #479, opened 2026-07-22
POC: call a wingfoil-next compiled() graph from Python

0-jake-0:
## What this demonstrates

`wingfoil-python` currently binds **only** the classic interpreted engine — a
Python graph is assembled node-by-node at runtime and dynamically dispatched.
Nothing in `wingfoil-python` touched `wingfoil-next` (the dual-mode prototype
whose `graph!` macro emits a fully-monomorphized `compiled()` runner).

This POC changes that for **one fixed graph**. It wires `wingfoil-next` into
the PyO3 bindings and exposes both engines of a single graph to Python:

```
squares:  ticker(10ms) -> count -> map(i*i) -> accumulate  ->  Vec<u64>
```

- `wingfoil.compiled_squares(realtime=..., cycles=...)` runs the graph through
  `wingfoil-next`'s `compiled()` — a straight-line static schedule with every
  op call monomorphized/inlined (no dynamic dispatch, no `RefCell` slots).
- `wingfoil.interpreted_squares(...)` runs the **same** graph through the
  interpreted engine.

Both entry points are generated by the `graph!` macro from a *single* wiring
definition, so they cannot drift — the POC's load-bearing result is that they
produce **identical output** when called from Python.

## Changes

- `wingfoil-python/Cargo.toml` — add `wingfoil-next` as a path dependency.
- `wingfoil-python/src/py_compiled.rs` (new) — the fixed `graph! { squares }`
  plus `compiled_squares` / `interpreted_squares` pyfunctions. Signatures
  mirror the classic `Node.run` (`realtime` / `start` / `duration` / `cycles`
  -> `RunMode` / `RunFor`), reusing `types::parse_run_args`, and release the
  GIL for the duration of the run. No `.unwrap()` in the binding code — errors
  propagate via `?` / `anyhow` and surface through the existing `ToPyResult`.
- `wingfoil-python/src/lib.rs` — register the module and the two functions.
- `wingfoil-python/tests/test_compiled.py` (new) — see below.

## How to build & run

```bash
cd wingfoil-python
python3 -m venv --system-site-packages .venv && . .venv/bin/activate
pip install maturin pytest pandas
maturin develop --release          # protoc required (apt install protobuf-compiler)
pytest tests/test_compiled.py -s
```

## Correctness result (required — passing)

```
tests/test_compiled.py::TestCompiledGraph::test_compiled_matches_expected      PASSED
tests/test_compiled.py::TestCompiledGraph::test_compiled_matches_interpreted   PASSED
tests/test_compiled.py::TestCompiledGraph::test_duration_bound                 PASSED
tests/test_compiled.py::TestCompiledGraph::test_timing_comparison              PASSED
```

The parity test checks `compiled == interpreted == [1, 4, 9, ...]` across
`cycles = 1, 5, 20, 100` and a duration-bounded run — exact agreement.

## Timing (nice-to-have)

A best-of-5, 50k-cycle release-build run prints, e.g.:

```
[compiled POC] 50000 cycles x best-of-5 (release): compiled=273.4ms interpreted=272.7ms interpreted/compiled=1.00x
```

Honest read: for **this** small, ticker-bound graph the two engines time out
roughly neutral (the ratio wanders ~0.9–1.15x across runs, i.e. noise). The
per-cycle cost here is dominated by work that is *identical* in both engines —
the kernel's ticker self-scheduling and the `accumulate` `Vec` growth — plus
the GIL detach/reacquire, so `compiled()`'s inlined dispatch is only a small
slice of the total. The prototype's interpreted engine is already lean (one
dyn boundary per op, shared `Kernel`), so the dispatch win is real but not
visible end-to-end on a graph this shape/size. The test therefore does **not**
assert on speed; it just prints the numbers.

## Limitations (it's a POC)

- **One fixed, compile-time graph.** This is *not* general Python-defined-graph
  codegen — the whole point of `compiled()` is that the graph is known and
  monomorphized ahead of time. A Python-authored arbitrary graph cannot be fed
  to `compiled()` without a code-generation step that is out of scope here.
- Timing is roughly neutral for this graph shape (see above); a dispatch-bound,
  many-node graph would show `compiled()`'s advantage more clearly, but that
  wasn't the goal — parity was.
- Built and tested against base branch `next` with Python 3.11 / maturin 1.14 /
  pyo3 0.29. `protoc` must be installed for the (unrelated) `etcd` default
  feature to build.

---


0-jake-0:
Migrated to #506 and closing ahead of the incoming refactor — branch `claude/py-compiled-poc` deleted. The issue captures what the POC established; revisit against the post-refactor `compiled()` API.

---

PR #481, opened 2026-07-22
feat(next-macros): add rolling min/max/var/std/median to the compiled path

0-jake-0:
## What

PR3 of the compiled-path completion (follows #475 and #477, branched fresh
from `next`). Completes the f64 rolling-window statistics family in the
`graph!` / `compiled()` / `nested()` path:

`rolling_min`, `rolling_max`, `rolling_var`, `rolling_std`, `rolling_median`

## How

Each is a one-line `OpKind` variant + a fully-spelled `info()` row modelled
exactly on the existing `rolling_sum` row — `Inputs::One`,
`CfgInit::Expr { usize window }`, `StateInit::Default` over the op's state
struct (`RollingExtremeState` for min/max, `RollingMomentState` for var/std,
`RollingMedianState` for median), `ValueSeed::Default` (they seed `0.0`),
`Edges::AllActive` — plus a parse arm. **No new emission machinery:** all five
are stateful but purely input-activated (`Activation::NONE`), so the
straight-line compiled schedule represents them exactly.

## Tests — 3-engine parity (interpreted == compiled == nested)

`compiled_rolling_ops.rs` runs each op over the fixed non-monotonic series
`[3, 1, 4, 1, 5]` (mapped from `count`) with a window of 3, so min/max/median
evict from both ends and have unambiguous expected values:

| op | expected |
|----|----------|
| `rolling_min` | `[3, 1, 1, 1, 1]` |
| `rolling_max` | `[3, 3, 4, 4, 5]` |
| `rolling_median` | `[3, 2, 3, 1, 4]` |
| `rolling_var` (ddof=1) | `[0, 2, 2.333…, 3, 4.333…]` |
| `rolling_std` | `[0, √2, 1.527…, 1.732…, 2.081…]` |

The three engines run the same `Op::cycle` in the same order, so their f64
outputs are bit-identical (`assert_eq!`); the hand-computed expectations are
checked within `1e-10` (var/std carry the usual incremental-Welford residue).
Updated the `unknown_combinator` trybuild `.stderr` for the expanded list.

## Compiled-path catalog status — what's genuinely left

With this PR the compiled path covers **essentially the whole
macro-expressible catalog** (32 ops). The ops in `wingfoil-next::ops` that
still have **no** compiled/`graph!` equivalent, and why:

- **`poll`** (`Poll`, `Activation::ALWAYS`) and **`for_each`** (`Sink`,
  `Out = ()`) — **IO-edge source / sink.** Excluded from `graph!` *by design*:
  "IO-edge sources and sinks (`external`, `poll`, `for_each`) are not
  expressible — IO lives at the fluent layer, feeding compiled islands through
  their inputs." Not a gap to close.
- **`print`** (`Print`), **`timed`** (`Timed`), **`finally`** (`Finally`) —
  their observable effect lives in a **`stop` / `teardown` lifecycle hook**,
  and the compiled/nested emission currently wires only the `start` hook
  (`node_start`); there is no `node_stop` / `node_teardown`. These are
  diagnostic / cleanup / side-effecting nodes (a pass-through value stream plus
  an end-of-run print or timing summary), so they're low-priority — but
  emitting `stop`/`teardown` in the compiled loop and island teardown is the
  one remaining piece of machinery if we want them.

Everything else (all the pure value-producing combinators, timers, stats, and
fallible/multi-input variants) is now on the compiled path.

## Merge note

Additively edits the same `OpKind` / `info()` / parse-arm / combinator-list
regions as #475 and #477; any textual conflict is trivial to resolve at merge.

## Verification

- `cargo fmt --all` — clean
- `cargo test -p wingfoil-next -p wingfoil-next-macros` — all pass (default **and** `--features async`)
- `cargo clippy -p wingfoil-next -p wingfoil-next-macros --all-targets -- -D warnings` — clean (default and `--features async`)

---

### DIFF
--- a/wingfoil-next-macros/src/lib.rs
+++ b/wingfoil-next-macros/src/lib.rs
@@ -93,7 +93,9 @@
 //! `.filter(&cond)`, `.fold(init, f)`, `.sample(&trigger)`, `.merge(&other)`,
 //! `.join(&other, f)`, `.delay(duration)`; statistics (on `f64` streams)
 //! `.ewma(decay)` / `.ewma_per_tick(alpha)` / `.ewma_half_life(dur)`,
-//! `.rolling_sum(window)`, `.rolling_mean(window)`; sugar `.count()` and
+//! `.rolling_sum(window)`, `.rolling_mean(window)`, `.rolling_min(window)`,
+//! `.rolling_max(window)`, `.rolling_var(window)`, `.rolling_std(window)`,
+//! `.rolling_median(window)`; sugar `.count()` and
 //! `.accumulate()`; and bounded repetition `.map_n(N, f)` (chain `map` N
 //! times) / `.fan(N, |s| <sub-chain>)` (N parallel copies of a sub-chain,
 //! merged). `map_n` / `fan` counts must be integer literals so the unrolled
@@ -144,6 +146,11 @@ enum OpKind {
     Ewma,
     RollingSum,
     RollingMean,
+    RollingMin,
+    RollingMax,
+    RollingVar,
+    RollingStd,
+    RollingMedian,
 }
 
 /// The shape of an op's `Op::In` tuple — how the cycle call assembles its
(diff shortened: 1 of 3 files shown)


0-jake-0:
Superseded by #496. That change deleted the `graph!`/`compiled()` `OpKind`/`OpInfo` table and made every `#[op]`-annotated op reach `compiled()`/`nested()` through a generic forwarder fallback (~1.01× a hand-written table row). `rolling_min`/`rolling_max`/`rolling_var`/`rolling_std`/`rolling_median` all carry `#[op]` and now reach the compiled path for free, so the table rows here no longer apply. Closing as superseded.

---

PR #483, opened 2026-07-22
docs: Phase-1 contract proposals (Tick::Silent, reset hook, supported_ops!())

0-jake-0:
**Proposal only — no engine, macro, or op code changes.** Adds `docs/contract-proposals.md` so a human can ratify the three remaining Phase-1 contract-shape decisions before any of them lands.

## Executive summary (the three recommendations)

1. **`Tick::Silent(T)`** — *add the variant* (vs keep permanent engine-level special-casing). It deletes the last name-keyed dispatch hack — `delay`'s store-without-tick + zero-delay emit, today duplicated across the interpreted `seeded` flag (`interp.rs:1057-1114`) and the macro's `if node.op == OpKind::Delay` branch (`lib.rs:1700-1739`) — and lets `delay` be an ordinary `OneTick` op. LLVM folds the dead arm away for infallible ops (same argument as `Ok(..)`).
2. **Per-node `reset`/`setup` hook** — add `Op::reset(cfg, state, ctx) -> Result<()>` + a `NodeRt.reset` `LifecycleFn` mirroring the `stop`/`teardown` plumbing, run before `start` on every run. Restores state to its **wiring-time initial** (fold → `init`, not continued, not `Default`) and re-arms self-scheduling sources, making a `Runner` re-runnable and letting `compat::Signal::run` drop its double-run bail (`compat.rs:134-145`) — the wingfoil-python pytest gate depends on this.
3. **`supported_ops!()` completeness test** — single-source the macro's op list so the parse-match (`lib.rs:779-935`) and the "unsupported op" error string (`lib.rs:925-934`) can't drift, expose it via a `supported_ops!()` proc-macro, and diff it in a wingfoil-next test against the fluent surface with an explicit "not expressible" allowlist (feedback, IO-edge sources, …). Guards the forward direction (fluent-but-not-macro); the reverse is already guarded by `wire()` verbatim compilation.

**Recommended landing order:** `Tick::Silent` **first** (widest blast radius — retrofit once before the catalog volume; unblocks the `delay` port), then `supported_ops!()` (cheap guard, before the Phase 2 ramp), then the `reset` hook (before the Phase 6 facade/python gate; parallelizable with #2 once #1 lands).

## Ratification checklist

**1. `Tick::Silent(T)`**
- [ ] Add `Tick::Silent(T)` (write slot, `did = false`) in all three engines.
- [ ] Fold zero-delay into `Delay::cycle`; delete the `cfg == ZERO` engine branch.
- [ ] Delete the `OpKind::Delay` macro branch, the `__seeded` local, and the interpreted `seeded` flag; `delay` becomes an ordinary `OneTick` op.

**2. `reset`/`setup` hook**
- [ ] Add `Op::reset(cfg, state, ctx) -> Result<()>` (default no-op).
- [ ] Add `NodeRt.reset: LifecycleFn`, run before `start` on every run.
- [ ] Confirm semantics: restore to wiring-time initial (fold → `init`) + re-seed value slot + re-arm self-scheduling.
- [ ] Confirm scope: re-run for historical/timer graphs; realtime stays single-run (documented).
- [ ] Rewrite `compat::Signal::run` to build once + re-run (delete the bail).

**3. `supported_ops!()` completeness test**
- [ ] Canonical op list drives `apply_call` and both error messages.
- [ ] `supported_ops!()` proc-macro exposes the list to `wingfoil-next`.
- [ ] Forward-direction diff test with an explicit `NOT_EXPRESSIBLE` allowlist.

Full argument, options, and per-item test plans in `docs/contract-proposals.md`.

---

### DIFF
--- a/docs/contract-proposals.md
+++ b/docs/contract-proposals.md
@@ -1,0 +1,581 @@
+# Phase-1 contract proposals
+
+Status: **one open proposal remaining.** This document originally captured three
+Phase-1 contract-shape decisions (`docs/port-plan.md`, "Phase 1 — contract
+completion") for a human to ratify *before* any of them landed, because each one
+touches the `Op` contract, all three engines, and the macro at once — the exact
+blast radius the plan front-loaded fallibility (spike 0.1) to avoid retrofitting
+later. Two have since resolved on `next`: **§1 `Tick::Silent(T)` is implemented**
+(`wingfoil-next/src/op.rs:85-89`, via #496) and **§3 `supported_ops!()` is
+superseded** (#496 deleted the macro op-table). Only **§2, the per-node
+`reset`/`setup` hook, is still an open decision.**
+
+The three items — of which **only item 2 is still open**; items 1 and 3 have since
+landed / been made moot on `next`:
+
+1. ~~**`Tick::Silent(T)`**~~ — **implemented on `next`** (see §1). The
+   `Tick::Silent(T)` variant landed in `wingfoil-next/src/op.rs:85-89` (via #496 /
+   the same line of work that deleted the op-table), and `delay`'s
+   store-without-tick semantics are now expressed through it generically across all
+   engines. No longer an open decision.
+2. **A per-node `reset`/`setup` hook** — the missing lifecycle hook that makes a
+   `Runner` re-runnable, which the compat facade (and the wingfoil-python gate)
+   requires. **This is the one remaining live proposal.**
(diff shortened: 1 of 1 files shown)


0-jake-0:
Migrated to #499 — tracking the remaining open item (the per-node `reset`/`setup` hook; `Tick::Silent` and `supported_ops!()` are already resolved on `next`). Closing this draft; the design doc stays on branch `claude/contract-proposals` (`docs/contract-proposals.md`) and is re-openable if needed.

---

PR #485, opened 2026-07-22
docs: design + feasibility spike for wingfoil-next runtime graph dynamism

0-jake-0:
## What this is

A **design + feasibility spike** (`docs/dynamism-design.md`) for adding runtime graph dynamism (add/remove nodes and sub-graphs mid-run) to the `wingfoil-next` interpreted engine. **No engine code changed** — the deliverable is the doc.

## Key finding — the mutable-frontier dirty-list ALREADY EXISTS

Commit #471 ("sparse dirty-list dispatch engine") landed. The next interpreted engine's **production** dispatch path is already a sparse, source-seeded, mutable-frontier dirty-list — not the O(N) per-cycle sweep the port-plan still describes.

- `Dispatch::Sparse` is `#[default]`; `run_cycles_sparse` drives a `BinaryHeap<Reverse<usize>>` work heap seeded from a precomputed `seed_nodes` frontier and propagated through `active_downs` (`wingfoil-next/src/interp.rs:1498-1505, 1639-1712`).
- Adjacency + frontier precomputed once at `build()` (`interp.rs:1459-1485`).
- The old O(N) sweep survives only as a differential oracle (`run_cycles_full_sweep`, `interp.rs:1720-1749`).

`docs/port-plan.md` is **stale** on this (still lists Phase 4.5 as an unstarted gap and dynamism as needing the dirty-list first). **Consequence:** dynamism no longer needs a scheduler rewrite — it needs a post-`build()` mutation API plus local bookkeeping to grow the already-existing sparse structures.

**The one deep catch (the crux):** the sparse engine's correctness rests on *node index order being a valid topological order over all edges*. Classic instead dispatches by an explicit `layer` field and re-runs `fix_layers` on every dynamic add (`wingfoil/src/graph.rs:205, 1249-1275`). That is exactly what lets classic splice a new node *into an existing lower-indexed caller* — the move next's index-order heap cannot express without help.

## Proposed API (sketch)

A scoped mutation session on the live `Runner`, reusing the existing node-registration paths and `builder_id`-stamped handles:

```rust
impl Runner { pub fn extend(&mut self) -> Extension<'_>; }

impl<'r> Extension<'r> {
    pub fn map<A,B>(&mut self, src: Handle<A>, f: impl Fn(&A)->B + 'static) -> Handle<B>;
    pub fn add_upstream<T>(&mut self, caller: Handle<T>, new: Handle<T>,
                           active: bool, recycle: bool) -> anyhow::Result<()>;
    pub fn remove(&mut self, node: Handle<impl Sized>);
    pub fn commit(self);
}
```

Changes stage on `commit()` and apply at the next cycle boundary — reproducing classic's "requested in cycle N, live in N+1". Fluent `Stream`/`GraphBuilder` stay build-time only (keep their post-build poison); dynamism is a `Runner`/`Handle` capability, matching classic.

## First implementation slice

**Phase 1 — append a single node onto a live graph.** `Runner::extend()` → `Extension::map(existing_handle, f)` → `commit()`; the new node ticks on the *next* cycle, value observable via `runner.value(handle)`. Restricted to pure append (no interior splice, no `Ctx` change, no layer field). Exercises growing `nodes`/`slots`/`ticked`/`active_downs`/`seed_nodes`/`dirty`/`node_dirty` + the +1-cycle timing.

Parity oracle: a next port of classic's `add_upstream_dynamically_fires_only_after_wired` (`graph.rs:2102-2139`) — add at end of cycle 3, assert 3 ticks on 4/5/6.

## Ratification checklist (decisions a human must make)

1. **Dispatch key (the crux):** ship append-only (A) first and defer interior splicing, or commit up front to the layered `(layer, index)` engine (B)? Faithful `dynamic_group`/`demux` parity is impossible under A. *Recommend A first, B behind a Phase-3 gate.*
2. **Mutation entry point:** is the between-cycles `Runner::extend` enough for v1, or is the in-`cycle` `Ctx::request_extend` hook (widens the deliberately-narrow `Ctx`) required from the start (the real `DynamicGroup` client mutates from inside a node)?
3. **Timing contract:** confirm "applied at the N/N+1 boundary, live in N+1".
4. **Removal = tombstone, never free** (classic parity, documented leak) for v1?
5. **`dirty` unchecked-write hazard:** grow scratch before any schedule can target a new index.
6. **Fluent/`Stream` stays static** — no `.add_child()` surface in v1.
7. **Feature gating** — mirror classic's `dynamic-graph` cargo feature.
8. **Update the stale `docs/port-plan.md`** to record #471 landed and dynamism is now narrow API work.

## Arena coupling recommendation

Dynamism does **not** need the arena/SoA value store first and should not wait for it. Append is append-only on `slots`; removal tombstones under either representation (classic already leaks by design). The only real constraint is the reverse — the future arena must keep handles index-stable across an append, which is exactly the frozen slot-boundary posture `interp.rs:26-31` already commits to. **Layer dynamism onto the current slot store now; treat "handles survive append, tombstone-don't-free on remove" as the frozen boundary the arena must honour.**

## Scope

Doc-only; `git status` shows only the new `docs/dynamism-design.md`. No probe committed. Every claim is `file:line`-backed (see the doc's evidence appendix).

---

### DIFF
--- a/docs/dynamism-design.md
+++ b/docs/dynamism-design.md
@@ -1,0 +1,602 @@
+# Runtime graph dynamism for wingfoil-next — design + feasibility spike
+
+Status: **design spike, no engine code changed.** This document establishes the
+actual current state of the `wingfoil-next` interpreted engine, the classic
+semantics dynamism must reproduce, a concrete proposed API, the engine mechanics
+and their hazards, and a phased plan with parity oracles. It closes the
+"Dynamic graphs" decision left open in `docs/port-plan.md` (Phase 2 and Phase
+4.5) with a recommendation and a ratification checklist for a human to sign off
+before implementation.
+
+All claims below are backed by `file:line` references against the branch this
+spike is based on (`origin/next`).
+
+---
+
+## 1. Current-state finding — the mutable-frontier dirty-list ALREADY EXISTS
+
+**Finding: commit #471 ("sparse dirty-list dispatch engine") landed. The
+wingfoil-next interpreted engine no longer does an O(N) per-cycle sweep as its
+production path — it runs a sparse, source-seeded, mutable-frontier dirty-list,
+and the old full sweep survives only as an executable reference oracle.**
+
+This directly contradicts `docs/port-plan.md`, which is **stale**: its capability
+matrix still lists "Sparse-graph efficiency" as `❌` for "Interpreted (today)"
+(port-plan.md:64), Phase 4.5 is written as an unstarted "Gap (must close)"
+(port-plan.md:388-399), and the risk register still frames the dirty-list as the
+future enabler for dynamism (port-plan.md:553). The code has moved past the plan.
+
+Evidence in `wingfoil-next/src/interp.rs`:
+
+- **Two dispatch strategies exist**, selected by an enum (interp.rs:1498-1505):
+  `Dispatch::Sparse` is `#[default]`; `Dispatch::FullSweep` is documented as the
+  "original `O(N)`-per-cycle topological sweep, retained as an executable
+  reference oracle."
+- **`Runner::run` dispatches on it** (interp.rs:1597-1600): `Dispatch::Sparse =>
+  self.run_cycles_sparse(...)`, `Dispatch::FullSweep =>
+  self.run_cycles_full_sweep(...)`.
+- **The sparse loop is a real dirty-list with a mutable frontier**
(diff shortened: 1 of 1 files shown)


0-jake-0:
Migrated to #500. Closing this draft; the feasibility spike stays on branch `claude/dynamism-design-spike` (`docs/dynamism-design.md`) and is re-openable if needed.

---

PR #484, opened 2026-07-22
wingfoil-next: port cumulative time-weighted moments

0-jake-0:
## What

Ports the **cumulative (unbounded-window) time-weighted** mean/variance/std from the classic `adapters::statistics` operators — the `Weighting::Time` path through `WeightedMoments` (West's weighted-Welford), driven by `MomentStream`. Follows on from the count-weighted cumulative family in #482; the two are additive (both touch `ops.rs`/`stats.rs` in adjacent regions).

### Stats landed

| op | classic reference | notes |
|----|-------------------|-------|
| `time_weighted_mean` | `MomentStream` (`Weighting::Time`) | Σwx / Σw, w = Δt in effect |
| `time_weighted_var` | `MomentStream` (`Weighting::Time`) | **population** form `m2 / Σ Δt` |
| `time_weighted_std` | `MomentStream` (`Weighting::Time`) | sqrt of var, clamped ≥ 0 |

## Exact weighting semantics reproduced

Each sample is weighted by **how long it was in effect**, read from the graph clock (`ctx.time()`):

- On every tick the **previous** value is credited with the elapsed Δt *before* the new value takes over — the correct treatment of a left-continuous step signal. So with N ticks only N−1 samples are ever weighted (the newest sample opens an interval closed only by the next tick).
- **First sample** seeds the tracker with no weight: the emitted mean is the raw sample (not `NaN`), variance is `0.0`.
- **Zero-Δt** (simultaneous) tick adds no weight (`push` guards `weight <= 0.0`).
- Variance is the **population** form `m2 / Σ Δt` — time weighting has no clean ddof correction, so (matching classic) it is not the sample (ddof = 1) form the count-weighted operators use.

The three ops share a `CumulativeTimeMomentState` carrying West's moments (`w_sum`/`count`/`mean`/`m2`) plus the last tick time and the pending previous value. Each is a single-input `Op` with `#[op(build = ...)]` (auto-generates the interpreted `Builder` wiring); fluent methods added to `StatisticsOps` in `stats.rs`.

## Parity approach

Tests in `wingfoil-next/tests/statistics_time_weighted.rs` (historical mode) pin the emitted series **tick-by-tick** against hand-computed time-weighted moments on a known `(value, Δt)` sequence with unambiguous timestamps:

[...]

PR #525, opened 2026-07-25
wingfoil-next: port the time-weighted statistics moment path

0-jake-0:
## What

Ports the classic **`Weighting::Time` moment path** (mean / variance / std) from the legacy statistics adapter (`MomentStream` / `RollingMomentStream`) to the wingfoil-next op catalog. This is the remaining moment slice of the Phase 4 statistics port (`docs/port-plan.md`): the count-weighted cumulative / count-windowed / time-windowed families and the time-weighted *cumulative* moments were already in place; this adds the time-weighted **windowed** paths and unifies the whole family behind consistent naming.

> Supersedes the earlier partial branch behind the (closed, unmerged) #484, which landed only the cumulative case under the non-scaling name `time_weighted_mean`. This PR reworks it on the current `next` and covers all three windows under a naming scheme consistent with the count-weighted ops.

### Ops landed (9), all single-input `Op`s via `#[op(build = …)]` + fluent methods

| window | ops | classic reference |
|---|---|---|
| cumulative (unbounded) | `cumulative_{mean,var,std}_time_weighted` | `mean/variance/std(Window::Unbounded, Weighting::Time)` → `MomentStream` |
| count window | `rolling_{mean,var,std}_time_weighted` | `…(Window::Count(n), Weighting::Time)` → `RollingMomentStream` |
| time window | `time_windowed_{mean,var,std}_time_weighted` | `…(Window::Time(d), Weighting::Time)` → `RollingMomentStream` |

## Exact weighting semantics reproduced

Each sample is weighted by **how long it was in effect** — the elapsed Δt until its successor, read from the graph clock (`ctx.time()`):

- on every tick the **previous** value is credited with the elapsed Δt *before* the new value takes over (a left-continuous step signal), so the newest sample contributes nothing until the next tick advances the clock;
- a windowed sample's committed interval is **reverted** when it leaves the window (an exact `remove` inverse keeps a tick O(1); count window evicts when more than `window` samples are retained, time window when a front entry ages strictly past `window` ns);
- the **mean** seeds to the current sample until any weight has accumulated (not `NaN`); the **variance** is the time-weighted *population* form `m2 / w_sum` (no ddof — reliability-weighted variance has no clean ddof), `0.0` until weight is present; **std** clamps at zero.

`WeightedMoments` (West's weighted-Welford with the exact `remove` inverse) is ported verbatim from the classic adapter. The count and time windows share one `WindowedTimeWeightedMomentState`, differing only in the eviction rule; the cumulative ops use `CumulativeTimeWeightedMomentState`.

## Parity tests

`tests/statistics_time_weighted.rs` (historical mode, `RunMode::HistoricalFrom(NanoTime::ZERO)`) pins emitted **values and tick times** against the classic unit tests and hand-computed series:

- cumulative mean lags by one interval → `1, 1, 1.5, 2, 2.5` (classic `mean_time_weighted_lags_by_one_interval`);
- cumulative variance → `0, 0, 0.25, 2/3, 1.25` (classic `variance_time_weighted_is_population_over_weight`);
- time-windowed mean → `…, 3.5` (classic `rolling_mean_over_time_window_count_and_time`, time case) and var/std → `0.25` / `0.5` (classic `rolling_var_std_over_time_window_time_weighted`);
- hand-computed count-window case; incremental-vs-direct-recompute cross-checks (count and time windows); a wide count window equals the cumulative statistic tick-for-tick; constant-stream std clamps to `0.0`; tick-time parity.

## Deferred (documented, not silently dropped)

The one remaining classic statistics capability is the time-weighted **median** (`median(_, Weighting::Time)` → `WindowStream::weighted_median`) — a distinct recompute-per-tick weighted median, not a moment. Noted in `docs/port-plan.md` and the example README.

## Verification

- `cargo lint` (default) and `cargo lint-all` (all features) — clean
- `cargo test -p wingfoil-next --all-features` — the new `statistics_time_weighted` binary passes (10/10) and the full suite is green **except** the `etcd_integration` tests, which fail only on `failed to initialize a docker client: Socket not found: /var/run/docker.sock` (testcontainers needs Docker, unavailable in this sandbox — unrelated to this change; the no-service `etcd_adapter` tests pass).

---

https://claude.ai/code/session_01YCSuEZ1SGPcVqKUT8RXyKU

---
_Generated by [Claude Code](https://claude.ai/code/session_01YCSuEZ1SGPcVqKUT8RXyKU)_

### DIFF
--- a/next/crates/wingfoil-next/examples/statistics/README.md
+++ b/next/crates/wingfoil-next/examples/statistics/README.md
@@ -39,6 +39,10 @@ second, and printed:
   a chain of `join`s building a `Vec<f64>` row. A variadic `combine`/`zip` op is
   a known gap.
 - **Time-weighting** — the classic example also showed a *time-weighted* mean
-  (`Weighting::Time` / TWAP). The next `StatisticsOps` trait currently exposes
-  the count-weighted rolling ops (`rolling_mean/std/min/max/median`, `ewma`); a
-  time-weighted aggregation family is not yet ported, so this column is omitted.
+  (`Weighting::Time` / TWAP). The `StatisticsOps` trait now exposes the
+  time-weighted moment family alongside the count-weighted ops —
+  `{cumulative,rolling,time_windowed}_{mean,var,std}_time_weighted` (each sample
+  weighted by how long it was in effect, read from the graph clock) — so a TWAP
+  column is expressible; this example keeps the count-weighted columns for
+  brevity. The only classic statistic still unported is the time-weighted
+  *median*.
--- a/next/crates/wingfoil-next/src/adapters/etcd.rs
+++ b/next/crates/wingfoil-next/src/adapters/etcd.rs
@@ -244,9 +244,17 @@ pub fn etcd_sub(
         // 4. Return the combined snapshot + live stream. `watch_stream` is moved
         //    in and kept alive for the stream's lifetime so the watch stays open.
         Ok(async_stream::stream! {
-            // Phase 1: emit snapshot events.
+            // Phase 1: emit the snapshot as a single atomic burst. Every
+            // snapshot KV shares ONE timestamp so they are grouped into one
+            // burst (never latest-wins, never split across cycles) — matching
+            // classic's single `HistoricalValue` snapshot burst. Stamping each
+            // event with its own `NanoTime::now()` would scatter them across
+            // distinct instants, so a bounded run (e.g. `RunFor::Cycles(1)`)
+            // could observe only the first key — an intermittent "missing key"
+            // under load.
+            let snapshot_time = NanoTime::now();
             for event in snapshot {
-                yield Ok((NanoTime::now(), event));
+                yield Ok((snapshot_time, event));
             }
 
             // Phase 2: drain the watch stream, deduplicating against the snapshot.
--- a/next/crates/wingfoil-next/src/ops.rs
+++ b/next/crates/wingfoil-next/src/ops.rs
@@ -1739,6 +1739,432 @@ impl Op for TimeWindowedMedian {
     }
 }
 
+// ── time-weighted moments (Weighting::Time) ──────────────────────────────────
+//
+// The time-*weighted* twin of the count-weighted moment ops above (cumulative,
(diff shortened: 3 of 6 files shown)


0-jake-0:
CI (`test.rust`) is red, but the failure is **unrelated to this PR** and looks like a flake:

```
etcd_integration::test_sub_no_race_between_snapshot_and_watch
  panicked at tests/etcd_integration.rs:257: concurrent key missing
```

- This PR touches only the statistics op catalog (`src/ops.rs`, `src/stats.rs`) plus a new `tests/statistics_time_weighted.rs` — **no etcd code**. The failing test is byte-identical to what's already on `next` (landed in #517).
- It's a timing-sensitive snapshot-vs-watch race test; the other 11 etcd integration tests passed in the same run. `test_sub_no_race_between_snapshot_and_watch` seeds `/race/new` from a side thread and asserts the `etcd_sub` snapshot observes it — a category prone to intermittent misses under load.
- I attempted to re-run the failed job but the integration lacks that permission (`403 Resource not accessible`).

Could a maintainer re-run the `test.rust` job? I expect it to go green. If it reproduces, it points at a real snapshot/watch race in the etcd source (#517 territory), not this change — happy to open a separate issue for that if so.

---

0-jake-0:
Update: root-caused and fixed it in `43a1cab` — it wasn't just a flake to re-run past.

`etcd_sub` stamped each snapshot KV with its own `NanoTime::now()`, so a multi-key snapshot was scattered across distinct instants and delivered as several bursts at different scheduled times. Under `RunFor::Cycles(1)` only the first burst is observed, so a later key intermittently went missing — which is exactly what `test_sub_no_race_between_snapshot_and_watch` caught once CI load widened the gap between the two `now()` calls. (This PR added a heavy concurrent test binary, which is likely why the latent bug surfaced here.)

The fix captures one timestamp for the whole snapshot so its events share a single burst, matching classic's single `HistoricalValue` snapshot delivery; live watch events keep their per-event timestamps. `cargo clippy --features etcd -- -D warnings` and `fmt` are clean. CI will re-run on the push.

---

PR #546, opened 2026-07-26
next: defer I/O source establishment to start() (source_at_start)

0-jake-0:
Adds `source_at_start` — the **deferred-connection** source primitive for
wingfoil-next I/O adapters — so an adapter's producer (thread / socket) is
established in `start()` rather than at graph construction.

## What changes

- **`Builder::source_at_start` + `SourceOps::source_at_start`** (`interp.rs`,
  `fluent.rs`): a `channel`-fed source whose `setup` closure runs on each run
  start, is handed the `ChannelSender`, connects/spawns the live producer, and
  returns a `StopHandle`. The factory itself performs **no I/O**, so adapter
  wiring (address parse, registry lookup, historical-mode rejection) stays pure
  and unit-testable; a `setup` error aborts the run at start with node context.
- **`StopHandle`** (`interp.rs`): a teardown guard held for the run and dropped
  at teardown — generalises the zmq `ThreadStopGuard` pattern into a
  start-scoped handle that wraps any `'static` guard whose `Drop` stops the
  producer. The `start`/`teardown` hooks are *composed* onto the underlying
  channel node (not clobbered), so the historical collect and any future
  channel teardown still run, in the right order.
- **zmq adapter** (`adapters/zmq.rs`): ported onto `source_at_start`, dropping
  its bespoke deferred-connection plumbing.
- **Tests** (`tests/fluent_primitives.rs`): setup deferred to run start and
  guard dropped at teardown; `setup` error aborts the run; historical replay of
  timestamped sends.
- **Docs / skill**: `docs/source-lifecycle-defer-to-start.md` and the
  `new-adapter-next` command updated to teach the deferred-connect pattern.

## Notes

- Targets **`next`** (all wingfoil-next work lives there).
- Single-run for now: the receive channel and waker are consumed by the first
  run — re-run is a documented follow-on (see the source-lifecycle doc).
- A *historical* deferred producer must `close()` explicitly: the source
  retains a live `ChannelSender` for the run, so the all-senders-dropped
  termination path doesn't apply. Realtime live sources (the primary use) are
  unaffected — they terminate on `close()` or the run bound.

### DIFF
--- a/.claude/commands/new-adapter-next.md
+++ b/.claude/commands/new-adapter-next.md
@@ -11,6 +11,9 @@ reference implementations; read them before writing code:
   realtime `poll` tail + sink), dependency-free.
 - `src/adapters/csv.rs` — the serde-typed parsing cousin (feature-gated deps,
   wiring-time `Result`, header introspection).
+- `src/adapters/zmq.rs` — a live sync-streaming subscriber over a background
+  thread, using [`source_at_start`] to **defer the socket connect + thread spawn
+  to graph `start()`** (see the source shape in step 7); plus a status stream.
 - `src/adapters/augurs.rs` — a pure-compute adapter (custom `Op`s +
   `#[op(build = ...)]` + config builder types, no I/O).
 - `src/async_source.rs` — `produce_async` for async client libraries.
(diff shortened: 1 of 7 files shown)


0-jake-0:
Closing as superseded: the `source_at_start` deferred-connection work landed on `next` via #547, so this branch has no unique commits remaining.

---

PR #550, opened 2026-07-26
wingfoil-next: latency capture infrastructure + otlp trace/span export (Phase 5)

0-jake-0:
## Summary

Two related pieces of Phase 5 infrastructure for wingfoil-next, in dependency order:

1. **Latency capture infrastructure** — ports classic `wingfoil::latency` onto the Op engine.
2. **OTLP trace/span export (`otlp_spans`)** — the classic `OtlpSpans` sink, which depended on the latency infrastructure above and was the last tracked gap in the otlp adapter.

---

## Part 1 — Latency capture (`src/latency.rs`)

The pure data layer is engine-agnostic and reused from the classic crate unchanged; only the node layer is re-implemented as ops.

- **`Kernel` (shared engine core, `wingfoil/src/codegen.rs`)** — gains a per-cycle wall-clock snap plus a `wall_time()` accessor, mirroring classic `GraphState::wall_time`. Distinct from the source-driven logical `time`, so stamps mean "wall-clock time spent" in both realtime and historical mode. Pure-additive change.
- **`Ctx` (`op.rs`)** — exposes `wall_time()` (cached per-cycle snap; stages in one cycle share it) and `wall_time_precise()` (fresh TSC read).
- **`src/latency.rs`** — re-exports the data layer (`Traced`, `Latency`, `Stage`, `HasLatency`, `StageStats`, `LatencyStats`, `latency_stages!`) from classic unchanged, and implements `Stamp`/`StampPrecise` pass-through ops (over `register_op1`) and the `LatencyReport` sink op (a `Builder` method with a stop hook). Exposed via `LatencyStreamOps` (`stamp`/`stamp_if`/`stamp_precise`/`stamp_precise_if`) and `LatencyReportOps` (`latency_report`/`latency_report_if`).
- **`tests/latency.rs`** — 9 node-layer parity tests (wall-clock stamping, shared-per-cycle vs fresh stamps, per-stage delta aggregation, the `_if` toggles).
- **`Cargo.toml` (workspace)** — declares `iceoryx2` as a known cfg (the `latency_stages!` derive emits that gate unconditionally); additive, `wingfoil` already has the real feature.

**Deviation**: fluent/interpreted only, matching classic — a stamp's stage is a compile-time *type* parameter that does not map onto the `graph!` value-dispatch table, so compiled/nested support is out of scope for this op family.

## Part 2 — OTLP trace/span export (`otlp_spans`)

- New `OtlpSpanOps` extension trait + `OtlpAttributeBuffer` in `src/adapters/otlp.rs`, alongside the existing `OtlpSinkOps` metrics sink. `stream.otlp_spans(&handle, name, config, attrs)` emits one parent span per tick plus one child span per stage hop.
- Same off-thread `consume_async` model as `otlp_push`: tracer provider built lazily on the first exported value and dropped at teardown to flush; a no-op under `RunMode::HistoricalFrom`.
- Every classic span capability preserved (parent + per-hop children, caller-supplied attributes, silent skip of all-zero / backwards timestamps). Adds the `trace` feature to the OTel deps, mirroring classic.
- Tests: `spans_historical_mode_drains_without_connecting` (`tests/otlp_adapter.rs`) and `otlp_spans_sends_successfully` (testcontainers, `tests/otlp_integration.rs`).

---

## Docs

`next/docs/port-plan.md` — Phase 5 Latency item marked landed; the otlp adapter status updated to *(metrics + traces done)*.

## Testing

- `cargo fmt --all` / `cargo lint` / `cargo lint-all` — all clean.
- `cargo test -p wingfoil-next` (default) and `--features otlp` — all pass, including the 9 latency tests and 4 otlp-adapter tests.
- `cargo test -p wingfoil --lib` — all 260 classic lib tests pass (guards the shared `Kernel` change).
- The `otlp_integration.rs` span test compiles under `otlp-integration-test` (needs a Docker collector to run).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,12 @@ rust-version = "1.88"
 
 [workspace.lints.rust]
 unsafe_op_in_unsafe_fn = "warn"
+# The `latency_stages!` derive unconditionally emits an
+# `#[cfg(feature = "iceoryx2")]` ZeroCopySend impl. Crates that invoke the
+# macro without an `iceoryx2` feature (e.g. wingfoil-next) would otherwise get
+# an `unexpected_cfgs` warning; declare the feature name as known so the gate
+# simply stays off there. `wingfoil` has a real `iceoryx2` feature already.
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(feature, values("iceoryx2"))'] }
 
 [workspace.lints.clippy]
 type_complexity = "allow"
--- a/next/crates/wingfoil-next/Cargo.toml
+++ b/next/crates/wingfoil-next/Cargo.toml
@@ -132,10 +132,10 @@ zmq = { version = "0.10.0", optional = true }
 # stream values as OTLP gauge metrics over HTTP/protobuf to any OTLP-compatible
 # backend (Grafana Alloy, Datadog, Honeycomb, New Relic, …). Built on the async
 # `consume_async` ergonomic so exports run off the graph thread (the OTel SDK's
-# `rt-tokio` PeriodicReader needs a tokio runtime). Only the metrics features are
-# enabled (the classic trace/span exporter is a deferred deviation — it needs the
-# `Traced`/`HasLatency` latency infrastructure, not yet ported to next; see
-# `src/adapters/otlp.rs`).
+# `rt-tokio` PeriodicReader needs a tokio runtime). Both the `metrics` and
+# `trace` features are enabled: the metrics gauge sink (`otlp_push`) and the
+# trace/span exporter (`otlp_spans`, over the ported `Traced`/`HasLatency`
+# latency infrastructure) — see `src/adapters/otlp.rs`.
 #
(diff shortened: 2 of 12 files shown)


0-jake-0:
The `Build, Test, & Lint` failure was **not** from this branch's changes — it failed only on the etcd integration test `test_sub_no_race_between_snapshot_and_watch`, a `RunFor::Cycles(1)` realtime snapshot-vs-watch race against a live etcd container. This PR touches only the latency / otlp / kernel code; the one shared change (the per-cycle `Kernel::wall_time` snap) is a nanosecond clock read added after all scheduling logic, so it can't affect etcd watch/snapshot interleaving. The other 11 etcd tests and the OTLP (next) integration tests — including the new `otlp_spans_sends_successfully` span test against a real collector — all passed.

I couldn't trigger a job re-run via the API (`403 Resource not accessible by integration`), so I've pushed an empty commit to reroll CI. If it recurs on the same test, that points to pre-existing flakiness in the etcd race test rather than anything here.

---

PR #579, opened 2026-07-26
Compiled islands: nest a graph! sub-graph as one node of an interpreted graph

0-jake-0:
A graph! wiring fn can now take input streams after the builder
(`fn ema(g: &GraphBuilder, price: &Stream<f64>) -> Stream<f64>`), and
every single-output graph gains a `nested(g, inputs...)` expansion that
mounts the whole sub-graph as a single composite node of an interpreted
graph under construction. Inside the composite one closure owns all
inner state and runs the same monomorphized straight-line dispatch
`compiled()` emits; the outer engine pays one dyn call per activation
for the entire island.

Ctx gains a nested mode backing this: an inner op's schedule lands in
the composite's private TimeQueue keyed by inner node index instead of
the outer kernel. The composite demultiplexes due entries into its
inner dirty flags each activation and forwards only the earliest
pending time outward, so inner tickers and delays run on the outer
kernel's clock with no drift. Passive inputs (sample's data edge) are
excluded from the composite's active upstreams and cannot activate it.

Graphs with inputs expand to wire + nested only (they cannot run
standalone) and must have exactly one output; source-only graphs keep
all four expansions. Stream-ref arguments now also accept a bare input
parameter name (`data.sample(trigger)`) since inputs are already
references. Tests cross-check each island against identical flat
wiring in the same graph — same kernel, same cycles, exact agreement —
covering input islands, inner delay scheduling, source islands, and
passive-input gating.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01RSP1oHTaGBeMmnGmYa5XnS


0-jake-0:
Closing as a duplicate — this branch's work already merged into `next` via #580.

Same situation as #594: this PR targets `main` while #580 targeted `next`, off the same branch. The `next` one landed. The diff here is pre-inversion (old top-level `next/` layout), so it would read as a large reverse diff against the post-cutover tree.

The `/review-adapters-next` skill this branch added is present on `next` at `.claude/commands/review-adapters-next.md`.

---

PR #582, opened 2026-07-26
next: layer-bucketed dispatch — close the dense-fanout interpreted gap

0-jake-0:
## Summary

Closes the interpreted-engine perf gap the four-way `tiers` bench surfaced (#574): next-interpreted was ~40% **slower** than classic on the wide `fanout` workload.

**Root cause.** The sparse dispatch (`interp.rs`, `drain_cycle`) drained each cycle's work set through a `BinaryHeap<Reverse<(layer, index)>>` — an O(N log N) per-node push/pop. On a *dense* graph where every node fires every cycle, that heap overhead (log-N comparisons + cache-unfriendly reordering, per node, per cycle) is pure loss versus classic's layer-bucketed vectors (`dirty_nodes_by_layer`, O(N)).

**Fix.** Replace the heap with the same layer-bucket structure: `buckets[layer] -> Vec<node>`, drained in ascending layer order (each bucket sorted by index to reproduce classic's within-layer wiring order). Every active downstream — and every `demux` mark — sits at a strictly higher layer, so a bucket never grows once its layer is reached; draining it fully is safe and each node fires exactly once after everything it reads. `max_layer` bounds the scan to `0..=max_layer`, keeping it **O(active + depth)** — the sparse property is preserved, with no per-node heap operations.

## Results

`cargo bench --bench tiers` (relative; numbers move with hardware):

| workload | classic | next-interpreted before | next-interpreted after |
|---|---|---|---|
| `fanout` | ~20 ms | ~36 ms (✗ ~40% slower) | **~17 ms** (✓ faster) |
| `dense_chain` | ~9 ms | ~6 ms | ~6 ms (unchanged) |
| `accumulate` | ~2.5 ms | ~1.7 ms | ~1.7 ms (unchanged) |

`next-interpreted ≥ classic-interpreted` now holds on **all three** workloads. compiled/nested continue to dominate (compiled fan-out ~25× either interpreter).

## Correctness

The output must stay byte-identical across `Sparse` / `FullSweep` / `compiled` / `nested` — the engine's whole differential-testing premise. All of it is green:
- full default-feature test suite ✅
- 285 `--features dynamic-graph` tests ✅ (exercises `run_dynamic`, the bucket-regrow path as nodes are spliced in, and `demux` same-cycle marks)

`cargo fmt --all`, `cargo lint`, `cargo lint-all` — clean.

## Notes

- One characteristic change: the bucket drain scans `0..=max_layer` (O(depth)) where the heap was purely O(active). Depth is tiny in practice (fanout ≈ 12, dense_chain ≈ 37) and `max_layer` tracks only layers that actually became active, so this is a clear net win.
- Updates the `tiers` bench module doc and `port-plan.md` (the standing perf item is now resolved).

---
_Generated by [Claude Code](https://claude.ai/code/session_01MatEkmguN4pYzgyMpSse3P)_

### DIFF
--- a/next/crates/wingfoil-next/benches/tiers.rs
+++ b/next/crates/wingfoil-next/benches/tiers.rs
@@ -34,14 +34,15 @@
 //! slower); running `cargo bench --bench tiers` puts `classic` and `interpreted`
 //! side by side per workload so that relationship is directly readable.
 //!
-//! Measured relationship at the time of writing (relative, not absolute — the
-//! numbers move with hardware): next-interpreted *beats* classic on the
-//! dispatch-bound `dense_chain` and the loop-bound `accumulate`, but *trails* it
-//! on the wide `fanout` (every node fires every cycle, so the sparse dirty-list
-//! path can't help). The compiled/nested tiers win decisively across the board
-//! — the compiled fan-out runs ~25x faster than either interpreter. The `fanout`
-//! interpreted gap is a standing perf item, tracked in `docs/port-plan.md`; this
-//! bench is the scaffold that keeps it honest.
+//! Measured relationship (relative, not absolute — the numbers move with
+//! hardware): next-interpreted **meets or beats** classic on all three
+//! workloads — the dispatch-bound `dense_chain`, the loop-bound `accumulate`,
+//! and the wide `fanout` (every node fires every cycle). The compiled/nested
+//! tiers win decisively across the board — the compiled fan-out runs ~25x faster
+//! than either interpreter. (An earlier `fanout` gap where next-interpreted
+//! trailed classic ~40% was the sparse dispatch's per-node `BinaryHeap`
+//! push/pop; replacing it with classic's layer-bucketed drain closed it. This
+//! bench is the scaffold that keeps the relationship honest.)
 
 use std::rc::Rc;
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


0-jake-0:
**CI status: red on flaky testcontainer integration tests, unrelated to this change.**

This PR only changes the interpreted engine's per-cycle dispatch (`interp.rs`) — replacing the `BinaryHeap` work-queue with classic's layer-bucketed drain. Correctness of that is what the engine differential suite checks, and it passes cleanly on **every** run here:

- `dynamic_graph` (14/14 — `demux`, `dynamic_group`, `add_upstream`, the `run_dynamic` + bucket-regrow + same-cycle marks paths)
- `engine_semantics`, `custom_op`, `custom_node`, `compiled_parity`, `compiled_stateful_ops` — the byte-identical `Sparse`/`FullSweep`/`compiled`/`nested` checks

The only failures are container-timing flakes in the bundled integration suites, and they differ run-to-run:

- Run 1 (`4b882b0`): `etcd_integration::test_sub_no_race_between_snapshot_and_watch` — "concurrent key missing" (etcd snapshot/watch race)
- Run 2 (`d86ff81`): `kafka_integration::{test_pub_round_trip, test_pub_multiple_records_in_burst}` — `left: 0` (published records not read back in the window)

Neither touches the dispatch path, and both are pre-existing flakes seen on other `next` PRs. A dispatch-only change cannot affect Kafka/etcd delivery timing.

I can't re-run individual jobs with the PR token, and re-triggering hits a different container flake each time. Recommend re-running just the **Build, Test, & Lint** job (or merging) — the change itself is validated by the engine suite. Happy to dig in if a *dispatch* test ever goes red.

---

PR #589, opened 2026-07-27
feat: add drop_small_change operator

### LINKED ISSUE #126: Add drop_small_change operator to StreamOperators
## Summary

Add a `drop_small_change` method to `StreamOperators` that suppresses ticks when the
change from the last emitted value is considered "small" by a caller-supplied closure.

This is the generalisation of `distinct` — where `distinct` uses `PartialEq` to drop
unchanged values, `drop_small_change` accepts any predicate over `(current, previous)`.

## Proposed API

```rust
/// Only propagates ticks where the change from the last emitted value is
/// considered significant by `is_small`.
///
/// `is_small(current, previous)` should return `true` to suppress the tick
/// (change too small to care about) or `false` to let it through.
///
/// The first upstream tick always propagates — there is no previous value to
/// compare against.
///
/// # Example
/// \`\`\`rust
/// // Drop price updates smaller than 0.01
/// price.drop_small_change(|curr, prev| (curr - prev).abs() < 0.01)
/// \`\`\`
fn drop_small_change(
    self: &Rc<Self>,
    is_small: impl Fn(&T, &T) -> bool + 'static,
) -> Rc<dyn Stream<T>>;
```

## Implementation

Add `src/nodes/drop_small_change.rs` following the pattern of `distinct.rs`:

```rust
pub(crate) struct DropSmallChangeStream<T: Element> {
    upstream: Rc<dyn Stream<T>>,
    is_small: Box<dyn Fn(&T, &T) -> bool>,
    value: T,
    last_ticked: Option<T>,
}

impl<T: Element> MutableNode for DropSmallChangeStream<T> {
    fn cycle(&mut self, _state: &mut GraphState) -> anyhow::Result<bool> {
        let curr = self.upstream.peek_value();
        let ticked = match &self.last_ticked {
            None => true,
            Some(prev) => !(self.is_small)(&curr, prev),
        };
        if ticked {
            self.last_ticked = Some(curr.clone());
            self.value = curr;
        }
        Ok(ticked)
    }

    fn upstreams(&self) -> UpStreams {
        UpStreams::new(vec![self.upstream.clone().as_node()], vec![])
    }
}
```

Wire up in `mod.rs` and expose via `StreamOperators`.

## Relationship to `distinct`

`distinct` is a special case:
```rust
stream.drop_small_change(|curr, prev| curr == prev)
// equivalent to:
stream.distinct()
```

`distinct` should be kept as a named convenience, but its implementation could delegate
to `drop_small_change` internally.

## Python bindings

Add to `wingfoil-python/src/py_stream.rs`, following the same pattern as `filter`:

```rust
/// Only propagates ticks where is_small(current, previous) returns False.
/// The first tick always propagates.
fn drop_small_change(&self, is_small: Py<PyAny>) -> PyStream {
    PyStream(self.0.drop_small_change(move |curr, prev| {
        Python::attach(|py| {
            is_small
                .call1(py, (curr.value(), prev.value()))
                .unwrap()
                .extract::<bool>(py)
                .unwrap()
        })
    }))
}
```

Python usage:
```python
# Drop price updates smaller than 0.01
price.drop_small_change(lambda curr, prev: abs(curr - prev) < 0.01)
```

Add a pytest test in `wingfoil-python/tests/`.

## Files to touch

- `wingfoil/src/nodes/drop_small_change.rs` — new node
- `wingfoil/src/nodes/mod.rs` — declare module, add `StreamOperators::drop_small_change`
- `wingfoil/src/nodes/distinct.rs` — optionally delegate to `drop_small_change`
- `wingfoil-python/src/py_stream.rs` — Python binding
- `wingfoil-python/tests/` — pytest test

## Tests

- First tick always propagates
- Subsequent tick suppressed when `is_small` returns `true`
- Subsequent tick propagated when `is_small` returns `false`
- Value seen by downstream is the last non-suppressed upstream value
- Works with `f64` threshold predicate (the primary use case)
- Python: callable receives current and previous values; suppresses correctly

XiaoXKKK:
## Summary

- Add `StreamOperators::drop_small_change` for suppressing changes considered insignificant by a caller-provided predicate.
- Compare each current value against the last emitted value, rather than the last observed value.
- Always propagate the first upstream tick.
- Expose the operator through the Python bindings with Python exception propagation.
- Add Rust and Python tests covering threshold behavior, callback arguments, and error handling.

## Testing

- `cargo clippy -p wingfoil --lib -- -D warnings`
- `cargo test -p wingfoil drop_small_change`
- `cargo test -p wingfoil --doc`

Closes #126

### DIFF
--- a/wingfoil-python/src/py_stream.rs
+++ b/wingfoil-python/src/py_stream.rs
@@ -8,6 +8,7 @@ use ::wingfoil::{Element, IntoStream, NodeOperators, Stream, StreamOperators};
 use pyo3::conversion::IntoPyObject;
 use pyo3::prelude::*;
 
+use std::cell::RefCell;
 use std::rc::Rc;
 
 use crate::proxy_stream::*;
@@ -273,6 +274,39 @@ impl PyStream {
     /// only propagates its source if it changed (uses PartialEq on PyElement)
     fn distinct(&self) -> PyStream {
         PyStream(self.0.distinct())
+    }
+
+    /// Only propagates ticks where is_small(current, previous) returns False.
+    /// The first tick always propagates.
+    fn drop_small_change(&self, is_small: Py<PyAny>) -> PyStream {
+        let last_emitted = Rc::new(RefCell::new(None::<PyElement>));
+        let evaluated = self.0.try_map({
+            let last_emitted = Rc::clone(&last_emitted);
+            move |current: PyElement| {
+                let previous = last_emitted.borrow().clone();
+                let should_suppress = match previous {
+                    None => false,
+                    Some(previous) => Python::attach(|py| {
+                        let result = is_small
+                            .call1(py, (current.value(), previous.value()))
+                            .map_err(py_callback_error)?;
+                        result.extract::<bool>(py).map_err(|e| {
+                            py_callback_error(e)
+                                .context("drop_small_change predicate must return a bool")
+                        })
+                    })?,
+                };
+
+                if should_suppress {
+                    Ok(None)
+                } else {
+                    *last_emitted.borrow_mut() = Some(current.clone());
+                    Ok(Some(current))
+                }
+            }
+        });
+
+        PyStream(evaluated.filter_map(|value| value))
     }
 
     /// drops source contingent on supplied predicate (Python callable)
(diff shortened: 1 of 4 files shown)


0-jake-0:
Thanks!

PR #629, opened 2026-08-01
Bind the aeron adapter, the first opt-in-only binding

0-jake-0:
Stacked on #628 (web) — merge that first.

## Surface

| Python | Rust | shape |
|---|---|---|
| `aeron_sub(graph, …)` | `aeron_sub_fragment` | subscription → `list[bytes]` |
| `aeron_sub_with_status(graph, …)` | `aeron_sub_fragment_with_status` | the same, plus a status stream |
| `aeron_pub(stream, …)` | `AeronSinkOps::aeron_pub` | publication sink |
| `aeron_pub_with_status(stream, …)` | `AeronSinkOps::aeron_pub_with_status` | the same, plus a status stream |

## Kept out of both roll-ups — deliberately

`rusteron-client` builds the Aeron **C** library from source (clang, libuuid, CMake ≥ 3.30). So this is the first binding in **neither** `all-adapters` **nor** the wheel's default feature set:

- joining `all-adapters` would break `next-python-test.yml`, which installs only protoc and patchelf;
- shipping it in the wheel would make the published artifact un-buildable for everyone else — the portability criterion the pyproject comment already states.

`maturin develop -F aeron` opts in. The consequence worth flagging: **the Rust `#[cfg(test)]` tests in this module do not run in the default job** — `aeron-next-integration.yml` is their only home, and its new Python leg is where both tiers run.

## Deviations from legacy `py_aeron.rs`

1. **`mode` is a string** (`"spin"` / `"threaded"`), not an `AeronMode` `#[pyclass]` enum — one fewer class to register, and an unknown value raises listing what is accepted.
2. **Publishing fails loudly.** Legacy's `extract::<Vec<u8>>(py).unwrap_or_else(|e| { log::error!(…); Vec::new() })` published an **empty Aeron message** when the value was not bytes — a real message on the wire, carrying nothing.
3. **New capabilities**: the `_with_status` pair (legacy bound neither), `fragment_limit`, and multi-message publish from one tick — legacy wrapped each value in a one-element burst.

## Validate before connecting

Aeron resolves `(channel, stream_id)` through the media driver at *wiring*, so a mistake in the arguments would otherwise surface as `driver timeout` after a failed connect. Every check — mode, `fragment_limit`, `timeout_secs`, and the historical-run rejection the engine also makes later — now runs first. The three `timeout_secs` tests took 30 s and reported a driver error before this; they now pass instantly with the right message.

## Testing

- **6 Rust tests** — the status→string mapping, both selectors, the options defaults, the timeout guard, the early historical rejection, and the fragment parser.
- **8 pytest cases with no driver** — everything reachable before the connect.
- **5 `requires_aeron` round trips through a real media driver**: publish→subscribe over `aeron:ipc` in **both** poll modes, a multi-message tick, both status streams reaching `"connected"`, and the non-bytes abort.

All five ran **locally**, not CI-first: I installed CMake 3.31 + libbsd in the sandbox, built `aeronmd` from the same rusteron sources the crate vendors, and ran the tier against it — 5 passed. The CI leg starts the same driver from the workflow's existing image, with the `/dev/shm` bind mount and uid mapping `tests/aeron_integration.rs` documents.

One CI-specific detail: the extension links `libaeron.so` out of the cargo build directory, which is not on the loader path, so the pytest step sets `LD_LIBRARY_PATH` — without it `import wingfoil_next` itself fails.

Local: `cargo fmt`, `cargo lint`, `cargo clippy -p wingfoil-next-python --features all-adapters,aeron --all-targets`, `cargo test … --features all-adapters,aeron` (178 lib tests), `pytest` (234 passed, 10 skipped, 38 deselected).

**Remaining after this: 1** — iceoryx2.

https://claude.ai/code/session_01Vgapk3HgHd5JezDGjicseq

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vgapk3HgHd5JezDGjicseq)_


0-jake-0:
The first CI run of the new aeron Python leg failed to build, and it turned up a latent trap worth recording — the fix is now in this PR.

**What happened.** `maturin develop -F aeron` **replaces** the `pyproject.toml` feature list rather than adding to it, so the build ran with `aeron` and nothing else. `adapters::common` is compiled for *any* adapter and names `async_source::RunParams`, but the engine's `aeron` feature does not imply `async` — so it failed with `unresolved import wingfoil_next::async_source`.

**Why it was invisible.** Every local check and every existing CI job builds `all-adapters`, where some *other* adapter supplies `async`. The roll-up hides a missing implication entirely. Checking each feature on its own found the same latent breakage in **three already-merged bindings**:

```
zmq    FAIL      augurs FAIL      fix    FAIL
```

**The fix.** Rather than sprinkle `wingfoil-next/async` onto the affected features, this adds an internal `_common` feature that every adapter feature lists — a single place expressing what `adapters::common` needs:

```toml
_common = ["wingfoil-next/async"]
zmq     = ["wingfoil-next/zmq", "_common"]
```

All 14 adapter features now build standalone:

```
postgres ok   kafka ok   redis ok   etcd ok   fluvio ok   csv ok   zmq ok
otlp ok   augurs ok   kdb ok   fix ok   prometheus ok   web ok   aeron ok
```

The CI leg also now names `-F extension-module,aeron`, since `-F` replacing the list dropped `extension-module` too.

`/bind-adapter-next` gained both rules: list `_common` in a new feature, and verify with `cargo check -p wingfoil-next-python --features <name>` rather than trusting the roll-up.

---

PR #661, opened 2026-08-02
Compiled islands: nest a graph! sub-graph as one node of an interpreted graph

0-jake-0:
A graph! wiring fn can now take input streams after the builder
(`fn ema(g: &GraphBuilder, price: &Stream<f64>) -> Stream<f64>`), and
every single-output graph gains a `nested(g, inputs...)` expansion that
mounts the whole sub-graph as a single composite node of an interpreted
graph under construction. Inside the composite one closure owns all
inner state and runs the same monomorphized straight-line dispatch
`compiled()` emits; the outer engine pays one dyn call per activation
for the entire island.

Ctx gains a nested mode backing this: an inner op's schedule lands in
the composite's private TimeQueue keyed by inner node index instead of
the outer kernel. The composite demultiplexes due entries into its
inner dirty flags each activation and forwards only the earliest
pending time outward, so inner tickers and delays run on the outer
kernel's clock with no drift. Passive inputs (sample's data edge) are
excluded from the composite's active upstreams and cannot activate it.

Graphs with inputs expand to wire + nested only (they cannot run
standalone) and must have exactly one output; source-only graphs keep
all four expansions. Stream-ref arguments now also accept a bare input
parameter name (`data.sample(trigger)`) since inputs are already
references. Tests cross-check each island against identical flat
wiring in the same graph — same kernel, same cycles, exact agreement —
covering input islands, inner delay scheduling, source islands, and
passive-input gating.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01RSP1oHTaGBeMmnGmYa5XnS


0-jake-0:
Closing as vestigial — everything this branch adds is already on `next`.

Checked file by file rather than by diff:

| This branch adds | On `next` |
|---|---|
| `scripts/bench-report.sh` | present |
| `crates/wingfoil-next/benches/plot_tiers.py` | present at `crates/wingfoil/benches/plot_tiers.py` |
| `benches/images/**` (tiers, ops, graph, lscpu) | present — 25 files, including the full `tiers/` set (`accumulate`, `dense_chain`, `fan_in_16/64/256`, `fanout`, `sparse`, `sparse_wide`, `summary.png`) |

The content landed on `next` and was then moved by the cutover rename (`36596a2` "Rename the crates and the Python module", `c7a982b` "Drop *next* as a name for this engine"), so this branch is stale twice over: its payload is already merged, and it writes to the pre-rename `crates/wingfoil-next/` path that no longer exists.

Two incidental notes for the record:

- The base is also wrong — this is `next` work opened against `main`, which is why GitHub renders it as ~621 files. That figure is an artifact of the merge-base predating "Invert the tree", not a real diff.
- The title ("Compiled islands: nest a graph! …") never matched the contents; #594 and #579 carried the same stale title.

---

PR #707, opened 2026-08-04
Add market data adapter: venue-neutral order book vocabulary

0-jake-0:
## Summary

Introduces the `market` adapter — a venue-neutral vocabulary layer for market data that all venue adapters (Binance, Coinbase, Databento, etc.) normalize into. This enables graphs wired against one venue to run unchanged against another without venue-specific dependencies in this crate.

The adapter provides:
- **Fixed-point price/quantity types** (`Px`, `Qty`) with exact decimal parsing and ordering semantics
- **Normalized event types** (`Trade`, `BookSnapshot`, `BookDelta`, `BookUpdate`, `MarketEvent`)
- **Sequence gap detection** across two venue numbering schemes (`Sequencing::Single` and `Sequencing::Span`)
- **Order book state machine** (`OrderBook`) with snapshot/delta maintenance, gap detection, and pre-snapshot delta buffering
- **Graph ops** (`order_book`, `order_book_bursts`) to maintain books on the graph

## Key Changes

- **`src/adapters/market.rs`** (1402 lines): Complete market data adapter implementation
  - `parse_fixed`/`fmt_fixed`: Hand-rolled decimal parsing (no `f64` round-trip) to preserve exact book keys
  - `Px`/`Qty` macros: Fixed-point types with `Ord + Eq + Hash` for keying book levels
  - `InstrumentId`, `Side`, `Level`, `LevelChange`: Value types for book representation
  - `Sequencing`: Normalizes two venue numbering schemes (single counter vs. span)
  - `Trade`, `BookSnapshot`, `BookDelta`, `BookUpdate`, `MarketEvent`: Normalized event types
  - `OrderBook`: State machine with `BTreeMap`-backed bid/ask sides, gap detection, and buffering
  - `OrderBookOp`/`OrderBookBurstOp`: Graph ops maintaining books from `Stream<BookUpdate>` and `Stream<Burst<BookUpdate>>`
  - Comprehensive module documentation pinning four non-negotiable adapter contracts

- **`tests/market_adapter.rs`** (297 lines): Integration tests on real graphs
  - Tick timing and burst handling
  - Gap detection and recovery
  - Pre-snapshot delta buffering and replay
  - Mixed-instrument wiring validation

- **`src/adapters/mod.rs`**: Added market adapter to module documentation

- **`src/adapters/CLAUDE.md`**: Added market adapter entry to adapter table

- **`Cargo.toml`**: Added `market` feature (no external dependencies)

- **`src/adapters/market/CLAUDE.md`**: Design documentation covering the four adapter contracts and implementation shapes

## Notable Implementation Details

1. **Exact decimal parsing**: `Px::parse` rejects exponent notation, overflow, and fractional precision finer than 9 decimal places rather than silently truncating — venues with finer tick sizes fail loudly at the adapter boundary.

2. **Two timestamps with different semantics**: `venue_time` (venue's clock, optional, never trusted for ordering) vs. `recv_time` (engine time from `Ctx::time()`, deterministic for replay). Adapters must never stamp `recv_time` from `NanoTime::now()`.

3. **Gap signalling, not papering over**: `OrderBook::apply` clears the book and moves to `BookStatus::Gapped` on sequence discontinuity. The op still **ticks** on a gap so downstream learns the book is invalid; `best_bid`/`best_ask`/`mid` return `None` while gapped.

4. **Pre-snapshot delta buffering**: Deltas arriving before the snapshot are buffered (capped at `MAX_BUFFERED_DELTAS = 16_384`), then replayed on snapshot. Dropping them would leave the book missing its first few updates.

5. **Copy-on-write book sharing**: `OrderBook` is held in `Arc` and mutated via `Arc::make_mut`, so cycles that don't retain the book mutate in place while downstreams get cheap snapshots.

Venue adapters live in separate crates (not in this tree) and depend on `wingfoil = { features = ["market"] }`, keeping transport dependencies

https://claude.ai/code/session_01K4uoFFyNGkHU43yV83igkc

### DIFF
--- a/crates/wingfoil/src/adapters/market.rs
+++ b/crates/wingfoil/src/adapters/market.rs
@@ -1,0 +1,1402 @@
+//! Market data adapter — the **venue-neutral vocabulary** every market data
+//! adapter produces, plus the order book state machine that consumes it.
+//!
+//! Unlike a messaging adapter, this one connects to nothing: it is the shared
+//! type layer that venue adapters (Binance, Coinbase, Databento, …) normalise
+//! *into*, so that a graph wired against one venue runs unchanged against
+//! another. Venue adapters are **separate crates** rather than modules here:
+//! each carries its own transport dependencies and release cadence, and wiring
+//! one in costs this crate nothing. What lives here is the part they must all
+//! agree on:
+//!
+//! - [`Px`] / [`Qty`] — fixed-point price and quantity, exact and orderable.
+//! - [`InstrumentId`], [`Side`], [`Level`], [`LevelChange`] — the value types.
+//! - [`Trade`], [`BookSnapshot`], [`BookDelta`], [`BookUpdate`],
+//!   [`MarketEvent`] — the events an adapter emits.
+//! - [`Sequencing`] — how a venue numbers its updates, normalised across the
+//!   two shapes venues actually use.
+//! - [`OrderBook`] — snapshot/delta book maintenance with gap detection, and
+//!   [`MarketBookOps::order_book`], the op that maintains one on the graph.
+//!
+//! # Layering
+//!
+//! Following the [`stats`](crate::stats) module's pattern, the ops are *not*
+//! in the [`prelude`](crate::prelude): bring the extension trait in explicitly
+//! with `use wingfoil::adapters::market::MarketBookOps;`.
+//!
+//! # The four decisions an adapter must not make for itself
+//!
+//! These are the semantics that make two independently written adapters
+//! interchangeable. An adapter that deviates is broken even if it compiles.
+//!
+//! **1. Prices and quantities are fixed-point, parsed from the venue's own
+//! decimal text.** Venues send `"43210.10000000"` as a JSON *string*. Parse it
+//! with [`Px::parse`], never via `f64`: a book keys levels *by price*, so a
+//! delete arrives as "remove the level at 43210.10" and must match the key
+//! exactly. Round-tripping through binary floating point is how that silently
+//! stops matching. [`Px`] and [`Qty`] are `Ord + Eq + Hash` precisely because
+//! `f64` is not.
+//!
+//! **2. Both timestamps are recorded, and they mean different things.**
+//! [`venue_time`](BookSnapshot::venue_time) is the venue's own clock — absent
+//! on venues that do not send one, and never trusted for ordering across
+//! venues. [`recv_time`](BookSnapshot::recv_time) is engine time when the
+//! adapter received the message, which is what the graph replays on. An
+//! adapter sets `recv_time` from [`Ctx::time`](crate::op::Ctx::time), *not*
+//! from [`NanoTime::now`] — that is what keeps a recorded session replayable.
+//!
+//! **3. A gap is signalled, never papered over.** When [`OrderBook::apply`]
+//! detects a sequence discontinuity it clears the book, moves to
+//! [`BookStatus::Gapped`] and reports [`BookApply::Gap`]. It does *not* keep
+//! applying deltas to a book it knows is wrong. The adapter's obligation is to
+//! notice and re-request a snapshot; downstream's obligation is to stop
+//! trusting the book — which is why [`best_bid`](OrderBook::best_bid) and
+//! friends return `None` while gapped, and why the op still ticks on a gap
+//! rather than going quiet.
+//!
+//! **4. Deltas that arrive before the snapshot are buffered, not dropped.**
+//! Every venue that serves a book as "REST snapshot + WebSocket deltas" has a
+//! race: the stream is subscribed first, and its early messages predate the
+//! snapshot. [`OrderBook`] buffers them ([`BookApply::Buffered`]), then on
+//! snapshot discards the ones the snapshot already covers and replays the
+//! rest. Dropping them instead leaves a book that is quietly missing its first
+//! few updates.
+//!
+//! # Example
+//!
+//! Driving a book directly — which is also how an adapter's own tests should
+//! exercise its normalisation, with no graph involved:
+//!
+//! ```
+//! use wingfoil::NanoTime;
+//! use wingfoil::adapters::market::{
+//!     BookApply, BookDelta, BookSnapshot, BookStatus, BookUpdate, InstrumentId,
+//!     Level, LevelChange, OrderBook, Px, Qty, Sequencing, Side,
+//! };
+//!
+//! let inst = InstrumentId::new("example", "BTC-USD");
+//! let mut book = OrderBook::new(inst.clone());
+//!
+//! book.apply(&BookUpdate::Snapshot(BookSnapshot {
+//!     instrument: inst.clone(),
+//!     bids: vec![Level::new(Px::parse("100.5")?, Qty::parse("2")?)],
+//!     asks: vec![Level::new(Px::parse("101.0")?, Qty::parse("3")?)],
+//!     sequencing: Sequencing::Single(7),
+//!     venue_time: None,
+//!     recv_time: NanoTime::ZERO,
+//! }));
+//! assert_eq!(book.status(), BookStatus::Live);
+//! assert_eq!(book.mid(), Some(100.75));
+//!
+//! // A better bid takes the touch.
+//! let outcome = book.apply(&BookUpdate::Delta(BookDelta {
+//!     instrument: inst.clone(),
+//!     changes: vec![LevelChange::new(Side::Bid, Px::parse("100.75")?, Qty::parse("1")?)],
+//!     sequencing: Sequencing::Single(8),
+//!     venue_time: None,
+//!     recv_time: NanoTime::ZERO,
+//! }));
+//! assert_eq!(outcome, BookApply::Applied);
+//! assert_eq!(book.best_bid().unwrap().price, Px::parse("100.75")?);
+//! # Ok::<(), anyhow::Error>(())
+//! ```
+//!
+//! On the graph, a `Stream<BookUpdate>` from a venue adapter becomes a stream
+//! of maintained books with [`MarketBookOps::order_book`]:
+//!
+//! ```ignore
+//! use wingfoil::adapters::market::{MarketBookOps, MarketEventOps};
+//!
+//! let books = venue_events.book_updates().order_book();
+//! let mids = books.map_filter(|b| b.mid());
+//! ```
+
+use std::collections::BTreeMap;
+use std::fmt;
+use std::sync::Arc;
+
+use anyhow::{Result, anyhow, bail};
+
+use crate::Burst;
+use crate::fluent::Stream;
+use crate::op::{Activation, Ctx, Op, Tick};
+use crate::runtime::time::NanoTime;
+use wingfoil_derive::op;
+
+// -------------------------------------------------------------------------
+// Fixed-point price and quantity.
+// -------------------------------------------------------------------------
+
+/// Number of decimal places [`Px`] and [`Qty`] represent exactly.
+pub const DECIMALS: u32 = 9;
+
+/// The integer scale factor behind [`Px`] and [`Qty`]: `10^DECIMALS`.
+pub const SCALE: i64 = 1_000_000_000;
+
+/// Parse a plain decimal string (`-?digits[.digits]`) into a scaled integer.
+///
+/// Deliberately hand-rolled rather than routed through `f64`: the whole point
+/// is to avoid a binary floating point round trip. Rejects exponent notation
+/// and any fractional precision that would be lost, so a venue whose tick size
+/// is finer than [`DECIMALS`] fails loudly at the adapter boundary instead of
+/// silently corrupting book keys.
+fn parse_fixed(s: &str) -> Result<i64> {
+    let t = s.trim();
+    if t.is_empty() {
+        bail!("empty decimal string");
+    }
+    let (neg, digits) = match t.as_bytes()[0] {
+        b'-' => (true, &t[1..]),
+        b'+' => (false, &t[1..]),
+        _ => (false, t),
+    };
+    if digits.is_empty() {
+        bail!("decimal string {s:?} has a sign but no digits");
+    }
+    let (int_part, frac_part) = match digits.split_once('.') {
+        Some((i, f)) => (i, f),
+        None => (digits, ""),
+    };
+    // An empty integer part is fine (".5"), but the remaining text must be
+    // digits only — this is what rejects "1e-9", "1_000" and "NaN".
+    let all_digits = |p: &str| p.bytes().all(|b| b.is_ascii_digit());
+    if !all_digits(int_part) || !all_digits(frac_part) {
+        bail!("decimal string {s:?} is not a plain decimal number");
+    }
+    if int_part.is_empty() && frac_part.is_empty() {
+        bail!("decimal string {s:?} has no digits");
+    }
+
+    let mut value: i64 = 0;
+    for b in int_part.bytes() {
+        value = value
+            .checked_mul(10)
+            .and_then(|v| v.checked_add((b - b'0') as i64))
+            .ok_or_else(|| anyhow!("decimal string {s:?} overflows the fixed-point range"))?;
+    }
+    value = value
+        .checked_mul(SCALE)
+        .ok_or_else(|| anyhow!("decimal string {s:?} overflows the fixed-point range"))?;
+
+    let mut scale = SCALE;
+    for (i, b) in frac_part.bytes().enumerate() {
+        let digit = (b - b'0') as i64;
+        if i < DECIMALS as usize {
+            scale /= 10;
+            value = value
+                .checked_add(digit * scale)
+                .ok_or_else(|| anyhow!("decimal string {s:?} overflows the fixed-point range"))?;
+        } else if digit != 0 {
+            bail!(
+                "decimal string {s:?} has more than {DECIMALS} decimal places of \
+                 precision, which the fixed-point representation cannot hold \
+                 exactly; the venue's tick size is finer than this adapter layer \
+                 supports"
+            );
+        }
+    }
+    Ok(if neg { -value } else { value })
+}
+
+/// Render a scaled integer back to a plain decimal string, trimming trailing
+/// fractional zeros (but never the whole fraction: `1.0` prints as `1`).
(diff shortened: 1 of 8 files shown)


crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+/// wingfoil::adapters::market::MarketEventOps;`.
+pub trait MarketEventOps {
+    /// Just the trade prints.
+    fn trades(&self) -> Stream<Trade>;
+    /// Just the book messages — wire this into
+    /// [`order_book`](MarketBookOps::order_book).
+    fn book_updates(&self) -> Stream<BookUpdate>;
+}
+
+impl MarketEventOps for Stream<MarketEvent> {
```
0-jake-0, line 1405: **`MarketEventOps` has no burst impl, so the documented wiring doesn't compile.**

`GraphBuilder::channel` / `external` / `spawn` all return `Stream<Burst<T>>` — and `market/CLAUDE.md` says so itself: *"The burst impl is the one a real adapter hits, since `channel`/`external` sources produce bursts."*

`MarketBookOps` correctly covers both shapes. `MarketEventOps` is implemented **only** for `Stream<MarketEvent>`, so the pattern in the module docs —

```rust
let books = venue_events.book_updates().order_book();
```

— has no impl for the `Stream<Burst<MarketEvent>>` a real adapter actually holds, and there's no burst-preserving `map_filter` in `fluent.rs` to bridge it. That's presumably why the doc example at line 107 is ` ```ignore ` and therefore never compile-checked.

Suggest adding `impl MarketEventOps for Stream<Burst<MarketEvent>>` producing `Stream<Burst<Trade>>` / `Stream<Burst<BookUpdate>>`, and turning the doc example into a real doctest against the burst shape.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+use crate::op::{Activation, Ctx, Op, Tick};
+use crate::runtime::time::NanoTime;
+use wingfoil_derive::op;
+
+// -------------------------------------------------------------------------
+// Fixed-point price and quantity.
+// -------------------------------------------------------------------------
+
+/// Number of decimal places [`Px`] and [`Qty`] represent exactly.
+pub const DECIMALS: u32 = 9;
```
0-jake-0, line 191: **`Qty`'s range rejects real crypto quantities.**

`DECIMALS = 9` in an `i64` caps both types at **±9,223,372,036.854775807**. Verified locally:

```
Qty::parse("10000000000")  -> Err("... overflows the fixed-point range")
Qty::parse("9223372036")   -> Ok
```

Book levels of 10 billion+ units are routine for SHIB / PEPE / BONK, so a venue adapter for those pairs can't be written against this layer at all — and by design it fails hard at the boundary rather than degrading.

The root cause is that `Px` and `Qty` share one scale, but they want opposite things: price wants precision at modest magnitude, quantity wants magnitude at modest precision. Options, cheapest first:

- Give `Qty` its own (smaller) `DECIMALS` — the `fixed_point!` macro already takes a name, so parameterise the scale.
- Store `i128` and keep 9 dp for both.
- Keep the newtype API and delegate to `rust_decimal::Decimal` internally.

On that last point, more broadly: `rust_decimal` already gives exact `FromStr` with no `f64` path, `Ord + Eq + Hash`, `Copy`, no allocation, and a 96-bit mantissa that makes this whole issue disappear. The hand-rolled parser buys zero deps and an 8-byte integer compare in the `BTreeMap`; it costs ~60 lines of parser to own plus this range ceiling. Given the module's thesis is *"an adapter that deviates is broken even if it compiles"*, the parser is the one place I'd lean on something battle-tested. Keeping the `Px`/`Qty` newtypes (the type-level price/quantity distinction is worth having, and the API surface is right) while swapping the backing store is a small internal change with no API break.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+            /// Parse the venue's decimal text exactly, with no `f64` round trip.
+            pub fn parse(s: &str) -> Result<Self> {
+                parse_fixed(s).map(Self)
+            }
+
+            /// Round an `f64` into fixed point. Lossy by nature — prefer
+            /// [`parse`](Self::parse) on anything that came off a wire.
+            pub fn from_f64(v: f64) -> Self {
+                Self((v * SCALE as f64).round() as i64)
+            }
```
0-jake-0, line 359: **`from_f64` silently corrupts, in a module whose thesis is "fail loudly".**

Verified locally:

```
Px::from_f64(f64::NAN)  ->  0
Px::from_f64(1e30)      ->  9223372036854775807   // saturating `as` cast
```

`Px::parse` goes to real lengths to reject what it can't represent; `from_f64` on the same type saturates and NaN-to-zeroes without a word. It's documented as "lossy … for tests and display code", but *lossy* and *silently wrong* aren't the same claim, and this is a public `pub fn` on the type an adapter author reaches for.

Either make it `try_from_f64() -> Result<Self>`, or at minimum document the saturating/NaN behaviour on the method itself.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+/// engine requires every stream's value type to be `Default` for its
+/// pre-first-tick value slot. It is not a meaningful instrument.
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct InstrumentId {
+    /// The venue's short name, lowercase by convention (`"binance"`).
+    pub venue: Arc<str>,
+    /// The venue's own symbol, verbatim and *not* normalised — venues disagree
+    /// about separators and casing, and rewriting the symbol loses the ability
+    /// to echo it back on a subscribe.
+    pub symbol: Arc<str>,
```
0-jake-0: **This re-invents `Sym`, which already exists one directory over.**

`adapters/kdb.rs` already has the interned-symbol pattern:

```rust
pub struct Sym(Arc<str>);

#[derive(Default)]
pub struct SymbolInterner { set: HashSet<Arc<str>> }
```

`InstrumentId` is two bare `Arc<str>`s with no dedup — strictly less than what kdb does, in the module whose whole pitch is that it's the shared vocabulary layer. Two allocations and **two atomic refcount bumps per clone**, and the doc comment right here says the id is cloned into *every* message by design.

Suggestion: lift `Sym`/`SymbolInterner` into `adapters/common.rs` (already exists on `next`, always compiled, documented as the home for cross-adapter helpers), re-export from `kdb` for compatibility, and make this `InstrumentId { venue: Sym, symbol: Sym }`. Fixes the per-event allocation, costs nothing else.

One caveat worth being explicit about, since it bounds how far this goes: kdb's `Sym` derives `PartialEq`/`Hash`, so comparison and hashing are **content**-based. Interning there buys allocation dedup, not comparison cost — and that's deliberate, because the interner is `&mut` and per-read-call, and `read_cached` explicitly notes interning isn't restored on deserialisation. So `Arc::ptr_eq` would give false negatives and content equality is the only sound choice under that design.

Getting cheap *comparison* too (a `u32` handle, or ptr-eq) needs a single interner with a lifetime long enough that pointer identity is meaningful — which, with venue adapters as out-of-tree crates each holding their own state, means a process-global one. That's a real design commitment and probably shouldn't ride along in this PR. See also the note on `book_for`.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+                self.bids.insert(lvl.price, lvl.qty);
+            }
+        }
+        for lvl in &s.asks {
+            if !lvl.qty.is_zero() {
+                self.asks.insert(lvl.price, lvl.qty);
+            }
+        }
+        self.last_seq = s.sequencing.last();
+        self.status = BookStatus::Live;
```
0-jake-0, line 920: **Stale snapshots silently rewind the book.**

`apply_snapshot` unconditionally clears, applies, and sets `Live` — there's no `s.sequencing.last() <= self.last_seq` guard. A late or duplicate REST response arriving after recovery will roll the book backwards to an older image, and downstream sees `Live` with no indication anything went wrong.

Given how careful `apply_delta` is about staleness (the `last <= last_applied` check just below), the asymmetry looks unintentional rather than deliberate. If a snapshot regression is meant to be accepted — e.g. because a re-request after a gap can legitimately return an older image than the deltas already seen — that's worth a comment saying so, since it's the one place the module lets the book move backwards.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+                    // the overflow against whatever the message carried.
+                    let got = d.sequencing.first().unwrap_or(0);
+                    return BookApply::Gap { expected: got, got };
+                }
+                self.pending.push(d.clone());
+                BookApply::Buffered
+            }
+            // A gapped book refuses deltas outright: applying them would build
+            // on an image already known to be wrong.
+            BookStatus::Gapped => BookApply::Stale,
```
0-jake-0: **Two overloadings of `BookApply` that lose information the adapter needs.**

1. **Buffer overflow reports a nonsense pair.** `Gap { expected: got, got }` where `got = first().unwrap_or(0)` — `expected == got` isn't a gap by any reading, and `0` is a lie for an unsequenced venue. The comment acknowledges the awkwardness; a distinct `BookApply::Overflow` variant would just say what happened.

2. **`Stale` is doing double duty.** A `Gapped` book returns `Stale` for refused deltas, which is indistinguishable from "already covered by the image we hold". Those are opposite situations for an adapter: one is normal during the snapshot race and needs no action, the other means the book is broken and a snapshot must be re-requested. A `Refused` variant (or `Stale { gapped: bool }`) would let the adapter branch on it.

Both are cheap to fix now and become breaking changes once venue adapter crates are matching on `BookApply`.

---

crates/wingfoil/src/adapters/market.rs
```diff
@@ -0,0 +1,1402 @@
...
+) -> Result<&'s mut Arc<OrderBook>> {
+    let book = state.get_or_insert_with(|| Arc::new(OrderBook::new(update.instrument().clone())));
+    if book.instrument() != update.instrument() {
+        bail!(
+            "order_book received an update for {} on a book tracking {}; \
+             demultiplex by instrument before wiring order_book",
+            update.instrument(),
+            book.instrument()
+        );
+    }
```
0-jake-0, line 1219: **A construction-time invariant checked on every message.**

This runs two `str` comparisons per update, forever, purely to catch a graph-*wiring* mistake — and the doc comment says as much: *"A mixed stream is a wiring bug, not a runtime condition."* The wiring is fixed after the first update, so checking it a million times a second buys nothing the first check didn't.

The abort behaviour is right and the test pinning it (`mixed_instruments_abort_the_run`) is worth keeping. It's the per-message cost that's misplaced. Options: check only on book creation plus a cheap identity compare (see the `InstrumentId` note — with a shared `Sym` or a `u32` handle this becomes free), or gate the full check behind `debug_assert`.

---

0-jake-0 (commented):
## Review — venue-neutral market data adapter

**Verified locally on `b75ff11`:** 16 unit tests, 6 integration tests and 1 doctest all pass; `cargo clippy --features market --all-targets -- -D warnings` is clean. CI's `--all-features` lib+tests pass picks the new feature up automatically, so no workflow change is needed.

This is high-quality work. The module docs pin four contracts and the code actually implements all four — that's rarer than it should be. The `Sequencing::Span` handoff logic is correct for the Binance case, gap-during-replay is handled, and `gap_out()` clearing the book so `best_bid()` returns `None` is a genuinely good fail-safe: a downstream that forgot to check `status()` still can't quote off a stale image. The tests each pin a stated contract rather than restating the implementation, and `market/CLAUDE.md` is unusually good — the "What an out-of-tree venue adapter owes" checklist is exactly the right artifact for a contract layer.

Comments inline. The three I'd want addressed before merge:

1. **`MarketEventOps` has no burst impl**, so the wiring the module docs demonstrate doesn't compile against the `Stream<Burst<MarketEvent>>` that `channel`/`external`/`spawn` sources actually produce — which is the shape this PR's own `CLAUDE.md` says real adapters hit.
2. **`Qty`'s ±9.22e9 ceiling** rejects quantities that are routine on crypto venues (SHIB/PEPE/BONK), so an adapter for those pairs can't be written against this layer at all.
3. **`InstrumentId` re-invents `Sym`**, which already exists in `adapters/kdb.rs` — and does less, since it has no dedup at all.

### Smaller notes not worth their own thread

- **Gap sequence numbers are lost on the graph path.** The ops map `Gap { expected, got }` to a plain tick, so downstream can read `status() == Gapped` — enough to trigger a re-snapshot — but the ids that would go in a log line are gone.
- **`level_count` isn't gated on `is_live()`** unlike `best_bid`/`best_ask`/`depth`. Harmless today because `gap_out` clears both sides, but it'll bite if gap handling ever changes to retain levels.
- **`parse_fixed` is slightly more lenient than the docs suggest:** `"1."` → `1`, `"+.5"` → `0.5`, and leading/trailing whitespace is trimmed. All benign, but for a parser whose contract is "reject what you can't represent exactly", one line on what it *does* accept would help.
- **`MarketEventOps` constructs a discarded `Trade::default()` / `BookUpdate::default()` per non-matching event**, because `map_filter` demands a value in the false branch. Not this PR's fault, but on a multiplexed venue stream it's per-message work on the filtered path.

Docs, feature gating, `Activation::NONE`, and the `stats`-style out-of-prelude extension trait all match the conventions on `next`.

---

0-jake-0:
Thanks — this was a genuinely useful review. All ten points addressed in `2213c36`. Two of them I decided differently than suggested; those are called out below.

## The three blocking items

**1. `MarketEventOps` burst impl.** Correct, and the diagnosis of *why* it went unnoticed is the important part: the doc example was ```` ```ignore ````, so nothing compile-checked the wiring the module told people to use. The trait now carries associated types (`Trades`/`Books`) and is implemented for both `Stream<MarketEvent>` and `Stream<Burst<MarketEvent>>`, so `events.book_updates().order_book()` wires either way. The example is a real doctest now.

While fixing it I took the "smaller note" about `map_filter` constructing a discarded `Trade::default()` as part of the same change rather than leaving it — the burst impls needed hand-written ops regardless (there is no burst-preserving `map_filter`), so writing the two scalar ops alongside them cost almost nothing and removed the per-message allocation on the filtered path. Four small ops, no closures.

**2. `Qty` range.** Went with `i128` at 9 dp rather than a separate scale or `rust_decimal`. Reasoning, since you laid out all three:

- *Separate `DECIMALS` for `Qty`* is the cheapest change but bakes in a precision guess. The failure mode is a venue that exceeds whatever I picked, discovered by someone writing an adapter out-of-tree against a published API — the worst place to find it.
- *`rust_decimal`* is the strongest argument you made and I nearly took it. What stopped me: `market` is currently dependency-free, which is what lets a venue adapter crate depend on `wingfoil = { features = ["market"] }` without inheriting anything. That property is worth more here than the ~60 lines of parser, and the parser is exhaustively tested and does one narrow job.
- *`i128`* removes the ceiling outright (±1.7e20 at 9 dp), keeps exact parsing and integer key comparison, and costs 16 bytes plus a two-word compare on the `BTreeMap` key.

If a venue ever needs more than 9 dp, `rust_decimal` becomes the right answer and the newtypes mean it stays an internal change. Noted as such in `market/CLAUDE.md`.

`from_f64` → `try_from_f64`, fallible. You were right that "lossy" and "silently wrong" are different claims; NaN-to-zero and saturation are precisely what the module's thesis rejects.

**3. `InstrumentId` re-inventing `Sym`.** Lifted `Sym`/`SymbolInterner` into `adapters/common.rs`, `kdb` re-exports both, `InstrumentId` is built from them, and `InstrumentId::interned` takes a connection-lived interner. Your caveat about content-vs-pointer equality was the useful part and I've written it into the type docs so the next person doesn't "optimise" it into `Arc::ptr_eq`. I did **not** take the process-global interner — agreed that it's a real design commitment and shouldn't ride along here.

## Where I went a different way

**`book_for`: kept as a hard per-message check.** You're right that it's a construction-time invariant, but I don't think either suggested fix is right. `debug_assert` means a release build silently interleaves two venues into one book — the exact failure the check exists to prevent, in the build where it matters. "Check only on creation" doesn't work either, since the book is created from the first update, so there is nothing to compare against yet.

What I did instead was make it cheap: `InstrumentId::same_as` tries `Arc` pointer equality on both fields first and falls back to content equality. The pointer path hits in the case that actually occurs (the adapter clones one id into every message), so it's two pointer compares, and the fallback keeps independently-built ids comparing equal. That gets your perf point without trading the guarantee. Test added for the fallback specifically, since a fast path that returns a wrong `false` would be a nasty bug.

**`BookApply::Overflow` → `Gap(GapCause)`.** Agreed the fabricated `Gap { expected: got, got }` was nonsense. I made `Gap` carry a `GapCause` — `Sequence { expected, got }` or `BufferOverflow { buffered }` — rather than adding a sibling variant, so the ops keep a single "the book went invalid" arm to match on while the cause still says exactly what happened. `Stale`/`Refused` split as you suggested.

## The stale-snapshot bug

Real bug, unintentional asymmetry, thank you. `apply_snapshot` now refuses a snapshot a live book has already passed. Scoped to live books deliberately: `gap_out` clears `last_seq`, so a recovery snapshot is still always accepted no matter what id it carries — otherwise a venue that resets its ids could never recover from a gap. Both directions have tests.

## Smaller notes

- **Gap ids lost on the graph path** — fixed properly rather than papered over: the cause is retained on the book (`OrderBook::gap_cause()`), cleared on recovery. The op ticks the *book*, so that's the only route by which the ids reach downstream. Graph test asserts a monitor can read them.
- **`level_count` not gated** — gated, with a comment on why it matters despite being redundant today.
- **`parse_fixed` leniency** — documented (trimmed whitespace, leading `+`, `"1."`, `".5"`).

## Verification

25 unit tests (was 16), 8 graph tests (was 6), 2 doctests (was 1), 3 new `Sym` tests in `common`. `cargo fmt`, workspace clippy on default features, and clippy on every feature except aeron/iceoryx2 all clean; the legacy workspace builds against the moved `Sym`. Aeron I can't build here (needs CMake ≥3.30, box has 3.28) — CI's `--all-features` pass covers it.

Both `CLAUDE.md` files updated, including the out-of-tree adapter checklist: bursts pass through intact, `Refused` ≠ `Stale`, ids built once per subscription.

---

PR #710, opened 2026-08-04
feat(latency): percentiles tight enough to gate a regression on

0-jake-0:
## What this changes

The latency histogram goes from 5 to 8 sub-bucket bits, so a reported quantile's worst-case relative error drops from **3.125% to 0.39%**, and the report gains a **p99.9** column. `QUANTILE_RELATIVE_ERROR` is exported so callers can size a threshold against the instrument. `StageStats` loses `Copy` (it is now ~55 KiB).

## Why

3.125% is fine for drawing a chart and useless as a gate. The smallest thing worth alerting on is a ~2% p99 regression, and buckets 3.125% wide cannot resolve one — two runs either side of a real regression can land in the same bucket and print the identical number. The far tail is worse: p99.9 on a heavy-tailed hop lands where a bucket may hold a single sample, so the reported value is the bucket, not the sample.

The histogram was already fixed once for being unusable (the log2 layout that printed a p99 above the observed max). This is the same argument one notch finer: for a library whose subject is latency, percentiles you cannot gate on are missing the point of having them.

Two things fall out of the wider exact region:

- **Deltas below 256 ns (up from 32) are now exact**, not estimated — one bucket per nanosecond. That covers every in-process hop, an engine cycle being ~27 ns/node.
- **`StageStats` is no longer `Copy`.** At 55 KiB an implicit memcpy is not something to leave lying around. It stays `Clone`, so `vec[..; n]` and `slice::fill` are unchanged and the copies are now visible at the call site.

### Why 8 bits and not 10

HdrHistogram's convention is 3 significant digits (0.1%, 10 bits). Here that would be 25,600 buckets and 205 KiB per stage — the point where a per-session aggregator (the `latency_e2e` showcase runs up to 8) starts to cost real memory. 8 bits is ~2.4 significant digits at 6,912 buckets and ~55 KiB, comfortably under any regression worth an alert. The histogram is heap-held (`LatencyStats::stages` is a `Vec`) and never crosses the wire, so the size change is contained.

## How it was verified

- [x] `cargo fmt --all`

Five new tests. **Three of them fail if `SUB_BUCKET_BITS` goes back to 5** — I checked by reverting the constant and re-running:

| test | what it pins |
|---|---|
| `a_two_percent_tail_regression_moves_the_reported_p99` | a 2% tail shift moves the reported p99, by ~2% |
| `p99_tracks_truth_to_the_absolute_documented_bound` | p99 within an **absolute** 1/256 of truth |
| `p999_resolves_the_far_tail_separately_from_p99` | p99.9 separates from p99, both within the bound |
| `sub_microsecond_hops_are_exact_not_estimated` | 1/27/128/255 ns report exactly |
| `the_report_carries_p999_between_p99_and_max` | header and row stay in step |

The existing `quantiles_are_within_the_documented_error_bound` derives its tolerance from `SUB_BUCKET_COUNT`, so it holds at *any* resolution and cannot notice the resolution being narrowed. That is why the new one spells 1/256 out as a literal.

`--all-features` tests are green except the `*_integration` files, which need live services and belong to the per-adapter workflow tier (identical on `next` without them).

## Notes for the reviewer

- **Both tail tests carry a far 5 ms outlier, and that is load-bearing.** An earlier draft passed at 5 bits for the wrong reason: `quantile_ns` clamps to `[min_ns, max_ns]`, and with a single-valued tail the clamp lands on `max_ns` — which *is* the tail value — so the two runs differed for a reason unrelated to the histogram. Lifting `max_ns` clear makes the clamp inert so the number under test is the interpolation. It is written up at the tests; please keep it if they are ever edited.
- **`examples/showcase/latency/README.md` has a captured report that now lacks the new column.** House rules say sample output must be real, and I cannot re-run an iceoryx2 two-process showcase here, so I left the numbers alone and added a note that the capture predates `p99.9`. It wants a re-capture on a machine with iceoryx2 rather than a hand-edited column.
- `StageStats` losing `Copy` is the one source-breaking change. Nothing in-tree needed updating (every use is already by reference or an explicit clone), and `cargo lint-all` is clean across the workspace including the Python bindings and the legacy tree.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,9 @@ npm install @wingfoil/client  # TypeScript client for the web adapter
 - **Latency tracing that survives a process hop.** Per-hop wall-clock stamps
   aggregating into one report, across shared memory and the wire — see
   [`showcase/`](crates/wingfoil/examples/showcase/). `count`, `min`, `mean` and
-  `max` are exact; percentiles come off a sub-bucketed histogram and are
-  accurate to 3.125%.
+  `max` are exact; percentiles (p50/p99/p99.9) come off a sub-bucketed
+  histogram and are accurate to 0.39% — tight enough to gate a regression on,
+  not just to draw.
 - **Fallible everywhere.** Every lifecycle function returns a `Result`; a
   producer error propagates into the graph and aborts the run with context,
   and cleanup still runs.
--- a/crates/wingfoil/examples/showcase/latency/README.md
+++ b/crates/wingfoil/examples/showcase/latency/README.md
@@ -71,16 +71,22 @@ VM — so the IPC row is a reading of *that* machine under contention, not a
 figure for the transport. `min` (8.2 µs) is the closest thing here to the hop
 itself.
 
+> **This capture predates the `p99.9` column**, which the report now prints
+> between `p99` and `max`. The numbers above are otherwise unchanged; re-capture
+> on a machine with iceoryx2 rather than hand-editing a column into them.
+
 The three in-process deltas are **zero by construction**: `.stamp()` reads a
 per-cycle wall-clock snap, and those stages all run in the same engine cycle,
 so they share a timestamp. Only the IPC hop crosses a cycle boundary and shows
 a real number. Swap the in-process stamps for `.stamp_precise::<..>()` to get
 intra-cycle resolution — see **Time source** below.
 
-`min`, `mean`, `max` and `count` are exact. `p50` and `p99` are read out of a
-sub-bucketed histogram and carry at most 3.125% relative error — and are
-clamped to `[min, max]`, so a percentile is always a value the stage could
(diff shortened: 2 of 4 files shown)


0-jake-0:
Closing — not worth its cost right now.

This tightens percentile resolution from 3.125% to 0.39% so a tail-latency regression gate becomes possible. But there is no such gate: benches are deliberately *not* CI-gated in this repo, so nothing today is blocked by the coarser buckets. Against that, the change is not free — `StageStats` goes 7.7 KiB → 55 KiB and loses `Copy`, which is source-breaking.

Worth reopening if p99/p99.9 gating in CI becomes something you actually want; the branch (`latency-tail-percentiles`) has the tests that would fail if the resolution were narrowed again.

---

PR #712, opened 2026-08-04
feat(interp): opt-in per-node error containment

0-jake-0:
## What this changes

`Stream::on_error(f)` attaches an error handler to one node. When that node's `cycle` returns `Err`, the handler sees it and returns an `ErrorAction`:

| action | effect |
|---|---|
| `Abort` | Stop the run with node context. **The default**, and what happens with no handler attached. |
| `SkipCycle` | Treat this cycle as `Tick::Quiet` and carry on. The node stays live and is cycled again next activation. |
| `Quarantine` | Retire the node for the rest of the run — it never cycles again, so it never ticks, so its downstream cone goes quiet while the rest of the graph keeps running. |

## Why

A node's `cycle` error aborts the whole run. That is right for a backtest and right for most live graphs, and it stays the default. It is the wrong shape for a graph carrying live orders, where a decode failure on one instrument's leg takes down the leg managing risk on another. Today the only way to say otherwise is to catch every error inside the closure and invent a sentinel value — which loses the error and pushes the failure downstream as data.

Two properties I tried to keep:

- **Containment is never silent.** The handler that selects it is also the thing that gets told, so "the graph quietly stopped producing" is not a reachable state. There is no config flag that turns errors off.
- **One handler can serve a whole graph.** `ErrorHandler` is an `Rc<dyn Fn(&NodeError) -> ErrorAction>`, so a kill switch or a channel into an alerting sink attaches to many nodes and still identifies which one spoke, via `NodeError::index` / `::label`.

**Scoped to `cycle`.** `start` keeps aborting — a connect or configuration failure means the graph is not wired to what it claims to be — and `stop`/`teardown` already run error-safely.

`Quarantine` silences the node's whole downstream cone, so it wants aiming at a leaf or at the root of a subtree you can afford to lose. That is on the enum docs.

## How it was verified

- [x] `cargo fmt --all`

New `tests/error_policy.rs`, seven tests, all `HistoricalFrom(ZERO)`:

- `without_a_handler_an_error_still_aborts_the_run` — the default is untouched, context intact.
- `skip_cycle_drops_the_value_and_keeps_the_node_live` — asserts the surviving ticks keep their **original times**, so a dropped cycle cannot quietly reschedule the ones after it.
- `quarantine_silences_one_branch_and_leaves_the_rest_running` — a sibling branch off the same source runs the full schedule; the handler fires exactly once, not once per remaining activation.
- `a_handler_can_still_choose_to_abort` — sees the error on the way past ("log it, then die").
- `one_handler_can_serve_several_nodes_and_still_identify_them`.
- `a_re_run_clears_a_quarantine` — a quarantine belongs to the run that imposed it, as node state does.
- `a_failing_sink_can_be_contained_without_stopping_the_graph` — the motivating case: a rejected publish drops the message while the risk leg still sees every tick.

## Notes for the reviewer

- **Hot-path cost is one bounds-checked bool load per node cycle** (`quarantined[i]`), on a vector that is all-false unless a handler is attached. The `Option<ErrorHandler>` is only read on the error path, and cloned there (an `Rc` bump) so the handler runs without holding a borrow of `self.nodes` — it may want to inspect the graph, and `Quarantine` has to write back.
- **`quarantined` is separate from the existing `removed` tombstone** rather than reusing it: `removed` is `dynamic-graph`-gated and also unlinks edges and suppresses the end-of-run `stop`/`teardown`. A quarantined node should still get its cleanup hooks, and containment should not require the feature.
- `run_dynamic` is covered — `quarantined` grows alongside `removed` when a node is appended mid-run.
- This is the engine-level half. A follow-up worth considering is a first-class errors *stream* (`GraphBuilder::errors() -> Stream<Burst<..>>`) so failures can be wired rather than callback-ed; I left it out because delivering into the same cycle needs a layer-ordering answer the `marks` mechanism does not straightforwardly give, and the `Rc` handler covers the wire-it-to-a-channel case today.
- Interpreted only. `compiled()` is a closed box with no error-handler plumbing, and this is squarely an open-world/live-graph feature — but worth a line in the capability matrix if you want it tracked.

---

### DIFF
--- a/crates/wingfoil/src/fluent.rs
+++ b/crates/wingfoil/src/fluent.rs
@@ -32,7 +32,8 @@ use wingfoil::{NanoTime, RunFor, RunMode};
 use crate::Burst;
 use crate::channel::ChannelSender;
 use crate::interp::{
-    AsHandle, Builder, ExternalSource, FeedbackSink, Handle, Runner, SlotRef, StopHandle,
+    AsHandle, Builder, ErrorAction, ExternalSource, FeedbackSink, Handle, NodeError, Runner,
+    SlotRef, StopHandle,
 };
 use crate::op::{Activation, Ctx, Tick};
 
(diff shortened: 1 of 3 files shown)


0-jake-0:
Closing — this is a design decision that should be made deliberately, not inherited from a batch.

"What happens when a node's `cycle` fails" is a core contract, and this adds three public types (`ErrorAction`, `NodeError`, `ErrorHandler`) to a pre-1.0 crate shortly before cutover. Nobody asked for it; it came out of a review observation that the all-or-nothing model is awkward for a graph carrying live orders. That observation stands, but it's an argument for *you* deciding the shape, not for merging mine.

If it's ever wanted, `node-error-policy` has the engine change (one bounds-checked bool load per node cycle on the hot path, `Option<ErrorHandler>` read only on the error path) and seven tests asserting values and tick times.

---

PR #713, opened 2026-08-04
feat(fluent): split_some — two independently-ticking outputs from one node

0-jake-0:
## What this changes

`Stream<(Option<A>, Option<B>)>::split_some()` fans one node's value out into two streams, each ticking **only** on the cycles where its own side is `Some`.

Also rewires the `order_book` example to emit fills *and* top-of-book, which is what its README said it could not do.

## Why

An `Op` has a single `Out` by construction — that is what lets the engine own every value slot, and what makes the compiled tiers possible. The consequence for wiring is that a node with two things to say at different rates had nowhere to say them.

`split()` exists, but it is two `map`s over a pair, so both branches tick together. Modelling "no fill this tick" through it means emitting a `Default` fill and filtering it downstream — which puts a sentinel in the type and a filter in every consumer.

`split_some` keeps the node single-output and does the fan-out at wiring time, so:

- the shared work (the book update) happens **once**, however many branches read it;
- each branch ticks at its own rate;
- downstream of the split each side is an ordinary stream and recombines glitch-free like any other.

### It closes a gap the example documented against itself

`examples/core/order_book/README.md` said:

> **`split()` / multi-output nodes** — the legacy example emits *both* fills and prices from one node. The wingfoil engine's ops are single-output, so this port emits a single top-of-book stream.

The example now emits both, at their natural rates:

```text
price   bid  99   ask 103   spread 4
fill    buy    7 @ 103
price   bid  99   ask   -   spread -
price   bid  99   ask 104   spread 5
price   bid  99   ask 101   spread 2
price   bid 100   ask 101   spread 1
fill    buy    1 @ 101
price   bid 100   ask 102   spread 2
fill    sell  15 @ 100
```

The README now describes the idiom rather than the absence, and that sample output is a real capture of the rewired example (per house rules — I ran it and pasted what it printed).

## How it was verified

- [x] `cargo fmt --all`
- [x] `scripts/check-example-docs.sh` — 46 targets, all documented and indexed

New `tests/split_some.rs`, five tests:

- `each_branch_ticks_only_when_its_own_side_is_some` — asserts tick **times**, so a branch that fired on the wrong cycle would be caught even if its values looked right.
- `a_branch_that_never_fires_stays_completely_quiet` — no `Default` values leak out, which is what distinguishes this from `split()` over an `Option` pair.
- `the_shared_upstream_is_evaluated_once_per_tick` — counts closure invocations; this is the property that makes the idiom cheaper than computing the book twice.
- `the_two_branches_recombine_glitch_free` — the recombine fires once per cycle, not once per branch.
- `plain_split_still_ticks_both_branches_together` — keeps the contrast pinned so the two are not confused for one another.

## Notes for the reviewer

- **This is sugar, not an engine change.** `split_some` is two `map_filter`s over the same upstream handle; nothing in `interp.rs` or the `Op` contract moves. That is deliberate — the single-`Out` contract is load-bearing and I did not want to touch it to solve a wiring-ergonomics problem.
- **Naming.** `split_some` reads as "split the `Some`s apart", pairing with the existing `split`. `unzip_filter` and `split_optional` were the alternatives; happy to change it.
- **Two outputs only.** Three-plus wants either a nested pair or a variadic form; I left that until something asks for it rather than guessing at the shape.
- `split()` is untouched, and its docs now point at `split_some` for the different-rates case.
- The example's `Book::apply` now returns `Option<Fill>` instead of `()`, and the fold state becomes `(Book, Option<Fill>)` — the fill is what *this* message produced, the book carries across ticks.

---

### DIFF
--- a/crates/wingfoil/examples/core/order_book/README.md
+++ b/crates/wingfoil/examples/core/order_book/README.md
@@ -1,47 +1,67 @@
-## Order book: stateful top-of-book
+## Order book: one stateful node, two output streams
 
 A stateful streaming example: a limit order book maintained entirely in `fold`
-state. Each tick applies one synthetic message (a resting limit order, or an
-occasional aggressive order that crosses and consumes a level), and the graph
-emits the *top of book* — best bid and best ask — but only when it changes.
+state. Each tick applies one synthetic message — a resting limit order, or an
+occasional aggressive order that crosses and consumes a level — and the graph
+emits **two** streams from that one node: the *fills* it produced, and the *top
+of book* (best bid and best ask).
+
+The two tick at different rates, which is the point. Every message moves the
+prices; only some produce a fill.
 
 ```rust
-let book = g
+// Fold state is `(Book, Option<Fill>)`: the book carries across ticks, the
+// fill is what *this* message produced.
+let applied = g
     .ticker(Duration::from_millis(1))
-    .fold(Book::new(seed), |book, _| book.apply());
+    .fold((Book::new(seed), None), |state, _| {
+        state.1 = state.0.apply();
+    });
 
-let top = book.map(|b| b.top()).distinct();
-let _out = top.for_each(|p| { println!(...); Ok(()) });
+// One node, two outputs at different rates.
+let (fills, tops) = applied
+    .map(|(book, fill)| (fill.clone(), Some(book.top())))
+    .split_some();
+
+let _prices = tops.distinct().for_each(|p| { println!(...); Ok(()) });
+let _fills = fills.for_each(|f| { println!(...); Ok(()) });
 ```
 
 Sample output:
 
 ```text
-bid  99   ask   -   spread -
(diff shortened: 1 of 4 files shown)


0-jake-0:
Closing — the API half doesn't earn its place.

`split_some` is sugar over two `map_filter`s; a caller can write the same three lines without a new method on `Stream`, and adding public surface to a pre-1.0 crate weeks before cutover wants a stronger reason than that.

**The other half may be worth salvaging separately.** This branch also rewires the `order_book` example to emit fills *and* top-of-book, and rewrites its README, which had been documenting a limitation against itself ("multi-output nodes … not yet available on the wingfoil engine"). That claim is misleading regardless of whether `split_some` exists — the idiom works today via `map_filter`. Say the word and I'll open a small PR with just the example and README, no API change.

Branch `split-multi-output` retains both halves.

---

PR #715, opened 2026-08-04
feat(fluent): shared/unshared — collapse the per-hop payload clone

0-jake-0:
## What this changes

`Stream::shared()` wraps values in an `Rc` so pass-through hops downstream cost a refcount bump instead of a deep copy; `Stream<Rc<T>>::unshared()` clones back out where a consumer needs an owned value.

Also records what the `store_baseline` numbers actually say about the deferred arena decision.

## Why

Every op that republishes its input clones it — `filter`, `sample`, `distinct`, `delay`, `merge`, `throttle` all end in `input.clone()`, because `Op::Out` is an owned value and the engine owns the slot it lands in. For a scalar that is a register copy and invisible. For a payload that owns heap — a decoded book update, a `Vec` of levels, a `String` symbol — it is a full `memcpy` per hop per tick, and a real graph reaches six forwarding hops without trying.

## The measurement, and what it says about the arena

Paired `store_baseline` re-run (8 KiB `Vec<u64>` through 16 `filter` hops, 5,000 cycles, shared 4-core VM — all three bars back to back, so the ratios hold even though absolute times differ from the capture already in the README):

| Payload | Time | Per hop |
|---|---|---|
| owned `Vec<u64>` | 18.24 ms | ~228 ns |
| `Rc<Vec<u64>>` | 3.80 ms | ~48 ns |
| `f64` (scalar floor) | 2.36 ms | ~30 ns |

**79% of the owned-`Vec` run is payload copying.** The README already noted that as a 4.3× ceiling. What it did not draw out is the third row: the `f64` bar is *the same graph with nothing to copy*, so it is where an arena / slot-aliasing store would land **at best** — and `Rc` gets within 1.6× of it.

```
owned Vec ──────────────── 18.24 ms
                             │
                   Rc closes │ 14.44 ms  (91% of the available distance)
                             ▼
Rc<Vec>          3.80 ms ────┤
      arena could close      │  1.44 ms  (9%)
scalar floor     2.36 ms ────┘
```

So: **the `Rc` payload idiom recovers ~91% of what slot aliasing could ever deliver on this shape** — today, from the wiring, with no engine change and no new invariants.

That does not close the arena question, and I have not written it as though it does: SoA cache locality is a separate axis this bench does not isolate, and a graph forwarding many *small* owned payloads sits elsewhere on the curve. It does mean the arena is not the thing standing between a heap-payload graph and its performance, and that the cheapest correct advice for such a graph is `shared()` rather than "wait for the arena".

`benches/README.md` gains that reading beside the existing capture (which is left as-is, matching how the other superseded readings there are handled).

## How it was verified

- [x] `cargo fmt --all`

New `tests/shared_payload.rs`, four tests. They use a payload that **counts its own clones**, so "did the hop copy?" is observable rather than inferred from a timing — which is what makes them a real regression guard rather than a benchmark that happens to be in the test suite:

- `sharing_collapses_the_per_hop_clone_to_one` — six forwarding hops: 1 deep copy shared, ≥6 unshared.
- `a_shared_graph_produces_the_same_values_and_tick_times` — the same graph either way through `filter` + `delay`, asserted equal including tick times. Transparency is the property that makes this safe to reach for.
- `unshared_recovers_the_owned_value` — exactly two copies (in and out), the hops in between free.
- `fan_out_over_a_shared_payload_stays_copy_free` — three branches add none, which is where clone-per-hop cost multiplies fastest.

## Notes for the reviewer

- **This is deliberately not the arena.** The port plan defers slot aliasing pending numbers; the numbers are now here and they argue for landing the wiring-level lever first. I did not touch the value store.
- **`Rc`, not `Arc`** — a graph runs on one thread. Values crossing the channel layer must be `Send`, so `unshared()` before sending. That is on the docs.
- **Aliasing the slot outright would not be sound for `filter`** and it is worth writing down why, since it is the obvious "why not just share the `Rc` the engine already holds": a filter's slot must retain the last *accepted* value for passive readers, so aliasing it to the upstream slot would expose values the filter rejected. That is a semantic change, not an optimisation — which is part of why the arena is a bigger piece of work than it looks.
- The two `impl` blocks are inherent on `Stream`, alongside the existing `Stream<()>` and `Stream<(A, B)>` blocks, rather than on `StreamOps` — they are type-directed (`T` → `Rc<T>` and back) rather than general combinators.

---

### DIFF
--- a/crates/wingfoil/benches/README.md
+++ b/crates/wingfoil/benches/README.md
@@ -369,6 +369,45 @@ The `Vec` − `Rc<Vec>` gap is what slot-aliasing could recover: **4.3×**, or
 17.2 ms of the 22.3 ms run. That is the ceiling on the zero-copy passthrough
 work, and it is large. [Violin plot](images/ops/store_forward_clone.svg).
 
+### What the third row means for the arena decision
+
+The `f64` floor is the row that answers the go/no-go, and it is easy to read
+past. A paired re-run on a 4-core VM (all three bars back to back, so the
+ratios hold even though the absolute times differ from the capture above):
+
+| Payload | Time | Per hop |
+|---|---|---|
+| `Vec<u64>` (8 KiB deep copy per hop) | 18.24 ms | ~228 ns |
+| `Rc<Vec<u64>>` (refcount bump per hop) | 3.80 ms | ~48 ns |
+| `f64` (scalar floor) | 2.36 ms | ~30 ns |
+
+**79% of the owned-`Vec` run is payload copying.** But the floor is where an
+arena would land at best — it is the same graph with nothing to copy — and
+`Rc` gets to within 1.6× of it. In distance terms:
+
+```
+owned Vec ──────────────── 18.24 ms
+                             │
+                   Rc closes │ 14.44 ms  (91% of the available distance)
+                             ▼
+Rc<Vec>          3.80 ms ────┤
+      arena could close      │  1.44 ms  (9%)
+scalar floor     2.36 ms ────┘
+```
+
+So the ordering that matters is: **the `Rc` payload idiom, available today from
+the wiring with no engine change, recovers ~91% of what an arena/slot-aliasing
+value store could ever deliver on this shape.** The remaining ~9% is the `Rc`
+clone/drop and the wider slot.
+
+That does not close the arena question — SoA cache locality is a separate axis
+this bench does not isolate, and a graph forwarding *many* small owned payloads
+sits somewhere else on the curve. It does mean the arena is not the thing
+standing between a heap-payload graph and its performance, and that the
+cheapest correct advice for such a graph is
+[`Stream::shared`](https://docs.rs/wingfoil/latest/wingfoil/fluent/struct.Stream.html),
+not "wait for the arena".
+
 ## Topological sort vs per-path propagation
 
 [`topological_vs_per_path/`](topological_vs_per_path/) — the branch/recombine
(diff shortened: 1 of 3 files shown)


0-jake-0:
Closing — the code is thin, but the measurement in it is the part worth keeping.

`shared()` / `unshared()` are one-line wrappers around `map(Rc::new)`, which is not enough to justify new public surface.

**The finding is separable and still stands.** A paired `store_baseline` run (8 KiB `Vec<u64>`, 16 `filter` hops, 5,000 cycles) gives:

| payload | time | per hop |
|---|---|---|
| owned `Vec<u64>` | 18.24 ms | ~228 ns |
| `Rc<Vec<u64>>` | 3.80 ms | ~48 ns |
| `f64` (floor — nothing to copy) | 2.36 ms | ~30 ns |

The `f64` row is the same graph with nothing to copy, i.e. where an arena / slot-aliasing store lands *at best*. `Rc` closes 14.44 ms of the available 15.88 ms — **~91% of what slot aliasing could ever deliver on this shape**, from the wiring, with no engine change.

That is decision-relevant for the arena item in `port-plan.md`, which is deferred pending exactly these numbers. This branch (`passthrough-clone-tax`) carries the `benches/README.md` write-up; happy to open it as a docs-only PR if useful.

One thing worth recording either way: aliasing the slot outright is **not sound for `filter`** — its slot must retain the last *accepted* value for passive readers, so aliasing to the upstream slot would expose rejected values. That is a semantic change, not an optimisation, and part of why the arena is bigger than it looks.

---

PR #717, opened 2026-08-05
Keep filter quiet until the source's first tick

### LINKED ISSUE #440: filter: emits a phantom T::default() before the source's first value
> **Rescoped 2026-08-03** (tracker audit). The original issue had two halves: an
> active/passive **divergence** between `FilterStream` and `FilterNode`, and a
> **startup phantom-default** bug. The divergence half is moot on next — there is
> one `filter` `Op` (`ops.rs:2456`), and `FilterStream`/`FilterNode`/`node_flow.rs`
> resolve to zero files (see #454). The phantom-default half **reproduces on
> next** and is what this issue now tracks.

## The bug

`filter`'s condition is active. When the condition is `true` but the source has
not yet produced its first value, `Filter::cycle` clones `input.0` unconditionally
— which at that point is `T::default()`. The graph emits a `0` / `""` / `false`
that the source never produced.

```rust
// crates/wingfoil/src/ops.rs:2464
fn cycle(_cfg: &mut (), _state: &mut (), input: (&T, &bool), _ctx: &mut Ctx<'_>)
    -> Result<Tick<T>>
{
    if *input.1 {
        Ok(Tick::Value(input.0.clone()))   // <-- input.0 may never have ticked
    } else {
        Ok(Tick::Quiet)
    }
}
```

Same class as the `distinct` default-emission bug fixed in #436.

## Confirmed repro (run against `next` @ c8cd740)

```rust
const HISTORICAL: RunMode = RunMode::HistoricalFrom(NanoTime::ZERO);

#[test]
fn filter_emits_phantom_default_before_first_source_value() {
    let g = GraphBuilder::new();
    // Source genuinely produces nothing until count reaches 10 (t=90ns).
    let source = g.ticker(Duration::from_nanos(10)).count()
        .map_filter(|i| (*i, *i >= 10));
    // Condition is true from t=0 — long before the source's first value.
    let condition = g.ticker(Duration::from_nanos(10)).map(|_| true);
    let acc = source.filter(&condition).with_time().accumulate();
    let mut r = g.build();
    r.run(HISTORICAL, RunFor::Cycles(13)).unwrap();
    println!("{:?}", r.value(&acc));
}
```

Observed:

```
[(NanoTime(0),  0), (NanoTime(10), 0), (NanoTime(20), 0), (NanoTime(30), 0),
 (NanoTime(40), 0), (NanoTime(50), 0), (NanoTime(60), 0), (NanoTime(70), 0),
 (NanoTime(80), 0), (NanoTime(90), 10), (NanoTime(100), 11),
 (NanoTime(110), 12), (NanoTime(120), 13)]
```

Nine phantom `0`s at t=0…80ns. The source's first real value is `10` at t=90ns.

## Semantics: already settled, keep as-is

The port picked **option 2 (active everywhere)** from the original issue and it
is consistent — the condition is an active input, so a condition tick re-emits
the source's current held value. A second probe (source ticking at t=0 and every
100ns, condition every 10ns) confirms the intended sampler behaviour:

```
[(0,1), (10,1), (20,1), … (90,1), (100,2), (110,2), …]
```

That is not in question here. **Only the pre-first-value window is wrong.**

## Fix shape

Guard with an `Option`, as `difference` and `distinct` already do — stay `Quiet`
until the source has produced at least one value. `Filter`'s `State` is currently
`()`; this needs a `bool`/`Option` seen-flag, or an engine-level "has this input
ever ticked" signal if one is preferred over per-op state.

## Acceptance

- [ ] No emission before the source's first real value
- [ ] Test covering an independent, out-of-phase condition stream (the repro above)
- [ ] Existing active-sampler behaviour unchanged (second probe above stays green)
- [ ] `filter` / `filter_value` rustdoc states the condition-trigger behaviour explicitly

Guflly:
## What this changes

Keeps `filter` quiet until its source has ticked once. Later condition ticks still resample the held source value.

## Why

Closes #440.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo lint`
- [ ] `cargo lint-all` (the local Windows host lacks libclang)
- [ ] `cargo test --manifest-path crates/wingfoil/Cargo.toml --all-features` (same local toolchain limitation)
- [x] `cargo build --workspace --all-targets`
- [x] `cargo test --workspace`

### DIFF
--- a/crates/wingfoil/src/fluent.rs
+++ b/crates/wingfoil/src/fluent.rs
@@ -842,7 +842,9 @@ pub trait StreamOps<T>: Sized {
         D: Clone + Default + 'static,
         F: Fn(&T, &B, &C) -> Result<D> + 'static;
 
-    /// Emit only when `condition`'s current value is true.
+    /// Emit when either this stream or `condition` ticks while the condition's
+    /// current value is true. Condition ticks resample the held source after the
+    /// first source tick; before that, the filter stays quiet.
     ///
     /// Gates on a *stream*. When the test is a pure function of this stream's
     /// own value, [`filter_value`](StreamOps::filter_value) says it in one
@@ -852,7 +854,9 @@ pub trait StreamOps<T>: Sized {
         T: Clone + Default + 'static;
 
     /// Emit only the values `predicate` accepts — the predicate twin of
-    /// [`filter`](StreamOps::filter):
+    /// [`filter`](StreamOps::filter).
+    /// The predicate runs only when this stream ticks, so it never resamples a
+    /// held value.
     ///
     /// ```text
     /// price.filter_value(|p| *p > 100.0)
--- a/crates/wingfoil/src/ops.rs
+++ b/crates/wingfoil/src/ops.rs
@@ -2666,24 +2666,30 @@ impl Op for TimeWindowedMedianTimeWeighted {
 }
 
 /// Emits its source value when the condition stream's current value is true.
+/// Condition ticks resample the held source value after the source has ticked
+/// at least once. Before the first source tick, the filter stays quiet.
 pub struct Filter<T>(PhantomData<T>);
 
 #[op(build = filter, fluent)]
 impl<T: Clone + 'static> Op for Filter<T> {
     type Cfg = ();
-    type State = ();
-    type In<'a> = (&'a T, &'a bool);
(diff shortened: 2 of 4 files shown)


0-jake-0:
Thanks!

Issue #726, opened 2026-08-05
Two-pass codegen from a wired graph: `func!` quotation (dynamic topology at compiled speed)


0-jake-0:
Tracking issue for an **accepted design direction that has never had a tracker entry** — it exists only as [`docs/wired-graph-codegen-decision.md`](../blob/next/docs/wired-graph-codegen-decision.md) (289 lines). Read that doc for the full reasoning; this issue is the tracker, not a copy.

## What

Generate compiled-graph code by *running* the wiring — traversing the wired (interpreted) graph — and emitting source that a second compilation pass turns into the monomorphized runner.

This is **not** a replacement for `nitro!` but a **second front-end into the same emission mechanism**: the generator's output artifact *is* a plain wiring fn wrapped in `nitro!`. It buys the one thing `nitro!` structurally cannot do — **dynamic topology**, graphs whose shape is decided by running code (config files, discovered instrument lists) — at compiled speed.

## The one new primitive

**`func!` — user-level quotation of closures.** Naive traversal fails because a wired graph gives back its topology but not its closure *bodies*. The eligible subset is: closures are closed or carry an explicit, literal-emittable capture list; configs are literal-emittable data.

## Sequencing (doc §8)

- [ ] **1. Now, nearly free** — validate the UX with zero code: hand-write what the generator would emit for one realistic config graph (it is just `nitro!` input) and live with the workflow.
- [ ] **2. Small, independently valuable, non-breaking** — `func!` + `OpFn` + `src`/`loc` metadata. Useful for graph introspection/debugging **even if the generator is never built**.
- [ ] **3. Gated on a named workload (doc §7)** — the generator + `EmitLiteral` + parity tests in house style: run interpreted, run generated-compiled, compare values *and* tick times.
- [ ] **4. Explicitly deferred** — mixed-mode islands (§4.2), capture-as-runtime-parameters (§3), multi-module wiring scope.

Step 2 is the piece worth doing on its own merits. Step 3 should not start without a workload that needs it.

## Known limitations (accepted, doc §9)

Implicit environment capture unavailable to quoted closures; frozen captures are partial evaluation; body paths must resolve in the artifact's scope; pass-1 wiring must be deterministic; an unquoted closure makes a node generator-ineligible (v1: loud error listing call sites, interpreted mode never complains).

## Related

Gates the FPGA/HDL backend, which reuses this front-end wholesale.

PR #759, opened 2026-08-08
Closure quotation and graph introspection (#726 step 2), plus an emission spike

0-jake-0:
## What this changes

A wired graph can now report **what its nodes compute**. `func!` captures a closure's source at the call site, `Stream::with_src` records it against the node, and `Runner::describe` / `GraphBuilder::describe` hand back `NodeInfo` — op label and method name, edges, activation, and the closure body where quoted.

Also lands a **spike** proving a walker can print valid `nitro!` wiring source from that description.

## Why

Implements #726 step 2, which is worth having on its own merits — introspecting a graph you did not wire — whether or not the generator in step 3 is ever built. Closes: nothing yet; #726 stays open for step 3.

**It does not implement `docs/wired-graph-codegen-decision.md` §4, because that design does not compile.** Recorded as **D28** in `docs/deviation-register.md`.

§4 proposes binding every closure-config op by an `OpFn` trait — a blanket impl for `Fn`, a specific one for `QuotedFn` — so `map` and friends accept either form through one signature. Coherence holds (the first commit here proves it). Inference does not: **rustc propagates closure *signature* inference only from `Fn`/`FnMut`/`FnOnce` bounds**, so behind any other trait a closure literal loses both parameter-type and higher-ranked-lifetime inference. Migrating the catalog produced **~370 errors across 41 targets**, and the residue after reverting the fluent layer to `Fn` bounds was *entirely inside `nitro!` blocks* — `compiled()` emits closure literals into forwarders whose bounds come from the op, so the macro's inference rooting depends on that bound being an `Fn` bound. §3's fn-pointer coercion fails separately: `fn(&_) -> _` names an arity, so it would leave `join` and `fold` unquotable.

So **the ops are untouched and never see a `QuotedFn`**. The quotation is unwrapped one level earlier and the source recorded against the node. One method covers the whole catalog — built-ins and user ops alike — instead of a quoted twin per combinator, and it puts the metadata where a traversal actually looks.

```rust
let double = func!(|i: &u64| i * 2);
let doubled = ticks.map(double.f).with_src(&double);
assert_eq!(Some("|i: &u64| i * 2"), doubled.src());
```

One supporting change: `NodeInfo` carries the op's `#[op(build = …)]` **method** name alongside the type-name label. An emitter writes `.map(..)`, not `.Map(..)`, and the two differ by more than case (`try_map`/`TryMap`, `for_each`/`Sink`).

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` ✅ — **`cargo lint-all` not run**: `--all-features` cannot build in this environment (aeron needs CMake ≥3.30, box has 3.28; etcd needs protoc). Both are documented prerequisites in `CLAUDE.md` and unrelated to this diff, but CI is the first real check on that feature set.
- [ ] `--all-features` blocked as above. Ran the full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green. Doctests green (9 passed).

`tests/quotation.rs` (9) pins the no-drift property — recorded text re-spliced at a different call site computes what ran — all four catalog arities, that quotation is inert at runtime, and that a *procedurally* wired graph describes every node it generated with capturing closures correctly invisible.

`tests/emission_spike.rs` (6) asserts parity on values **and** tick times between the source graph and the generated artifact, plus the compiled tier.

## Notes for the reviewer

**The emission spike is a spike, and reads as one.** It lives in a test file, not `src/`. Its trick for proving pass 2 without invoking a second compiler: the intended artifact sits in the same file as a real `nitro!` block, and the walker must emit byte-identical text — so "this is valid wiring source" is proven by the file compiling, not by inspection. It also shows a config-driven loop emitting as N unrolled pipelines with no loop surviving, which is the topology `nitro!` structurally cannot express and the whole reason the two-pass design exists.

**Three gaps are pinned by tests rather than fixed**, so none can regress into a silent partial emission:
1. Data configs are not recorded — no ticker `Duration`, no `fold` seed. That is `EmitLiteral`, and the bulk of step 3. The walker models it honestly as a caller-supplied table.
2. An unquoted closure is indistinguishable from no closure (both `src: None`), so the walker cannot refuse precisely. §4.1's loud error needs a per-node "takes a closure config" flag — which `#[op]` knows and doesn't record — plus `#[track_caller]`.
3. Multi-edge and passive-edge ops are refused. `active_ups` keeps receiver-first order so `join` is probably recoverable, but nothing distinguishes receiver from argument and `sample`'s passive leg is absent from `active_ups`.

**Ergonomics are honest but not lovely.** `ticks.map(double.f).with_src(&double)` is two steps and easy to forget, and forgetting is silent until a generator complains. A `quoted!(stream.map(|x| …))` call-wrapping macro could collapse it, but that needs a macro that destructures a method call — a design question, not a mechanical one, so I left it out.

**Closedness enforcement moves to the generator**, which is where the requirement bites: it is the thing about to splice a body into another scope. §3's tier-2 explicit capture lists remain the route to emittable captures.

**Housekeeping:** a stale `quotation-spike` ref still points at this branch's head — the remote rejected the delete (proxy disconnect). Harmless, but worth pruning server-side.

---


0-jake-0:

This PR's work — `func!` / `QuotedFn`, `Stream::with_src`, and the emission spike — is carried forward in full. Closing to keep the queue readable; no content is being dropped.

---

PR #761, opened 2026-08-08
Record whether an op's config is a closure, so eligibility is decidable

0-jake-0:
**Stacked on #760**, which is stacked on #759. Base is `emit-literal`; this diff is only the top commit. Review order: #759 → #760 → this.

## What this changes

`NodeInfo` now says whether an op takes a **closure** config. That makes emission eligibility decidable, and the walker in `tests/emission_spike.rs` refuses erased-closure nodes instead of printing calls that don't compile.

## Why

Gap 2 of the three #759's emission spike pinned. A traversal could not tell `count` — which genuinely has no config — from a `map` whose closure the engine erased: **both report `src: None`**. So a walker had no way to refuse the second without also refusing the first, and emitted `map()` instead.

`#[op]` already knows the difference and simply never recorded it. It can see that the op's `Cfg` is one of the impl's own type parameters *and* that the parameter carries an `Fn`-family bound — `map`'s `F: Fn(&A) -> B`, and not `ticker`'s `Cfg = Duration`. Recorded now alongside the build name, which makes the statement precise:

```rust
takes_closure_cfg && src.is_none()
// => "a closure the engine erased and the wiring did not quote"
```

The walker refuses exactly those, with a reason naming `func!` and `with_src` — the "loud, precise failure" §4.1 asks for.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green.

Two tests pin **both** directions, which is the part that matters: an erased closure is refused, and a config-free op is *not* swept up with it. A one-directional test would pass for a flag that was simply always true.

## Notes for the reviewer

**The bound match is deliberately whitespace-insensitive.** `quote!` renders the predicate as `F : Fn (& A) -> B`, and the exact spacing is not a stable contract — matching it literally is the kind of thing that breaks on a proc-macro2 bump for no visible reason. Normalising first costs nothing.

**It also caught a real bug in my own test.** Tightening `multi_edge_ops_are_refused_with_reasons` revealed that its graph contained an unquoted `map` alongside the join, so the walker now (correctly) reported two refusals. The test asserted one. Fixed by quoting the `map`, which makes it a genuine test of the multi-edge rule rather than of eligibility in general.

**Still missing from §4.1: the call site.** Refusals name node indices, where `#[track_caller]` would name the line the user actually wrote. That is plumbing rather than discovery, and it has one wrinkle worth recording: the closure passed to `Stream::wire` breaks caller propagation, so the location has to be captured in the *generated fluent method* and recorded against the node the way `with_src` does — not simply annotated onto the builder method.

**Remaining after this**: emission covers single-edge chains only. Multi-edge (`join`) and passive-edge (`sample`) ops are refused rather than emitted. `active_ups` preserves receiver-first order so `join` is probably recoverable, but nothing yet distinguishes a receiver from an argument, and `sample`'s passive leg is absent from `active_ups` entirely.

---

### DIFF
--- a/crates/wingfoil-derive/src/lib.rs
+++ b/crates/wingfoil-derive/src/lib.rs
@@ -2070,6 +2070,31 @@ fn expand_builder(args: &OpArgs, b: &BuilderShape<'_>) -> syn::Result<TokenStrea
     // The `build = <name>` token as a string literal, recorded on the node so a
     // graph traversal can recover the method an emitter would have to write.
     let build_name = name.to_string();
+    // Does this op's `Cfg` hold a **closure**? True when `Cfg` is one of the
+    // impl's own type parameters and that parameter carries an `Fn`-family
+    // bound — `map`'s `F: Fn(&A) -> B`, and not `ticker`'s `Cfg = Duration`.
+    //
+    // Recorded on the node because a traversal cannot otherwise tell an op that
+    // takes no config from one whose closure the engine erased: both leave
+    // `NodeInfo::src` empty. Without it an emitter cannot say which nodes are
+    // ineligible, and silently prints a call with a missing argument.
+    let takes_closure_cfg = {
+        let cfg_name = quote! { #cfg_ty }.to_string();
+        let is_param = b.type_params.iter().any(|p| *p == cfg_name);
+        is_param
+            && b.preds.iter().any(|pr| {
+                // Whitespace-insensitive: `quote!` spaces tokens out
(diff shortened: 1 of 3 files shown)


0-jake-0:

The `takes_closure_cfg` / `build` node metadata that makes eligibility decidable is carried forward in full. Closing to keep the queue readable; no content is being dropped.

---

PR #762, opened 2026-08-08
Record the passive-edge mask, so multi-edge ops can be emitted

0-jake-0:
**Stacked on #761 → #760 → #759.** Base is `eligibility`; this diff is only the top commit. Review order: #759 → #760 → #761 → this.

## What this changes

`NodeInfo` records the op's passive-edge mask and exposes `edges_in_call_order()`. The emission walker consequently handles `join` and `sample` instead of refusing them — emission now covers single-edge, multi-edge and passive-edge shapes.

## Why

Gap 3, the last of the three #759's spike pinned.

`join` only needed the receiver-first convention `active_ups` already keeps. **`sample` needed something the graph did not have.** Its data leg is edge 0 and *passive*, its trigger is edge 1 and active, so the node records:

```
active_ups  = [tick]
passive_ups = [count]
```

That pair is **identical whichever way round the call was written**. The two lists are a partition — each preserves its own order, but the interleaving between them is gone, so `count.sample(&tick)` and a hypothetical reverse are indistinguishable.

The op's `#[op(passive = [..])]` mask is exactly the missing bit, and the derive already had it. Recorded on the node now, with `edges_in_call_order()` walking positions `0..n` and taking from whichever list each bit selects.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green.

Each shape is proven the same way as the rest of the spike: the expected artifact is a real `nitro!` block in the test file, so byte-identical output plus a compiling file means the emitted wiring is valid by construction. Parity then runs on values *and* tick times.

The `sample` test asserts the reconstructed order explicitly (`[data, trigger]`) as well as the emitted text, because that is the case where getting it backwards would still produce something that compiles — `n0.sample(&n1)` is valid Rust that computes the wrong thing.

## Notes for the reviewer

**Variadic ops stay refused, and correctly so.** `merge_all` → `MergeN` and `combine` → `CombineN` have hand-written forwarders, so they carry no `build` name. Emitting them as the n-ary call they do not accept would be worse than failing, since it would produce an artifact that fails at pass 2 in generated code. A test pins the refusal rather than leaving it to chance.

**What remains after this**: refusals name node indices rather than the call sites `#[track_caller]` would give; tier-2 captures (`func!([thresh] |x| …)`) are unbuilt; and the walker is still test-only code rather than a `codegen::generate` entry point.

**Stack note, repeated from #761**: this is now four PRs deep on unreviewed work, and #759 is load-bearing for the three above it. If #759 wants changes, the rebases compound. The remaining #726 items are mutually independent, so this is a natural place to stop adding layers.

---

### DIFF
--- a/crates/wingfoil-derive/src/lib.rs
+++ b/crates/wingfoil-derive/src/lib.rs
@@ -2078,6 +2078,9 @@ fn expand_builder(args: &OpArgs, b: &BuilderShape<'_>) -> syn::Result<TokenStrea
     // takes no config from one whose closure the engine erased: both leave
     // `NodeInfo::src` empty. Without it an emitter cannot say which nodes are
     // ineligible, and silently prints a call with a missing argument.
+    // The op's passive-edge mask, recorded so an emitter can rebuild the
+    // original call order from the partitioned active/passive lists.
+    let passive_bits = args.passive;
     let takes_closure_cfg = {
         let cfg_name = quote! { #cfg_ty }.to_string();
         let is_param = b.type_params.iter().any(|p| *p == cfg_name);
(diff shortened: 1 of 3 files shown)


0-jake-0:

The `passive_mask` and `edges_in_call_order` work is carried forward in full. Closing to keep the queue readable; no content is being dropped.

---

PR #763, opened 2026-08-08
`codegen::generate`: promote the walker from test code to shipping surface

0-jake-0:
**Stacked on #762 → #761 → #760 → #759.** Base is `multi-edge`; this diff is only the top commit. Review order: #759 → #760 → #761 → #762 → this.

## What this changes

The emission walker existed only inside a test file. It now lives in `src/codegen.rs` behind a real entry point, so generation is something a build script can call rather than something a test demonstrates.

```rust
let src = codegen::generate("desk", "u64", |g| { /* wiring */ })?;
codegen::write_artifact("src/desk.gen.rs", &src)?;
```

`generate` runs pass 1 itself, so a caller writes a wiring *function* and never touches a graph description — the shape §5's example uses. `emit` / `emit_with_tail` remain for a caller that already holds one.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green. Doctests green (11 passed).

`tests/emission_spike.rs` is renamed `codegen_emission.rs` and now exercises the real module rather than a private copy — so the tests that proved the walker are the tests guarding the shipped code. 14 tests.

## Notes for the reviewer

**Three decisions worth a look:**

- **`out_ty` is a parameter, not derived from `T`.** `std::any::type_name` is explicitly not guaranteed to produce valid source — it renders `Vec<u64>` as `alloc::vec::Vec<u64>`, which does not compile. Deriving it would work for primitives and fail on everything else, which is a worse failure than asking.

- **The tail is what the wiring returned, not the last node it wired.** These differ the moment a graph has a side sink: `let out = ..; out.for_each(..); out` leaves the sink last in wiring order. Emitting the last node would return the wrong stream — and since `for_each` yields `Stream<()>`, the artifact would fail pass 2 with a type error *inside generated code*, which is about the worst place for it. **I found this while writing the entry point, not from a failing test** — every existing test happened to end on its output. There is now a test that wouldn't.

- **Refusals report every reason at once**, not the first. Fixing one quotation per generator run is exactly the workflow this exists to avoid.

**`write_artifact` stamps a generated-file header, and that is not decoration.** A stale artifact is the one failure this design cannot detect, so saying so at the top of the file is the whole mitigation — hence it is not optional.

**`Breadcrumbs` is off by default.** Annotating each node with the wiring location its quotation came from is §5's mitigation for pass-2 errors landing in generated code, and genuinely useful. But it embeds line numbers, so any edit to the wiring file changes the artifact even when the graph did not — which would make a byte-comparison against a checked-in artifact fire spuriously. Opt in when diagnosing; leave off when diffing.

**Still open on #726**: refusals name node indices rather than call sites (`#[track_caller]`); tier-2 captures (`func!([thresh] |x| …)`); and the `.with_src()` / `.with_cfg()` ergonomics remain two-step and forgettable.

**Stack**: six PRs now, none reviewed, five of them chained. #759 is load-bearing for four above it.

---

### DIFF
--- a/crates/wingfoil/src/codegen.rs
+++ b/crates/wingfoil/src/codegen.rs
@@ -1,0 +1,368 @@
+//! Generate compiled-graph source by **running** the wiring.
+//!
+//! Pass 1 executes an ordinary Rust wiring function against the interpreted
+//! builder; this module walks the graph it produced and prints a `nitro!`
+//! wiring function. Pass 2 is an ordinary `cargo build` over that artifact,
+//! which turns it into the monomorphized runner.
+//!
+//! The point is the one thing `nitro!` structurally cannot do: **topology
+//! decided by running code**. A macro sees tokens, not a config file — so a
+//! loop over N discovered instruments is inexpressible. Pass 1 *runs* the loop,
+//! so the generator emits the N unrolled pipelines it produced, and `nitro!`
+//! never sees a loop at all.
+//!
+//! Design: `docs/wired-graph-codegen-decision.md`; sequencing: #726.
+//!
+//! # What the artifact is
+//!
+//! Plain, reviewable Rust: a `nitro!` invocation with the topology unrolled and
+//! every config inlined. `nitro!` stays the single backend, so there is exactly
+//! one place in the system that knows how to turn wiring into a monomorphized
+//! runner — the generator does not emit forwarder calls, state locals or kernel
+//! loops itself, and cannot drift from the macro that does.
+//!
+//! ```no_run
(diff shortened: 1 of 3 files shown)


0-jake-0:

`codegen::generate` and the `Ineligible` / `NotEmittable` refusal path are carried forward in full. Closing to keep the queue readable; no content is being dropped.

---

PR #765, opened 2026-08-08
`quoted!`: wire a node and record its closure in one step

0-jake-0:
**Stacked on #764 → #763 → #762 → #761 → #760 → #759.** Base is `tier2-captures`; this diff is only the top commit.

## What this changes

```rust
let out = quoted!(ticks => map(|i: &u64| i * 2));
```

instead of

```rust
let double = func!(|i: &u64| i * 2);
let out = ticks.map(double.f).with_src(&double);
```

## Why

The pair was correct but forgettable, and forgetting is **silent**: the graph runs fine interpreted and the omission surfaces much later as a generator refusal. I walked into it myself writing the worked example, in the same session I'd spent predicting other people would.

## The `=>` is a concession, and it's the interesting part

`quoted!(ticks.map(..))` reads better and **cannot be built** as `macro_rules!`: an `expr` fragment may only be followed by `=>`, `,` or `;` — never `.` — so the pattern cannot destructure a method call.

A **proc** macro can parse the chain, and I built that first. It fails on the one property that makes quotation worth having:

- `stringify!` recovers verbatim source only for a fragment that came straight from a user's file. Tokens re-emitted by a proc macro fall back to pretty-printing, so the artifact carries `| i : & u64 | i * 2`.
- I checked whether `rustfmt` would repair it: **it does not format inside a macro body**, so the ugliness would ship into every generated file.
- I checked whether `Span::source_text` could recover the original: on stable it returns only the first token (`Some("|")`), because joining spans across an expression is nightly-only.

So the syntax cost buys a readable artifact — which is what a generated file exists to be. A test pins the verbatim property through both macro layers, since that is the entire reason for the shape.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green. Doctests green (15 passed).

Six tests, including one asserting the shorthand records *byte-identically* to the manual form — so it stays a convenience rather than becoming a second mechanism that can drift.

## Notes for the reviewer

**Deliberately narrow.** Data configs still need `.with_cfg(..)` explicitly: recording them here would mean either evaluating the argument expression twice or silently requiring `Clone` of it, and a trait bound that appears from inside a macro is worse than a second method call. Captures still need `func!([name] ..)` from #764.

**It knows nothing about the catalog.** The expansion is an ordinary fluent call, so it covers user-defined ops as well as built-ins, and is not coupled to the generator's eligible set.

**Two arms only** — `recv => method(closure)` and `recv => method(arg, closure)`. That covers every closure-config op in the catalog (the leading arg being a stream edge for `join` or a seed for `fold`). A third arm is a one-line addition if an op ever needs two leading args.

**Chaining is lost** — each node is its own statement. That is the honest cost of the `=>` form; it is still one line per node against the previous two.

**Stack**: eight PRs, none reviewed, seven chained on #759.

---

### DIFF
--- a/crates/wingfoil/src/quote.rs
+++ b/crates/wingfoil/src/quote.rs
@@ -220,3 +220,76 @@ macro_rules! func {
         }
     };
 }
+
+/// Wire one node and record its closure in a single step.
+///
+/// The two-step form is correct but forgettable, and forgetting is **silent**:
+/// the graph runs fine interpreted, and the omission only surfaces later as a
+/// generator refusal.
+///
+/// ```
+/// use wingfoil::prelude::*;
+/// use wingfoil::{func, quoted};
+/// # let g = GraphBuilder::new();
+/// # let ticks = g.ticker(std::time::Duration::from_millis(1)).count();
+/// // Two steps, easy to half-do:
+/// let double = func!(|i: &u64| i * 2);
+/// let a = ticks.map(double.f).with_src(&double);
+///
+/// // One step, no opportunity to forget:
+/// let b = quoted!(ticks => map(|i: &u64| i * 2));
+///
+/// assert_eq!(a.src(), b.src());
+/// ```
+///
+/// Also takes a leading argument, for the ops that have one — a stream edge
+/// (`join`) or a seed (`fold`):
+///
+/// ```
+/// use wingfoil::prelude::*;
+/// use wingfoil::quoted;
+/// # let g = GraphBuilder::new();
+/// # let a = g.ticker(std::time::Duration::from_millis(1)).count();
+/// let b = quoted!(a => map(|i: &u64| i * 10));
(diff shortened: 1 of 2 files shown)


0-jake-0:
Superseded by #769, which squashes this stack (#759–#768) onto `main` — but unlike its neighbours, this one is **not** carried forward.

`quoted!` was retired in favour of the `#[op(fluent)]`-generated `_q` twins. The reason: `quoted!` still needed an explicit closure type annotation, and it could not be chained — so it was strictly worse than `map_q(func!(..))`, which does the same job in one call and composes. #769 contains no `quoted!`.

Closing.

---

PR #766, opened 2026-08-08
`quoted!` covers all three recording vocabularies

0-jake-0:
**Stacked on #765 → #764 → #763 → #762 → #761 → #760 → #759.** Base is `quoted-macro`; this diff is only the top commit.

## What this changes

`quoted!` previously handled a plain closure. That left the *interesting* case — a per-instrument parameter — needing the manual three-line `func!` + `.with_src()` form, which is exactly backwards: the shorthand worked for the easy case and not for the one the generator exists to serve.

One rule now covers everything, marked in the argument list:

| Write | Records |
|---|---|
| `map(\|x\| ..)` | the closure |
| `map([fee] move \|x\| ..)` | the closure **and** the captured value |
| `ticker(cfg period)` | the data config |
| `fold(cfg 0u64, \|a, v\| ..)` | both |
| `join(&other, \|x, y\| ..)` | the closure; `&other` is an edge, left alone |

So a per-instrument leg reads without leaving the macro:

```rust
let ticks = quoted!(g => ticker(cfg inst.tick)).count();
let px    = quoted!(ticks => map(|n: &u64| *n as f64));
let net   = quoted!(px => map([fee] move |p: &f64| p - fee));
```

`cfg` also removes a redundancy worth naming on its own: `ticker(p).with_cfg(&p)` wrote the value **twice**, so it could be forgotten in one place or — worse — changed in one place.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green. Doctests green (16 passed).

Twelve tests, one per arm plus the composite. Two assert the shorthand records **byte-identically** to the manual form, so it stays a convenience rather than a second mechanism. The last one builds a two-instrument desk and asserts *no* node with a closure config is left unrecorded — which is the property a user actually cares about, rather than any individual arm working.

## Notes for the reviewer

**Arm order is load-bearing, and the reason is a `macro_rules!` sharp edge.** Each arm is discriminated by a *literal* token (`cfg`, `[`) before any fragment is parsed, and arms run most-specific first. `macro_rules!` falls through cleanly on a literal mismatch, but a failed `$x:expr` parse is a **hard error**, not a fallthrough — so an arm that could swallow another's input has to come second. The comment in the macro says this, because the ordering looks arbitrary otherwise and would be "tidied" into breakage.

**A `cfg`-marked argument is cloned** into the call so the original survives to be rendered. Every `EmitLiteral` type is `Clone`, and for the common `Copy` ones (`Duration`, numbers) the clone compiles away.

**Still not detected: an undeclared capture.** `map(move |p| p - fee)` without `[fee]` records a body that will not resolve in an artifact. It surfaces as a generator refusal — or, if the name happens to resolve to something else at the splice site, as a pass-2 compile error. That is the one remaining hole in the quotation story, and closing it needs the fn-pointer coercion §3 specifies, which cannot be written arity-generically.

**One doctest caught me**: the markdown pipe-escaping (`\|`) needed for the table leaked into the adjacent code block, so the example was invalid Rust until the doctest run caught it. Worth knowing if you add rows to that table.

**Stack**: nine PRs, none reviewed, eight chained on #759.

---

### DIFF
--- a/crates/wingfoil/src/quote.rs
+++ b/crates/wingfoil/src/quote.rs
@@ -290,12 +290,12 @@ macro_rules! func {
 /// use wingfoil::prelude::*;
 /// use wingfoil::quoted;
 /// # use std::time::Duration;
-/// # struct Instrument { tick: Duration, fee: f64 }
-/// # let cfg = [Instrument { tick: Duration::from_millis(1), fee: 2.5 }];
+/// # struct Instrument { period: Duration, fee: f64 }
+/// # let cfg = [Instrument { period: Duration::from_millis(1), fee: 2.5 }];
 /// # let g = GraphBuilder::new();
 /// for inst in &cfg {
 ///     let fee = inst.fee;
-///     let ticks = quoted!(g => ticker(cfg inst.tick)).count();
+///     let ticks = quoted!(g => ticker(cfg inst.period)).count();
 ///     let px = quoted!(ticks => map(|n: &u64| *n as f64));
 ///     let net = quoted!(px => map([fee] move |p: &f64| p - fee));
 ///     assert!(net.src().is_some());
--- a/crates/wingfoil/src/quote.rs
+++ b/crates/wingfoil/src/quote.rs
@@ -271,25 +271,97 @@ macro_rules! func {
 /// trade. `Span::source_text` would settle it, but joining spans across an
 /// expression is nightly-only.
 ///
+/// # One rule for all three vocabularies
+///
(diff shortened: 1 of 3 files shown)


0-jake-0:
Superseded by #769, and like #765 **not** carried forward — `quoted!` was retired rather than extended, for the reason given there.

Closing.

---

PR #767, opened 2026-08-08
`#[op(emit_cfg)]`: data configs record themselves

0-jake-0:
**Stacked on #766 → #765 → #764 → #763 → #762 → #761 → #760 → #759.** Base is `quoted-full`; this diff is only the top commit.

## What this changes

An op with a concrete config records it during ordinary wiring:

```rust
let ticks = g.ticker(period).count();   // recorded, no annotation
```

Applied to `ticker`, `delay`, `window`, `buffer`, `limit` and `throttle`.

## Why

A reviewer noticed that `quoted!(g => ticker(cfg PERIOD))` isn't quoting anything — `ticker` takes a `Duration`. I had extended a *quotation* macro to also do data-config recording and named it after only one of the two jobs. Fair hit, and it pointed at real accidental complexity rather than a design constraint:

- A **closure** is *erased*. Nothing can recover it, hence a call-site macro. That's information-theoretic.
- A **data config** is not erased at all. The value sits in the op's `Cfg` cell, intact, at runtime. It simply never reached the node's type-free metadata, which is where `describe()` looks.

The second is plumbing. It should never have needed user vocabulary, and now doesn't: the `#[op]`-generated `Builder` method renders the config through `EmitLiteral` and records it itself, before the value moves into the cell.

Which value is "the data" is unambiguous — an `init_arg` op takes its seed there and carries a *closure* as its `Cfg` (`fold`), while every other emittable op carries the data in `Cfg` (`ticker`). No op has both.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `--all-features` blocked as above. Full default suite plus `bench,async,csv,cache,augurs,market,ws,redis,postgres,kafka,web,kdb,prometheus,fix,dynamic-graph,tracing` — green. Doctests green (16 passed).

**The strongest evidence is what didn't change:** the byte-exact artifact assertions in `codegen_emission.rs` still hold after removing every `.with_cfg(&period)` from those graphs. Automatic recording emits precisely what the manual form did.

## Notes for the reviewer

**Opt-in, and the reason is the interesting part.** The `EmitLiteral` bound lands on the op's **public** signature, so it can only go on ops whose data type is *concrete*. `constant`'s `Cfg` is `T`; `fold`'s seed is `B`. Adding the bound there would forbid folding into any accumulator the generator cannot render — a breaking change imposed on users who aren't buying generation at all. I tried it, hit exactly that, and backed it out; those ops keep the manual `with_cfg`, and a test pins the asymmetry so it reads as a decision rather than an oversight.

That is also why it can't simply be inferred from "the `Cfg` is not a closure" (which #761 already computes): `logged`'s config is `(String, log::Level)`, and `log::Level` isn't `EmitLiteral`.

**`quoted!`'s `cfg` arm stays**, now scoped to exactly the generic-config ops. Its doc comment says so, so nobody deletes it as dead.

**`Builder::set_node_cfg_src` becomes `pub`** because generated code calls it. Same reason the other `set_node_*` seams are public.

## Where this leaves the user surface

One thing to learn instead of three:

| | |
|---|---|
| Closures | `quoted!(s => map(\|x\| ..))`, or `func!([cap] ..)` when it captures |
| Concrete data configs | nothing — automatic |
| Generic data configs (`fold`/`scan` seeds) | `.with_cfg(&seed)` |

**Stack**: ten PRs, none reviewed, nine chained on #759.

---


0-jake-0:

`#[op(emit_cfg)]` and `tests/auto_cfg.rs` are carried forward in full — in fact #769 is the only branch that has that test file. Closing to keep the queue readable; no content is being dropped.

---

PR #770, opened 2026-08-08
`#[wiring]`: record closures and their captures with no call-site annotation

0-jake-0:
**Stacked on #769 — base is `lightning-codegen`, not `main`.** Review #769 first; this is the delta on top.

## What this changes

Wiring records itself. Where #769 needs a `_q` twin or a `func!` quotation per closure, one attribute on the wiring function covers all of them — captures included:

```rust
#[wiring]
fn desk(g: &GraphBuilder, book: &[Instrument]) -> Stream<f64> {
    book.iter().map(|inst| {
        let fee = inst.fee;
        g.ticker(inst.period).count().map(move |n: &u64| *n as f64 - fee)
    })
    .reduce(|a, b| a.join(&b, |x: &f64, y: &f64| x + y))
    .expect("at least one instrument")
}
```

No `func!`, no `_q`, no `with_src`, no `with_cfg`, no capture list. That emits:

```rust
let n2 = n1.map({ let fee = 0.5f64; move |n: &u64| *n as f64 - fee });
```

Also lands a `codegen` example, staleness detection, and two bug fixes found by probing the generator's edges (below).

## Why

#769 leaves two things unresolved. The annotation burden is one — a `_q` method per closure-config op, and a hand-written `[fee]` list for every capture. The other is quieter: recorded text was the token stream spaced out (`move | n : & u64 | * n as f64`), and it landed in the artifact and stayed there, because `rustfmt` does not format inside macro bodies. That matters more than it looks — "the artifact is plain reviewable Rust you can diff" is the design's *only* mitigation for undetectable stale generation, and text nobody wants to read undermines it.

## How it can rewrite blindly

The macro sees tokens, so it cannot tell `Stream::map` from `Iterator::map`. It does not try: it rewrites **every** method call carrying a closure and lets method resolution decide. `Stream` has an *inherent* `__wf_src` that records; everything else picks up a blanket `MaybeSrc` no-op. Inherent methods win over trait methods, so nothing has to opt out and iterator chains inside the wiring compile to exactly what they did before — asserted, not assumed.

## The load-bearing decision: capture detection is **soft**

Free-variable analysis finds the captures; rendering them goes through a new `emit::Probe` — an autoref-specialisation ladder returning `Some(literal)` when the type implements `EmitLiteral` and `None` when it does not.

That softness is the whole design. The attribute annotates a *whole wiring function*, most of whose nodes are never emitted, so an `EmitLiteral` bound per detected capture would make `orders.map(move |o| book.lock()…)` over an `Arc<Mutex<Book>>` a hard compile error in ordinary wiring — which would make `#[wiring]` unusable on any real graph. Instead the wiring compiles and runs, and only *generation* refuses, naming the binding rather than saying "not quoted" (which would send a reader to fix something already done).

The ladder was spiked before anything was built on it. The double `&&` at the call site is load-bearing: with a single `&` the fallback matches first and always wins.

## Two bugs, both found by probing rather than by a failing test

**`fold`/`scan` silently dropped their seed.** `Fold`'s `Cfg` *is* its closure; the seed arrives via `#[op(init_arg)]`, a separate call-site argument with no other trace on the node. So quoting the closure satisfied `takes_closure_cfg` while saying nothing about the seed, and the walker emitted `.fold(f)` — an artifact missing the accumulator's starting value, **reported as fully emittable**. That is exactly what `codegen`'s docs promise cannot happen ("a refusal, never a partial artifact"). `#[op]` now records `has_init_arg` and `ineligible` refuses.

**`[T; N]` and 5-tuples fell off `EmitLiteral`.** An unsized slice is only seen behind a reference, so `[f64; 3]` never reached the `[T]` impl; tuples stopped at 4 because that is where the list stopped. Quiet gaps before — but capture detection routes through `EmitLiteral`, so they had become *refusals*: capturing a `[f64; 3]` made a graph unemittable while the `&[f64]` spelling of the same data worked.

## Also in here

- **`check_artifact`** — staleness has no runtime symptom (an artifact built from last quarter's fees compiles, runs, and looks plausible). One comparison in a test catches both a config change nobody regenerated for and a hand edit, with no hashing and no new dependency.
- **`poll` is now emittable** — the piece #758 unblocked. `no_builder` suppressed the `set_node_build` call that records the method name, so the walker saw a nameless node and refused with advice about variadic ops that does not apply. A config-driven busy-poll ingest graph now generates.
- **The `codegen` example** — the house-rule gap #769 left. Runs the whole loop in one binary: the checked-in artifact is `include!`d, so pass 2 happens at build time and both tiers run side by side.
- **`tests/waked_source_under_spin.rs`** — pins that a wake-driven source keeps delivering while the kernel spins, so "develop against the waker, deploy to busy-spin" is one wiring rather than two graphs. Without it, `poll` (realtime-only) and `channel` (deterministic replay) are structurally different graphs, and you would verify one and ship the other.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo lint` and `cargo lint-all` — both green
- [x] `cargo test --manifest-path crates/wingfoil/Cargo.toml --all-features` — 92 suites green, plus 15 doctests, 5 derive unit tests, and `scripts/check-example-docs.sh` at 48 targets. The nine `*_integration` suites fail on `Socket not found: /var/run/docker.sock` — testcontainers, no Docker in this sandbox, unrelated to this diff.

Emission is proven the way the rest of the generator is: each expected artifact sits in the test file as a real `nitro!` block and the walker must emit byte-identical text — strings match *plus* the file compiles means the output is valid wiring source by construction. Parity then runs across interpreted, nested and compiled.

## Notes for the reviewer

**This supersedes part of #769's surface but deletes nothing.** With capture detection, the `_q` twins buy nothing the attribute does not, and `func!` drops to an escape hatch for the one case analysis cannot reach — a closure built outside the annotated function, where there are no tokens at the call site. Removing the twins is deliberately a **separate** follow-up, so it reads as its own decision rather than churn inside the change that introduced them.

**Recorded text is canonical, not verbatim, and that is the better of the two.** The macro's input is a token stream carrying no whitespace, so any two spellings that tokenise alike record alike — reformatting the wiring cannot churn a checked-in artifact. That holds by construction, so it is documented rather than tested; a test that cannot fail would be worse than none.

**Three classes of capture escape detection**, each documented on `free_vars`: a name used only inside a macro invocation (`syn` sees an opaque token stream); a captured *callable* invoked as `f(x)` — excluded deliberately, so calling a free helper function is not mistaken for a capture; and a name both bound and free in one closure (bindings are collected flat, no lexical scoping). All three fail as pass-2 compile errors, never as silently wrong graphs, and `func!` remains the explicit override.

**One rejected alternative, spiked and recorded.** Giving an op a mode type parameter (`Ingest<T, Waked>` / `Ingest<T, Spun>`) with `const ACTIVATION = M::ACTIVATION` cannot work: `#[op]` lifts that expression verbatim into a module-level non-generic const where the parameter is not in scope. Making it generic would mean the macro naming type arguments at every dispatch site, which is what its name-based forwarder mechanism exists to avoid.

**`prettyplease` is a new dependency of `wingfoil-derive`** and costs nothing in the graph — it depends only on `syn` and `proc-macro2`, both already there. The wrapping is a little grubby (the closure is parked in a throwaway `let` inside a throwaway `fn`, formatted, cut back out and dedented) because `prettyplease` formats a `syn::File`, not an expression; it falls back to the raw token string if any step surprises us, since layout must never panic a proc macro.

**Deviation register**: D28 updated for capture detection, C4 for what #758 settled and what remains excluded.

---

https://claude.ai/code/session_01X4eojqRWBJ6uK5v5JDzmkd

---


0-jake-0:
Folded into #769, which now carries the whole of project lightning in one PR rather than a two-deep stack.

No content is lost and nothing was rewritten: `lightning-codegen` was **fast-forwarded** to this branch's head (it was already an ancestor), so all thirteen commits from here are now in #769 with their hashes intact — including the two added since this PR was opened, which #769's description was blocking anyway:

- `5cd2b28` deletes the `_q` twins, so `main` never carries a surface we intended to remove
- `ef94f23` makes refusals name a wiring line even for a node nothing annotated

Closing.

---

PR #794, opened 2026-08-14
Improve nitro errors for Stream method collisions

### LINKED ISSUE #784: `nitro!`: denylist `Stream`'s inherent methods so typos there keep a curated error
Filed out of `docs/decisions/macro-extensibility-decision.md` §4 "Remaining work to productize", which had been carrying real open work inside a decision record instead of in the tracker.

## The gap

Since the op table was deleted, `nitro!` no longer knows the set of valid method names — an unrecognised method is emitted as a forwarder call and fails at resolution. That is the right default (it is what makes user ops work with no macro-crate edit), but it degrades the error for one specific case: `Stream`'s own **inherent** methods — `clone`, `handle`, `wire`.

Typo one of those inside a `nitro!` block and you get a forwarder-resolution failure pointing at generated code, rather than a message saying the method exists on `Stream` but is not a dual-mode combinator.

## The fix

A small denylist of `Stream`'s inherent methods in the macro crate, producing a curated diagnostic. Deliberately *not* an allowlist of valid ops — that would reintroduce the table the design exists to avoid.

Low priority: it is an error-message quality issue, not a correctness one.

floze-the-genius:
## What this changes

`nitro!` now rejects the `clone`, `handle`, and `wire` names before forwarder emission, with a curated call-site diagnostic explaining each `Stream` method collision. A trybuild fixture pins all three errors and verifies that generated `__WF_OP_*` failures do not leak through.

## Why

Ordinary fluent wiring resolves these names on `Stream`, while compiled emission could resolve same-named op forwarders. Reserving the closed three-name collision set keeps the two tiers unambiguous without reintroducing an allowlist of valid ops.

Closes #784.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo test --manifest-path crates/wingfoil/Cargo.toml`
- [x] `cargo test --manifest-path crates/wingfoil/Cargo.toml --all-features --lib` (321 passed)
- [x] `cargo test --manifest-path crates/wingfoil/Cargo.toml --test trybuild` (11 fixtures passed)
- [ ] `cargo test --manifest-path crates/wingfoil/Cargo.toml --all-features`
  - The all-features run passed 321 library tests and 20 Aeron adapter tests, then the Docker-backed `aeron_integration` binary timed out waiting for `/dev/shm/aeron-integration-test/cnc.dat` on macOS (17 failures, 1 expected no-driver test passed). The dedicated CI workflow supplies that media driver.
- [ ] New behaviour is covered by a test asserting **values and tick times**
  - Not applicable: this changes compile-time diagnostics only; graph construction and runtime semantics are untouched.

## Notes for the reviewer

The denylist remains the issue's closed `clone` / `handle` / `wire` set. Unknown built-in or user-defined op names still take the naming-convention forwarder path unchanged.

### DIFF
--- a/crates/wingfoil-derive/src/lib.rs
+++ b/crates/wingfoil-derive/src/lib.rs
@@ -176,7 +176,17 @@
 //!
 //! Because dispatch is by naming convention rather than a table, a method the
 //! macro cannot dispatch surfaces as an *unresolved forwarder* rather than as a
-//! message about the method. There are two cases, and they read differently:
+//! message about the method. There are three cases, and they read differently:
+//!
+//! **A name reserved by `Stream`'s inherent interface** — `clone`, `handle`, or
+//! `wire` — is rejected with one message explaining the collision. Those names
+//! cannot identify user ops: ordinary fluent wiring would resolve the inherent
+//! method while compiled emission resolved a same-named forwarder, making the
+//! two tiers mean different things.
+//!
+//! This denylist is deliberately the small, closed inherent surface — it is not
+//! an allowlist of valid ops. New built-in and user-defined op names still need
+//! no macro-crate change.
 //!
 //! **A method that cannot be an op** — `split` (two outputs, where an `Op` has
 //! one `Out`) or `feedback` (a cycle) — is rejected with one message naming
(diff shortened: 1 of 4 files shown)


0-jake-0:
Merged as 6594d4e — thanks @floze-the-genius. The diagnostic is a real improvement: `.clone()` in a `nitro!` block went from four errors naming only macro internals, each carrying a "similarly named constant" suggestion that offers to rewrite your call as `.__WF_OP_COUNT_PASSIVE()`, to one message at the call site. Pinning it with a trybuild fixture was the right call, and the comment trail explaining *why* each name can't be an op is the part that will keep it correct.

Follow-up in #795 extends the same mechanism rather than changing it:

- The inherent set is seven names, not three. `impl<T> Stream<T>` in `fluent.rs` also has `graph`, `value_slot`, `build` and `upstream`, and those were still falling through to the leaked `__WF_OP_*` path. `.build()` is the one that matters in practice — `Stream::build` is the documented way to close a *fluent* chain (`.print().build().run(..)`), so it's the one a user carries into a `nitro!` block by habit.
- `collapse_accumulate` and `filter_none` join the fluent-only set. Both are inherent one-liners over `fold` / `map_filter`, so they had no forwarders either and produced the same leaked errors; they now take the "spell the primitive" advice `split` gets.

Worth saying that the short list wasn't your reading of the issue — it was correct against what #784 asked for. The `(clone, handle, wire)` parenthetical traces back to the decision record's own remaining-work item, written when the op table was deleted, and got copied forward verbatim into the issue and then into the implementation without anyone re-deriving it against `fluent.rs`. So #795 also swaps the enumeration for something checkable: one trybuild block per method, so the denylist and the impl block can be diffed by eye.

---

PR #839, opened 2026-08-14
fix(fix): complete frames across WouldBlock

### LINKED ISSUE #820: fix adapter: no WouldBlock handling on non-blocking sockets — torn frames and run aborts under backpressure

Every outbound FIX message goes through `write_encoded`'s `sock.write_all(&bytes)?; sock.flush()?;` with no `WouldBlock` handling (`crates/wingfoil/src/adapters/fix.rs:1915-1916`) — but every path that reaches it in spin mode runs a **non-blocking** socket: the spin initiator (`fix.rs:2443`), the spin acceptor's accepted stream (`fix.rs:2333`), and the `fix_send` sink (`fix.rs:3059`).

`std::io::Write::write_all` retries only `Interrupted`; on `WouldBlock` it returns `Err` with data *partially written*. So when the kernel TCP send buffer fills (slow counterparty, TCP stall):

1. a **torn frame** goes on the wire — the peer sees a garbled stream mid-session;
2. the error propagates as a run abort via `?`;
3. the teardown guards (`fix.rs:2400-2406`, `fix.rs:3110-3116`) then append a Logout **after** the partial frame, garbling the session further.

`fix/CLAUDE.md:76-78` claims `fix_send` is "back-pressured by the kernel TCP send buffer" — true only of a blocking socket; the doc currently contradicts the code. Threaded mode is unaffected (blocking socket + read timeout, `fix.rs:2517`, `fix.rs:2683`). No retry exists upstream.

**Fix:** in `write_encoded`/`write_at_seq`, loop on `ErrorKind::WouldBlock`/`Interrupted` (busy-retry is consistent with spin mode's design), or track unwritten bytes and complete the in-flight frame on the next cycle before sending anything else. A frame must never be half-written. Update the `fix/CLAUDE.md` back-pressure sentence to match whichever behavior lands.

shin4141:
## What this changes

FIX writes now retain and busy-retry the unwritten suffix when a non-blocking socket reports `WouldBlock`, so a later send cannot follow a torn frame. The FIX contributor documentation now states the corresponding backpressure behavior.

## Why

`Write::write_all` returns immediately on `WouldBlock`, even after writing a prefix. The resulting error could tear a FIX frame and end the run under temporary TCP backpressure.

Closes #820.

## How it was verified

- [x] `cargo lint` (upstream pre-commit hook)
- [x] `cargo clippy -p wingfoil --features fix --lib -- -D warnings`
- [x] `cargo test -p wingfoil --features fix --lib` — 137 passed
- [x] `cargo test -p wingfoil --features fix --test fix_adapter` — 2 passed
- [x] `cargo test -p wingfoil --features fix-integration-test --test fix_integration -- --test-threads=1` — 6 passed
- [x] Upstream pre-push hook: `cargo build --workspace --all-targets` and `cargo test --workspace`
- [x] New behavior is covered by a test that accepts a frame prefix, reports `WouldBlock`, then verifies exactly one complete heartbeat frame
- [ ] `cargo lint-all` — local all-feature build stopped in the unrelated Aeron dependency because `cmake` is unavailable
- [ ] `cargo test -p wingfoil --all-features` — not run locally

## Notes for the reviewer

Busy-retrying is deliberate for the latency-first `AlwaysSpin` mode and means its graph thread remains pinned until the socket becomes writable. Threaded sessions keep their existing blocking-socket behavior.

### DIFF
--- a/crates/wingfoil/src/adapters/fix.rs
+++ b/crates/wingfoil/src/adapters/fix.rs
@@ -162,7 +162,8 @@
 //!
 //! [`FixOperators::fix_send`] opens its own outbound session (connect + logon at
 //! graph `start()`, realtime-only) and writes each [`FixMessage`] from the graph
-//! thread; back-pressure is the kernel TCP send buffer. A historical run aborts
+//! thread. When its non-blocking socket is backpressured, it busy-retries the
+//! current frame until the remaining suffix is written. A historical run aborts
 //! at `start()` with a "real-time" error (matching legacy's `start` check). The
 //! [`FixSender`] handle (from [`FixConnection::sender`]) is the *other* outbound
 //! path — a lock-free bounded queue drained by the `Threaded` session thread,
@@ -1542,6 +1543,25 @@ enum SeqDecision {
     Fatal(String),
 }
 
+/// Write one complete frame, treating `WouldBlock` as transient backpressure.
+///
+/// `AlwaysSpin` sessions use non-blocking sockets, so `Write::write_all` can
+/// return after writing only a prefix. Keep the unwritten suffix here instead
+/// of letting the caller tear down a connection with a partial FIX frame on it.
+/// Blocking sockets retain their normal behavior.
+fn write_frame<W: Write>(writer: &mut W, mut bytes: &[u8]) -> io::Result<()> {
+    while !bytes.is_empty() {
+        match writer.write(bytes) {
+            Ok(0) => return Err(io::ErrorKind::WriteZero.into()),
+            Ok(written) => bytes = &bytes[written..],
+            Err(error) if error.kind() == io::ErrorKind::Interrupted => {}
+            Err(error) if error.kind() == io::ErrorKind::WouldBlock => std::hint::spin_loop(),
+            Err(error) => return Err(error),
+        }
+    }
+    Ok(())
+}
+
 impl FixSession {
     fn new(sender: &str, target: &str) -> Self {
         Self::new_with_logon(sender, target, FixLogon::None)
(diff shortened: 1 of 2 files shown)


0-jake-0 (approved):
Thanks for this — merging it.

The diagnosis is right and the fix lands in the right place. Both socket write paths are covered (`write_encoded` and `write_at_seq` are the only socket `write_all`s in the file, so nothing outbound is left tearing), the test genuinely exercises a partial write followed by `WouldBlock` rather than just asserting the happy path, and the `fix/CLAUDE.md` sentence that contradicted the code was actually corrected rather than left behind. The verification list being explicit about what *wasn't* run is appreciated — the `lint-all` failure you hit is the known Aeron/CMake>=3.30 constraint documented in `CLAUDE.md`, not anything to do with this change.

One follow-up we'll take ourselves, and it is a flaw in how #820 was written rather than in this PR: the issue offered busy-retry as an option and called it "consistent with spin mode's design". That holds inside the spin cycle, but not on three paths the issue failed to think through:

- `SpinLogoutGuard::drop` and `FixSenderGuard::drop` both do `let _ = send_logout(..)` — written to tolerate a failed write. With an unbounded retry that path can no longer fail, only spin, so a stalled counterparty at teardown pins the graph thread inside a destructor.
- In `SpinState::cycle` the same thread does the read phase and the writes, so while a write spins we stop draining our receive buffer. If the peer stalls because *it* is blocked writing to us, that is a mutual zero-window stall, and TCP's persist timer probes it indefinitely rather than timing out.
- `fix_send` is an `Activation::NONE` sink that sets the socket non-blocking unconditionally, so it can run in a timer-driven graph where a full-tilt spin starves every other node. The latency argument for spinning does not apply there.

So we'll move to the issue's second option — retain the unwritten suffix on the session and complete it at the top of the next cycle before anything else is sent — which fixes the tearing without giving up the read phase or the heartbeat clock. The back-pressure wording in `fix/CLAUDE.md` will change again with it.

None of that blocks this. It is a strict improvement over what is on `main` today: a torn frame with a Logout appended after it is the worse failure for a live session. Follow-up issue linked below.

---

PR #844, opened 2026-08-15
fix(cache): touch LRU metadata without rewriting payload

### LINKED ISSUE #825: cache: every hit rewrites the whole file, non-atomically, just to bump mtime

`FileCache::get`'s LRU "touch" is `tokio::fs::write(&path, &bytes)` (`crates/wingfoil/src/adapters/cache.rs:203-206`) — a full in-place rewrite of the entire cache file on **every hit**, purely to feed the mtime-sorted eviction in `put`. Two problems:

1. **Doubled I/O on exactly the path the cache exists to make cheap** — a backtest slice cache can be hundreds of MB, and each hit pays a full write on top of the read.
2. **It silently undermines `put`'s documented atomicity.** `put` carefully uses tmp+rename so "the `.cache` file is never left in a torn state" (`cache.rs:216`, `239-244`) — but a crash or concurrent reader mid-touch sees a truncated file, which then silently degrades to re-fetch-every-run (`kdb_read_cached` treats read errors as misses, `crates/wingfoil/src/adapters/kdb/read_cached.rs:160`).

Nothing depends on the rewrite (the code comment itself says "unchanged bytes").

**Fix:** bump mtime without rewriting — `spawn_blocking` + `std::fs::File::options().append(true).open(&path)?.set_modified(SystemTime::now())`, keeping the existing ignore-errors semantics. `File::set_modified` is stable since Rust 1.75; the workspace MSRV is 1.88 (root `Cargo.toml:18`).

maxi-maxima:
## What this changes

Replaces the cache-hit full payload rewrite with a metadata-only mtime touch on Tokio blocking workers. The cache keeps the existing best-effort touch semantics and LRU ordering without doubling hit-path I/O or exposing a transiently truncated file to concurrent readers.

The cache adapter guidance now records the metadata-only touch invariant.

## Why

Closes #825.

A cache entry can be hundreds of megabytes, so rewriting identical bytes on every hit defeats the point of the cache. It also bypasses the atomic tmp-and-rename discipline used by `put()`.

## How it was verified

- [ ] `cargo lint` and `cargo lint-all` (local Windows host has no MSVC `link.exe`; deferred to CI)
- [ ] `cargo test -p wingfoil --features cache --test cache_adapter` (same local linker prerequisite; deferred to CI)
- [x] Isolated `rustc --emit metadata` API check for `OpenOptions::append(...).set_modified(...)`
- [x] Existing `test_lru_eviction` exercises the metadata touch and verifies the touched entry survives eviction

## Notes for the reviewer

The touch remains intentionally best-effort: a failure to update metadata does not turn an already successful cache read into an error. This PR is draft until the upstream Linux checks provide the unavailable test and clippy evidence.

## AI disclosure

AI-authored under human direction and review. Model: OpenAI Codex.

### DIFF
--- a/crates/wingfoil/src/adapters/cache.rs
+++ b/crates/wingfoil/src/adapters/cache.rs
@@ -200,10 +200,17 @@ impl<T> FileCache<T> {
             .ok_or_else(|| anyhow::anyhow!("cache file missing header newline"))?;
         let raw: Vec<(u64, T)> = bincode::deserialize(payload)?;
 
-        // Touch mtime so LRU eviction treats this file as recently used.
-        // We rewrite the unchanged bytes; any IO error is silently ignored
-        // since the data was already read successfully.
-        let _ = tokio::fs::write(&path, &bytes).await;
+        // Touch mtime so LRU eviction treats this file as recently used without
+        // rewriting (and briefly truncating) the cached payload. Keep this
+        // best-effort: the data was already read successfully.
+        let touch_path = path.clone();
+        let _ = tokio::task::spawn_blocking(move || {
+            std::fs::OpenOptions::new()
+                .append(true)
+                .open(touch_path)?
+                .set_modified(SystemTime::now())
+        })
+        .await;
 
         info!("cache hit: {}", path.display());
         Ok(Some(
--- a/crates/wingfoil/src/adapters/cache/CLAUDE.md
+++ b/crates/wingfoil/src/adapters/cache/CLAUDE.md
@@ -51,6 +51,10 @@ cache = ["dep:sha2", "dep:bincode", "dep:serde", "async", "tokio/fs"]
   not abort a backtest.
 - **LRU eviction is by file mtime**, applied so total on-disk size stays under
   `max_size_bytes`.
+- A cache hit advances that mtime with a metadata-only touch. **Never rewrite
+  the payload to mark a hit**: entries can be hundreds of megabytes, and an
+  in-place rewrite both doubles hit-path I/O and exposes a transiently
+  truncated file to concurrent readers.
 - **The key does not encode the run window.** It is derived from the query
   string, so the cache stores the *full* slice a query returned. Any window
   clamping belongs in the caller, on emit, on hits **and** misses — see how


0-jake-0:
Merged as 1f836d5 — thank you, @maxi-maxima, this was a clean first contribution.

A few things that made it easy to review, worth naming so they carry into the next one:

- **The scope was exactly the fix.** Two files, 15 lines, no opportunistic tidying alongside it. That matches how the other adapter fixes land here (#839, #842 have the identical `adapters/<name>.rs` + `adapters/<name>/CLAUDE.md` shape), so there was nothing to argue about.
- **You recorded the invariant in the adapter's `CLAUDE.md`, not just the code.** That is the part that actually stops the rewrite coming back, and it is easy to skip.
- **You were explicit about what you could not verify.** Flagging the missing MSVC `link.exe` and marking the PR draft pending CI is the right call — much better than a checkbox ticked on trust.

On that last point: the evidence you were waiting for came back green. CI passed `Lint (fmt & clippy)`, `Test (wingfoil)` and the audits, and I also ran the two deferred commands locally against your head commit — `cargo lint` clean, and `cargo test -p wingfoil --features cache --test cache_adapter` at 12/12 including `test_lru_eviction`. I additionally confirmed with a throwaway test that after a hit the file is byte-identical, the mtime advances, and the entry still deserialises.

Two notes for your interest:

- `.append(true)` works on Windows too, which you had no way to check from a host that could not link. Rust maps `append` to `FILE_GENERIC_WRITE & !FILE_WRITE_DATA`, which still carries `FILE_WRITE_ATTRIBUTES`, so `set_modified` succeeds there. It is also the better choice over `.write(true)` for a reason worth keeping: it structurally cannot truncate, which is the exact failure being fixed.
- The change quietly improves one more thing your description does not claim. The old `tokio::fs::write` would **recreate** the cache file if `evict_lru` had deleted it between the read and the touch, resurrecting an entry eviction had just reclaimed. `open()` now returns `NotFound` and the touch no-ops. I folded that into the squash commit message so it is on the record.

No new test was needed here and that was the right judgement, though it is worth understanding why: a rewrite writes back the identical bytes it just read, so no assertion on content, size, or mtime can tell the two implementations apart — the difference is I/O volume, which a unit test cannot see. `test_lru_eviction` already covers the LRU semantics the touch feeds. The `CLAUDE.md` note is the real guard, which is why adding it mattered.

Thanks also for the AI disclosure. The repo does not ask for one, so that was your own initiative and it is appreciated.

---

PR #846, opened 2026-08-15
feat(ops): add skip combinator

### LINKED ISSUE #797: Add `skip(n)` combinator

**What:** `skip(n)` suppresses the first `n` values, then passes everything through.
**Why:** `limit(n)` (= `take`) is the only member of this family present. `skip` is its mirror and one of the combinators every stream library is expected to have.

### Semantics

Ticks `Quiet` for the first `n` values (nothing propagates downstream), then `Tick::Value` for every value after that. The count is of *values seen*, not of engine cycles.

### Shape

| | |
|---|---|
| `Cfg` | `u32` — the number to skip |
| `State` | `u32` — how many have been seen |
| `In<'a>` | `(&'a T,)` |
| `Out` | `T` |
| Attribute | `#[op(build = skip, fluent)]` |

**Reference op — copy its structure:** `Limit`, `crates/wingfoil/src/ops.rs:389`. This is `Limit` with the branch inverted.

### Sketch

```rust
fn cycle(cfg: &mut u32, state: &mut u32, input: (&T,), _ctx: &mut Ctx<'_>) -> Result<Tick<T>> {
    if *state < *cfg {
        *state += 1;
        Ok(Tick::Quiet)
    } else {
    }
}
```

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = skip, fluent)]`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_skip!(T)`)
- [ ] Exercised in a `nitro!` block in `crates/wingfoil/tests/op_completeness.rs` — that file is the two-sided guard proving the op has *both* a fluent method and a compiled forwarder, and it asserts `interpreted() == compiled()`
- [ ] Rustdoc on both the op struct and the fluent method

### Getting started

The full recipe is `/new-op skip` (`.claude/commands/new-op.md`) — it walks the op shape, the `#[op]` attribute flags, the fluent method, `nitro!`/compiled coverage and the tests. Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md).

PozziTiv4ik:
## What this changes

Adds the core `skip(n)` combinator, which suppresses the first `n` values and passes every later value through. It is available through both `StreamOps` and `Signal`, with generated interpreted/compiled wiring.

The regression coverage pins exact values and tick times across interpreted, compiled, and nested execution, and the completeness guard exercises both the quiet and pass-through paths.

## Why

Closes #797.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo lint`
- [x] `cargo test -p wingfoil`
- [x] Exact pre-commit hook: workspace formatting + default-feature Clippy
- [x] Exact pre-push hook: `cargo build --workspace --all-targets` + `cargo test --workspace`
- [x] New behaviour is covered by a test asserting values and tick times across all three engines
- [ ] `cargo lint-all` / `cargo test -p wingfoil --all-features` — the local Windows GNU build is blocked in existing optional native adapters (`rdkafka-sys`/`zmq-sys`); Linux CI provides this coverage

## Notes for the reviewer

No Python binding is included: `skip` has no legacy Python twin, and this issue is scoped to the core Rust `StreamOps` surface.


0-jake-0:
Merged — thank you, @PozziTiv4ik. This follows the `/new-op` recipe closely, which is not a given for a first op: the `Op` mirrors `Limit` exactly (`Cfg`/`State` as `u32`, `ACTIVATION::NONE`, `Tick::Quiet`), `skip(0)` falls out correctly, and `*state` can't overflow since it only increments while `< cfg`. You also remembered the `__wf_signal_skip!(T)` line in `signal.rs` — step 4b, the one the skill flags because the facade silently fell 15 methods behind the catalog once. Updating the port-plan row unprompted is appreciated too.

The `op_completeness` comment checks out, for the record: I traced that chain to `[8, 8]` reaching `skip(1)`, so both the suppressed and pass-through paths really are exercised on interpreted and compiled.

One follow-up, opened as #848, closing the coverage gap rather than fixing anything: there's no `tests/catalog*.rs` test, where `limit` has both `catalog.rs::limit_caps_ticks` and `signal.rs::legacy_limit_passes_first_n`. That left the boundaries of the count untested and the `__wf_signal_skip!` line you added unexercised. #848 adds three tests — values and tick times for the main case, `skip(0)` and skip-past-the-run for the boundaries, and the facade twin.

One thing still genuinely open, which is a product decision rather than a gap: your note about no Python binding follows the skill's rule (state the choice), so nothing was violated. But `limit`, the op this mirrors, *is* bound in six lines of plain forwarding (`graph.rs:426`, `python.rs:185`). Worth deciding on the merits rather than by legacy-parity default — happy to take it either way.

---

Issue #837, opened 2026-08-14
Post-cutover wording sweep: stale "legacy" / "next" / "future" / "prototype" references


0-jake-0:
From the fable-review-2 codebase review. All verified; each is a one-or-two-line wording fix, bundled here as one sweep.

1. **User-facing error cites the deleted legacy engine.** The out-of-order `send_at` error (`crates/wingfoil/src/interp.rs:294-297`) ends with "(legacy errors on out-of-order)" — meaningless to a downstream user and runtime-triggerable. State the rule; the porting rationale already lives in the code comment at `interp.rs:289-291`.
2. **`Message::ValueAt` doc says the historical replay receiver is "future"** (`crates/wingfoil/src/channel.rs:38-40`) — it exists and is the fully-documented core of the channel layer (module doc `channel.rs:14-18`, `send_at` doc `:139-146`, implementation `interp.rs:279-299`). Reword: replayed deterministically at `time` in a historical run; timestamp ignored in realtime.
3. **`common.rs` slicer-gate docs are stale** (`crates/wingfoil/src/adapters/common.rs:39-42`): "gated on `postgres` today, and the kdb port adds its own feature to the gate later" / "`kdb_read` when it ports" — the gate is already `#[cfg(any(feature = "postgres", feature = "kdb"))]` (`common.rs:247`, `288`, `369`) and kdb is ported. This paragraph is the pointer a third time-sliced adapter follows; a stale account invites the duplication the module exists to prevent.
4. **PyPI description still says "prototype"** (`crates/wingfoil-python/pyproject.toml:8`): "Python interop prototype for wingfoil" ships on every wheel of a feature-complete, post-cutover binding. Reword in the next `bump.yml` cycle.
5. **13 integration workflows still say "main/next" in concurrency comments** (`aeron-integration.yml:15`, `etcd-integration.yml:14`, `web-integration.yml:19`, and 10 more — "pushes to main/next are not [cancelled]"), contradicting CLAUDE.md's record that `next` was stripped 2026-08-12. Comment-only; mechanical sweep.

Issue #835, opened 2026-08-14
CI coverage gaps: wingfoil-derive tests and doctests never run; python-test paths filter omits itself and wire-types


0-jake-0:
From the fable-review-2 codebase review (verified by reading every workflow's test invocations). Successor to #449's theme; distinct residual holes.

**1. `wingfoil-derive`'s unit tests never run in CI.** The only non-integration Rust test invocations are `cargo nextest run -p wingfoil --all-features --lib --tests` (`.github/workflows/rust-test.yml:106`) and `cargo test -p wingfoil-python` (`python-test.yml:97`); the bare `cargo test` at `web-integration.yml:143` runs in `crates/wingfoil-wasm` only. The derive crate has a real `#[cfg(test)] mod tests` (`crates/wingfoil-derive/src/lib.rs:3895`, 3 tests — pass locally, never executed in CI).

**2. Doctests never run in CI.** nextest does not run doctests, and no workflow invokes `cargo test --doc`; the wingfoil crate has ~40 runnable doc fences. CLAUDE.md's pre-commit checklist (`cargo test -p wingfoil --all-features`) also misses them — only the skippable local `.cargo-husky` pre-push hook (`cargo test --workspace`) covers either gap, so a broken derive test or doctest merges green.

**3. `python-test.yml`'s paths filter omits the workflow file itself and `crates/wingfoil-wire-types/**`** (`python-test.yml:7-16` lists only the three python-side crates). The wire-types dependency is real: CI builds with `all-adapters` (`python-test.yml:97,109`) → `web` (`wingfoil-python/Cargo.toml:100,134`) → `wingfoil/web-tls` → `dep:wingfoil-wire-types` (`wingfoil/Cargo.toml:405-413`). `rust-test.yml:5-8` documents the self-omission trap as a standing ruling ("a change to this file was the one change CI never validated") and removed its own filter for that reason; python-test kept the hole — a wire-types change currently merges without the Python leg running.

**Fix:** add `cargo nextest run -p wingfoil-derive` and a `cargo test --doc -p wingfoil` step to `rust-test.yml`; add `.github/workflows/python-test.yml` and `crates/wingfoil-wire-types/**` to both of python-test's paths lists.

Issue #819, opened 2026-08-14
`run_dynamic` bypasses the single-run/reset guard — silent stale-state reruns


0-jake-0:
From the fable-review-2 codebase review (finding adversarially verified against the code).

`Runner::run` guards re-runs at `crates/wingfoil/src/interp.rs:2871-2883` — `has_run` → `reset()`, or bail on a single-run graph. `run_dynamic` (`interp.rs:3424-3465`) neither sets nor checks `has_run` and never calls `reset()`. Nothing documents it as single-use ("Run the graph like `run`…", `interp.rs:3418`), and both are public.

**Failure modes (verified):**
- `finished` is set only by a channel `EndOfStream` (`interp.rs:960`) and cleared only in `reset()` (`interp.rs:2998`). On a historical-channel graph, `run()` then `run_dynamic()` hits `self.finished.get()` at `interp.rs:3513` and returns `Ok(())` having done nothing.
- On re-runnable graphs, any ordering (`run`→`run_dynamic`, `run_dynamic`→`run_dynamic`, `run_dynamic`→`run` — `has_run` stays false) silently continues stale fold/accumulator state.

This is exactly the failure `run`'s guard exists to prevent ("a second `run` returns an error rather than silently producing wrong values"), and it is drift produced by duplication: the validation + kernel-construction preamble (`interp.rs:2886-2925` vs `3436-3465`) and the stop/teardown loops (`2930-2969` vs `3558-3581`) are verbatim copies differing only in the `removed[]` tombstone skip.

**Fix:** make the deduplication the fix vehicle — extract a shared `make_kernel(run_mode, run_for) -> Result<Kernel>` preamble and a tombstone-aware lifecycle helper (safe for both: `removed` is an unconditional field, initialized all-false at `interp.rs:2633`, so the skip is a no-op on static runs), and set/check `has_run` in `run_dynamic`.

Issue #461, opened 2026-07-11
Harden supply-chain controls (cargo-deny, action pinning, provenance) beyond vulnerability scanning


0-jake-0:
## Context

Known-vulnerability scanning is being added in #460 via `.github/workflows/security-audit.yml`:

- `cargo audit` — Cargo.lock vs the RustSec advisory DB (fails on vulnerabilities)
- `pnpm audit` — `wingfoil-js` tree (fails on moderate+)
- `dependency-review` — blocks PRs that introduce a dep with a known advisory (GitHub Advisory DB)
- weekly schedule so advisories disclosed against already-pinned deps are caught

That covers *known* advisories. This task tracks the remaining **supply-chain attack** controls — the ones that defend against malicious/compromised dependencies and CI, which advisory scanning alone does not.

### High value / low effort
- [ ] **Adopt `cargo-deny`** (superset of `cargo audit`). Add a `deny.toml` with:
  - `[sources]` — restrict crates to crates.io; deny unknown registries and git sources (guards against dependency-confusion / source injection).
  - `[bans]` — deny wildcard version reqs, optionally flag `multiple-versions`, and provide a hook to ban a specific crate quickly.
  - `[advisories]` — keep the vulnerability gate; explicitly `ignore` the reviewed, no-fix informational advisories (currently unmaintained/unsound: `bincode`, `paste`, `mach`, `instant`, `memmap`, `dotenv`, `rustls-pemfile`, `anyhow` RUSTSEC-2026-0190, `lru` RUSTSEC-2026-0002) so they are documented rather than silently passing.
  - `[licenses]` — allowlist to catch a dependency silently introducing an incompatible license.
- [ ] **Add README status badges** so the security posture is visible at a glance:
  - `security.audit` workflow status — `[Security Audit](https://github.com/wingfoil-io/wingfoil/actions/workflows/security-audit.yml/badge.svg)`
  - Dependabot status — a shields.io badge (e.g. `https://img.shields.io/badge/Dependabot-enabled-brightgreen`) once `.github/dependabot.yml` lands, or link the badge to the repo's security/Dependabot page.
  - (If OpenSSF Scorecard is adopted below, add its badge here too.)
- [ ] **Pin GitHub Actions to full commit SHAs**, not mutable tags (`@v4`). A compromised/retagged action is a direct CI-execution supply-chain vector. Consider `pinact`/`ratchet` + Dependabot `github-actions` updates to keep the SHAs fresh.
[...]

PR #850, opened 2026-08-16
Clear the 9.0.0 release blockers: browser payload corruption, run_dynamic's missing guard, CI holes

### LINKED ISSUE #821: web/wasm: browser payload path is broken under the default bincode codec — both directions
From the fable-review-2 codebase review. Verified **empirically** with a scratch crate against the tree's exact bincode version (1.3.3).

**Client → server (silent corruption):** `encode_payload`'s bincode leg serializes the JS value as a `serde_json::Value` (`crates/wingfoil-wasm/src/lib.rs:104-108`), but the server decodes with `codec.decode::<T>(&payload)` (`crates/wingfoil/src/adapters/web/read.rs:79`). Bincode is non-self-describing, so a JS object encodes as a length-prefixed map (keys included) while `deserialize_struct` expects bare fields in declaration order. The result is not an error — it is **silent garbage**: in the empirical test, `{"button":"ok","count":1}` decoded as `UiClick { button: "\u{6}\0", count: 0 }`, and a JS integer routed through `Value`'s `f64` decoded as `2.5e-323`.

**Server → client (total failure):** the decode side (`wingfoil-wasm/src/lib.rs:135-138`) does `codec.decode::<serde_json::Value>()`, which under bincode always fails with `DeserializeAnyNotSupported` — under the default codec the JS client cannot decode any data payload at all.

No translation layer exists (`wingfoil-wire-types/src/lib.rs:106-122` is a straight bincode/serde_json dispatch). The Rust integration test `sub_round_trip_bincode` (`crates/wingfoil/tests/web_integration.rs:234`) encodes natively and never exercises the wasm path. Only `js/src/tracing.ts:16-18` and `js/README.md:125-126` document a narrow JSON-only caveat, while the README's main example (~lines 29-37) shows `client.publish("ui", {...})` under the default bincode — a documented-as-working path that corrupts.

**Fix:** either reject bincode in `encode_payload`/`decode` with a clear error pointing at the JSON codec, or route browser payloads as JSON regardless of the frame codec. Either way, add a cross-language round-trip test (wasm encode → server typed decode, and the reverse) to `web-integration.yml`, and promote the tracing.ts caveat into the client README and `encodePayload` docs.

0-jake-0:
## What this changes

Clears the issues that should not ship in 9.0.0: a silently-corrupting browser payload path, a public run entry point that skipped the re-run guard, two CI holes that let changes merge without the thing they broke ever running, and a lockfile pin on a yanked crate.

Closes #821, #819, #835. Addresses #837 (item 4 only) and the security half of #452.

## Why

These came out of a triage of all 47 open issues against the tree. 9.0.0 is not published yet — the last tag is `v8.0.0` — so these are the last things standing between the tree and a first publish to three registries.

**#821 — browser data payloads corrupt silently under the default codec.** `encode_payload` serialized the JS value as a `serde_json::Value`, but the server decodes with `codec.decode::<T>()`. bincode is schema-driven and non-self-describing, so a JS object encodes as a length-prefixed map with keys inline while `deserialize_struct` expects bare fields in declaration order. The server did not error — it decoded garbage: `{"button":"ok","count":1}` arrived as `UiClick { button: "\u{6}\0", count: 0 }`. The other direction always failed with `DeserializeAnyNotSupported`. `js/README.md`'s quick start demonstrated exactly this, under the default codec.

A JS client cannot produce a bincode encoding of a Rust struct — there is no schema in the browser — so the path is impossible rather than merely buggy. Both legs now fail loudly with a message naming the call, `CodecKind::Json` and `codec: "json"`, and the docs stop promising otherwise.

**#819 — `run_dynamic` bypassed the single-run/reset guard.** `run` guards re-runs so that "a second `run` returns an error rather than silently producing wrong values"; `run_dynamic` neither set nor checked `has_run` and never called `reset()`, though both are public. On a historical-channel graph, `run()` then `run_dynamic()` returned `Ok(())` having run zero cycles; on re-runnable graphs every ordering silently continued stale fold state. The cause was duplication — the preamble and teardown were verbatim copies — so the dedup is the fix vehicle.

**#835 — CI never ran the derive tests or any doctest**, and `python-test.yml`'s paths filter omitted both itself and `crates/wingfoil-wire-types/**`, so a wire-types change could merge without the Python leg running at all.

## How it was verified

Full gate run over the combined branch, not per-commit:

- [x] `cargo fmt --all`
- [x] `cargo lint` and `cargo lint-all` — both clean
- [x] `cargo test -p wingfoil --all-features` — **985 passed**

64 failures, all confined to the nine container-dependent suites (`aeron`, `etcd`, `fluvio`, `kafka`, `otlp`, `postgres`, `redis`, `zmq_cross_lang`, `zmq_etcd`) failing on `Socket not found: /var/run/docker.sock` — no daemon in the dev sandbox. These run in their own workflows here. Every other binary is green, including `web_integration` 11/11, `rerun_dynamic` 7/7, `engine_semantics` 8/8 (untouched), and 17 doctests.

Also verified: the two new CI steps pass locally on this tree, so they go green rather than red on merge (`cargo test -p wingfoil-derive` → 3 passed; `cargo test --doc -p wingfoil --all-features` → 17 passed, 18 ignored). `wasm32-unknown-unknown` builds; 8/8 wasm unit tests; `pnpm lint` and 27/27 JS tests.

The #819 tests were checked for bite, not just for green: with the new `begin_run()?` temporarily removed from `run_dynamic`, 6 of the 7 new tests fail.

## Notes for the reviewer

**Routing browser payloads as JSON regardless of frame codec was considered and rejected.** `read.rs`/`write.rs` both encode payloads with the frame codec, and real bincode payload peers exist (the `sub_round_trip_bincode` test and the Python binding), so it would need a per-frame signal the envelope has no field for — a wire-version bump. Failing loudly is the release-safe fix; the richer protocol change can come later.

**#819 documents a caveat it did not change.** Re-running a *mutated* graph restores only the statically-built region, because `rt_append_node` gives dynamically-appended nodes a no-op `reset` by construction. That is an existing design decision, previously uncommented in the public contract; it is now in the rustdoc and pinned by a test.

**`spin` 0.9.8 → 0.9.9 is a yank fix, not an advisory fix.** CI's `cargo audit` exits 0 either way, but shipping a release whose committed lock pins a yanked crate is a reproducibility wart. Trivially revertible if you would rather the release PR carry zero lockfile churn.

**#452 is not a release blocker and its body is wrong.** Re-derived against a freshly fetched RustSec DB: 0 vulnerabilities in the root lock (750 crates), 0 in `crates/wingfoil-wasm/Cargo.lock`, and "no known vulnerabilities" from `pnpm audit`. Everything remaining is RustSec *informational*, which does not fail CI. The issue's "1 critical, 6 high" is stale by two rounds of remediation. Caveat: the Dependabot alert list itself could not be read from this session (no `gh`, and `api.github.com` is proxy-blocked), so that is inference from the tree — worth a glance at the security tab before relying on it.

**#835's body contains an error worth not propagating.** It claims CLAUDE.md's local checklist misses doctests. Plain `cargo test` runs the `--doc` target by default and `wingfoil` sets no `doctest = false`, so the existing `cargo test -p wingfoil --all-features` already covers the fences locally. Only *nextest* skips them, which is why CI needed a separate step but the local checklist only needed the derive line.

### Follow-ups deliberately not folded in

- **#461's accepted-advisory list needs a refresh**: drop `dotenv` (gone from the tree) and `anyhow` RUSTSEC-2026-0190 (lock is at patched 1.0.104); add `lru` RUSTSEC-2026-0253, sibling of the RUSTSEC-2026-0002 already listed and unfixable from the lockfile (`lru >= 0.18.2` is a semver-major move pinned upstream by `rv`/`augurs`).
- `security-audit.yml` justifies its pytest allowance by citing `uv.lock`, which no longer exists. The allowance is still correct; the sentence naming the file is not.
- The remaining four items of #837 (the wording sweep) are untouched; only the PyPI description is fixed here, because that string ships in wheel metadata.

---
_Generated by [Claude Code](https://claude.ai/code/session_01H3wQjYrfk8RpF3TQFjeii7)_

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -2,18 +2,40 @@ name: test.python
 
 on:
   workflow_call:
+  # This leg keeps a `paths` filter where rust-test.yml dropped its own: a
+  # maturin wheel build plus pytest is the slowest thing in the matrix, and
+  # most commits genuinely cannot affect it. But a filter is only ever as good
+  # as its list, and two entries were missing from this one:
+  #
+  # * **this file itself.** A filter that does not name its own workflow makes
+  #   an edit to the workflow the one change CI never validates — a broken gate
+  #   surfaces only after it has merged. That is the standing ruling recorded
+  #   at the top of rust-test.yml, which removed its filter outright rather
+  #   than live with the trap.
+  # * **`wingfoil-wire-types`.** The bindings depend on it, just not by name:
+  #   both build steps below pass `all-adapters`, which turns on `web`, which
+  #   turns on `wingfoil/web-tls`, which pulls in `dep:wingfoil-wire-types`. A
+  #   wire-format change could therefore break the wheel with this leg never
+  #   having run.
+  #
+  # `push` and `pull_request` are independent filters and GitHub will not warn
+  # when they drift, so any addition goes in both lists.
   push:
     branches: [ "main" ]
     paths:
+      - ".github/workflows/python-test.yml"
       - "crates/wingfoil-python/**"
       - "crates/wingfoil-python-derive/**"
       - "crates/wingfoil/**"
+      - "crates/wingfoil-wire-types/**"
   pull_request:
     branches: [ "main" ]
     paths:
+      - ".github/workflows/python-test.yml"
       - "crates/wingfoil-python/**"
       - "crates/wingfoil-python-derive/**"
       - "crates/wingfoil/**"
+      - "crates/wingfoil-wire-types/**"
   workflow_dispatch:
 
 # One in-flight run per ref. Superseded PR pushes are cancelled; pushes to
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -108,6 +108,43 @@ jobs:
         env:
           RUST_LOG: INFO
 
+      - name: Run wingfoil-derive tests
+        # `nitro!` reasons about the wiring function it is handed — which
+        # parameters a type mentions, and therefore which streams a node
+        # depends on — and that reasoning has unit tests of its own in
+        # `crates/wingfoil-derive/src/lib.rs`. Every other Rust test invocation
+        # across these workflows names its package — `-p wingfoil` here and in
+        # the integration legs, `-p wingfoil-python` in test.python, and
+        # wingfoil-wasm's own leg in web-integration.yml — and none of them
+        # names this one, so the derive tests had never run in CI at all: a
+        # regression in the macro's own logic could only be caught indirectly,
+        # by whichever generated graph happened to come out wrong.
+        #
+        # The crate is proc-macro-only and declares no features, so a bare `-p`
+        # is its whole surface, and it compiles in seconds — syn/quote are
+        # already built for the step above.
+        run: cargo nextest run -p wingfoil-derive
+
+      - name: Run doctests
+        # **nextest cannot run doctests**, by design: it is a libtest-harness
+        # runner, and `--doc` targets are compiled and driven by rustdoc
+        # instead. So until this step existed, no job in any workflow ran a
+        # single doc fence — and the fences are exactly the snippets users copy
+        # out of the API docs, which a renamed method or a changed signature
+        # rots silently.
+        #
+        # `--all-features` is not optional here. Most of the fences live in
+        # `adapters/`, and every adapter module is gated off by default: a
+        # default-feature run collects 12 of them, all-features collects 35 (17
+        # executed, the rest `ignore`d illustrations). It also means this step
+        # shares its feature set — and so its artifacts — with the step above.
+        #
+        # Cheap despite the count: rustdoc compiles the fences as one merged
+        # binary, ~2s, and running them takes hundredths of a second.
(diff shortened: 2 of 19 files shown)


0-jake-0:
Superseded — split into one PR per issue so they can be reviewed and merged independently, and rebased onto `ea46466` to pick up #851.

| PR | Closes | Scope |
|---|---|---|
| #852 | #821 | Browser + Python payload corruption under bincode |
| #853 | #819 | `run_dynamic`'s missing single-run/reset guard |
| #854 | #835, #837 (item 4) | Derive tests + doctests in CI, `python-test` paths holes |
| #855 | #452 (security half) | `spin` off the yanked 0.9.8 |

Same five commits, same verification, no content changes. The #821 fix and its Python sibling stayed together in #852 because both edit `adapters/web/mod.rs` and the Python fix only makes sense on top of the browser one.

Note that #851 landed **#832** (the `#[op]` passive-mask range check under `no_builder`) while this was open, so that item is already done upstream — and it makes #854 more valuable, since `wingfoil-derive` just gained a trybuild fixture in a crate CI still never tested.

Closing in favour of the four above.

---

PR #851, opened 2026-08-16
fix(derive): validate passive masks for no-builder ops

### LINKED ISSUE #832: #[op]: passive-mask range check is skipped under no_builder — silent wrong dispatch const
From the fable-review-2 codebase review (finding adversarially verified by reading both macro code paths).

The `passive` mask's bit-range validation lives only in `expand_builder` (`crates/wingfoil-derive/src/lib.rs:2230-2235`), but `expand_op` short-circuits `expand_builder` entirely for `no_builder` ops (`lib.rs:2149-2153`) — while the `__WF_OP_*_PASSIVE` const is emitted **unconditionally** in the forwarders (`lib.rs:2112-2114`) and consumed by compiled/nested dispatch (`lib.rs:3247`, `3489`).

So `#[op(build = x, no_builder, passive = [7])]` on a 1-edge op compiles cleanly and emits a mask whose out-of-range bit silently deactivates nothing — a silent scheduling divergence between tiers, the exact class of bug the forwarder design exists to prevent. Latent today (no in-tree op combines `no_builder` with `passive`: `with_time`, `poll`, and the latency stamps don't), but user crates write `no_builder` ops too.

**Fix:** hoist the `passive >> n != 0` check from `expand_builder` into `expand_op`, right after `parse_in_shape` (the edge count is known there), so it runs on every path. While touching it, delete the vacuous `n < 32 &&` guard — `n > 26` already errored six lines earlier. Turning silent macro misbehavior into a compile error has no downside.

maxi-maxima:
## What this changes

Validate `#[op]` passive edge positions immediately after parsing the op's input shape, before the expansion can skip the generated Builder path. Add a compile-fail regression covering an out-of-range passive mask on a `no_builder` op.

## Why

`no_builder` ops bypass `expand_builder`, which was the only place that rejected passive bits beyond the declared input edges. Nitro forwarders still emit and consume the passive-mask constant, so an invalid mask compiled successfully and could silently diverge from the intended dispatch semantics.

Closes #832.

## How it was verified

- [x] `cargo fmt --all` and `cargo fmt --all -- --check`
- [x] `cargo lint` and `cargo lint-all` in the upstream `Lint (fmt & clippy)` workflow
- [x] `cargo test -p wingfoil --all-features` in the upstream `Test (wingfoil)` workflow, including the new trybuild fixture
- [ ] New behaviour is covered by a test asserting **values and tick times** — not applicable; this is a proc-macro compile-fail boundary
- [x] Added `tests/trybuild/passive_no_builder_out_of_range.rs` with the exact expected diagnostic
- [x] `git diff --check`
- [x] Python interop, Cargo/pnpm audits, and dependency review workflows

## Notes for the reviewer

The local Windows host has no MSVC `link.exe`, so local Rust compilation stopped before crate code; the complete upstream Linux lint and test workflows passed. `checked_shr` treats bits above a 32-or-more-edge input shape as absent, avoiding a shift overflow for `no_builder` ops while preserving the existing 26-edge limit on generated Builder methods.

AI assistance was used for call-path tracing and patch drafting; the final diff, issue match, and repository instructions were reviewed directly.

### DIFF
--- a/crates/wingfoil-derive/src/lib.rs
+++ b/crates/wingfoil-derive/src/lib.rs
@@ -1698,6 +1698,18 @@ fn expand_op(args: &OpArgs, imp: &ItemImpl) -> syn::Result<TokenStream2> {
     let state_ty = state_ty.ok_or_else(|| missing("State"))?;
     let out_ty = out_ty.ok_or_else(|| missing("Out"))?;
     let shape = parse_in_shape(&in_ty)?;
+    let edge_count = shape.edge_ref_tys.len();
+    if args
+        .passive
+        .checked_shr(edge_count as u32)
+        .unwrap_or_default()
+        != 0
+    {
+        return Err(syn::Error::new(
+            self_ty.span(),
+            format!("#[op]: `passive` names an edge position beyond this op's {edge_count} inputs"),
+        ));
+    }
     let activation_expr = activation_expr
         .ok_or_else(|| syn::Error::new(imp.span(), "#[op]: impl has no `const ACTIVATION`"))?;
 
@@ -2225,12 +2237,6 @@ fn expand_builder(args: &OpArgs, b: &BuilderShape<'_>) -> syn::Result<TokenStrea
         return Err(syn::Error::new(
             self_ty.span(),
             "#[op]: a generated Builder method supports at most 26 input edges",
-        ));
-    }
-    if n < 32 && args.passive >> n != 0 {
-        return Err(syn::Error::new(
-            self_ty.span(),
-            format!("#[op]: `passive` names an edge position beyond this op's {n} inputs"),
         ));
     }
 
--- a/crates/wingfoil/tests/trybuild/passive_no_builder_out_of_range.rs
+++ b/crates/wingfoil/tests/trybuild/passive_no_builder_out_of_range.rs
@@ -1,0 +1,25 @@
+use wingfoil::anyhow::Result;
+use wingfoil::op;
+use wingfoil::op::{Activation, Ctx, Op, Tick};
+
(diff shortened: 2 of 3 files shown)


0-jake-0:
Approved and merging — thank you, @maxi-maxima, this is a clean fix.

Verified locally on `d60c4c5`: `cargo test -p wingfoil --test trybuild` passes all 14 fixtures, and reverting `wingfoil-derive/src/lib.rs` to `main` while keeping the fixture makes `passive_no_builder_out_of_range.rs` fail — so the test is a genuine negative control, not a tautology. `cargo clippy -p wingfoil-derive --all-targets -- -D warnings` is clean.

Two things worth calling out:

- Hoisting the check into `expand_op` right after `parse_in_shape` puts it on the one path every `#[op]` takes, and `shape.edge_ref_tys.len()` is the same quantity `expand_builder`'s `n` was computed from, so there's no semantic drift.
- Declining #832's suggestion to drop the `< 32` guard outright was the right call, and the reasoning in the PR description is correct: that guard was vacuous at the old site because the 26-edge cap preceded it, but the cap is Builder-only and `no_builder` ops never hit it. `checked_shr` is what keeps a ≥32-edge op from panicking the macro on shift overflow, and it's sound in the other direction too since parse already rejects `pos >= 32`.

Non-blocking follow-ups, none worth holding the merge for: `tests/trybuild.rs`'s module doc and test name still say "`nitro!` macro" though the suite now pins an `#[op]` diagnostic, and the message's "this op's 1 inputs" grammar is now pinned in a `.stderr` (pre-existing wording; fixing it later just costs a `TRYBUILD=overwrite` regen).

---

PR #874, opened 2026-08-16
Enforce the no-panic and binding-registration rules, and fix three doc gaps found by a friction audit

0-jake-0:
## What this changes

Two rules the repo states but nothing checked are now enforced in CI, and three
documentation gaps are closed.

No engine behaviour changes. The only Rust in this diff is a crate-level lint
attribute.

> The `/dogfood` skill that produced these findings was **split out into #876,
> which has now merged**. This PR is the findings; #876 was the instrument.
> Rebased onto that `main`, so the skill is already in the base.

## Why

Eight small-model agents were run through `/new-op`, `/bind-adapter` and the
examples rules, each implementing a real gap in an isolated worktree and
recording where the repo made the work harder than it needed to be. The point
of using a deliberately weak model is that it hits under-specified steps
head-on instead of silently inferring past them.

Two findings were rules we write down but do not verify:

* **The no-panic rule.** An agent introduced the first `panic!` into
  `wingfoil-python` — a crate that had zero — in an "unreachable" match arm
  guarded by a check at the Python boundary. Neither the skill nor clippy
  objected. `CLAUDE.md` § "Error Handling" already forbids this.
* **Binding registration.** A new adapter binding needs six separate
  declarations. Missing one compiles perfectly and silently drops the adapter
  from the wheel, which is the worst failure shape available: green build,
  absent feature.

The three doc gaps were places contributors were sent somewhere wrong:

* An example whose README pins output must be deterministic. The realtime
  `error_handling` example printed a different interleaving on consecutive
  runs, so the "sample output must be real" rule and realtime mode are in
  direct conflict — whatever you paste is wrong next run.
* `#[op]`'s `State` is `Default`, so a latch that must start *on* needs a
  `start` hook. `take_while` and `skip_while` were both silently wrong with
  `bool::default()`, in opposite directions, and each looked plausible until a
  test asserted tick times.
* `/bind-adapter` told you to skip the feature gate when the engine module has
  no `#[cfg]`. That is wrong for `lines`: the module is unconditional but five
  items inside it are `#[cfg(feature = "async")]`, so the binding needs a
  feature after all — one that enables only `_common`, since there is no engine
  feature to turn on.

## How it was verified

Re-run after rebasing onto the current `main`, not just on the original base:

- [x] `cargo fmt --all`
- [x] `cargo lint` (workspace, all targets) — clean
- [x] `cargo clippy -p wingfoil-python --all-targets -- -D warnings` — clean
- [x] `scripts/check-example-docs.sh` — 46 targets
- [x] `scripts/check-python-bindings.sh` — 16 bindings
- [ ] `cargo lint-all` — **not run**, needs `protoc`, unavailable in this
      sandbox. Mitigated by scanning every feature-gated adapter binding for
      production `unwrap`/`panic`/`todo` (none found), so the `--all-features`
      leg should not newly fail on the added `deny`.
- [ ] `cargo test -p wingfoil --all-features` — **not run**; no engine
      behaviour changed and the crate is untouched by this diff.
- [ ] Values-and-tick-times test — not applicable, no op added.

**Both guardrails were verified against an injected violation, not just
observed green** — a planted `panic!` fails the clippy gate, and a dropped
`all-adapters` entry fails the checker.

## Notes for the reviewer

**Why the lint gate needs no CI step.** `#[cfg_attr(not(test), deny(...))]` in
`crates/wingfoil-python/src/lib.rs` rides the existing
`cargo clippy --workspace --all-targets` in `rust-test.yml` and the pre-commit
hook. `not(test)` is what makes it practical: the 300+ `.unwrap()`s in test
modules — which the policy explicitly permits — stay legal with no per-module
opt-outs. `expect` is deliberately absent from the deny list; the policy allows
it for documented invariants and the mutex-poisoning pattern depends on it.

**Why the checker is in the lint job, not `python-test.yml`.** That leg keeps a
`paths` filter, so a registration dropped by an edit outside those paths would
never be checked. The lint job has no filter.

**The checker encodes current practice, not a law.** It asserts every binding
reaches `_common`, true of all sixteen today. A future binding that legitimately
needs no async would require relaxing that check. The `aeron`/`iceoryx2` wheel
exemptions are in a `wheel_exempt` list with their reasons, so they stay
recorded decisions rather than silent holes.

**Prototype code was discarded.** The agents' patches are not in this diff —
they were the vehicle, not the product. Two were actively broken (one op failed
3 of 4 tests; another silently deviated from its spec and encoded the deviation
as a passing test). `.claude/friction-runs/2026-08-16/` keeps the logs, the
independently verified outcome per task, and — deliberately — the reported
findings that did **not** survive checking: `Burst`/`Activation`/`Ctx`/`Tick`
are all in the prelude, `#[op(fluent)]` is not broken, and `_common` was
already documented. Three of four op agents reported needing `cargo clean`;
that was fingerprint thrash from the shared `CARGO_TARGET_DIR` the harness
imposed, not a repo defect.

**Findings not acted on**, left for a follow-up rather than folded in:
documenting `GraphBuilder::custom_node` in `CLAUDE.md`'s Wiring section (two
agents reached for it because only `source`/`wire` are named there, and had to
reverse-engineer the `value_slot()` → `borrow()` pattern from test code);
adding `Op` to the prelude; adding `anyhow::{Context, Result}` to the prelude
so the mandated `.context()` resolves; and `NanoTime::as_nanos()`.

### DIFF
--- a/.claude/commands/bind-adapter.md
+++ b/.claude/commands/bind-adapter.md
@@ -88,6 +88,29 @@ gating a binding whose engine side is always compiled only creates a way to
 build a wheel that is missing it for no reason. Everything else in this file
 still applies.
 
+**But grep the module's *contents* before concluding that, because "the module
+is unconditional" and "everything in it is unconditional" are different
+claims.** `lines` is the case that breaks the shortcut: `pub mod lines;` in
+`crates/wingfoil/src/adapters/mod.rs` carries no `#[cfg]`, yet
+`replay_lines` / `replay_lines_scheduled` inside it are
+`#[cfg(feature = "async")]`. A binding that exposes those needs a
+`wingfoil-python` feature after all — not to gate the adapter, which is always
+there, but to reach `wingfoil/async` via `_common`. So:
+
+```toml
+# lines: tail/write/append are unconditional; read/read_scheduled need async.
+lines = ["_common"]   # no `wingfoil/lines` — there is no such feature
+```
+
+Note the shape: the feature enables **only** `_common`, because there is no
+engine feature to turn on. That is legitimate and is the tell for this case.
+The binding then gates in three places that must agree — the `use` imports, the
+function definitions, and the `register_adapters` block in `src/python.rs`,
+which splits into an unconditional part and a `#[cfg(feature = "lines")]` part.
+A mismatch between the three compiles cleanly and drops functions silently, so
+check with both `cargo check -p wingfoil-python` and
+`cargo check -p wingfoil-python --features <name>`.
+
 Each *adapter* binding lives behind a `wingfoil-python` cargo feature of
 the **same name** as the adapter, which turns on the matching engine feature:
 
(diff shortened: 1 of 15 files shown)


zaydmulani09 (commented):
Nice tightening — the no-panic enforcement and the binding-registration script
are exactly the kind of "documented but unchecked" gap that's worth catching
early, and check-python-bindings.sh's six-point breakdown is genuinely clear.

One friction-run finding didn't make it into the doc changes: take_skip_while.md
#2 flags that adding a new #[op(build = ..., fluent)] can require a `cargo clean`
before the generated __wf_fluent_* macros resolve (stale incremental/proc-macro
caching), and its own suggested fix is "document that clean rebuilds are needed."
That's the only one of the five friction items without a corresponding change in
new-op.md or CLAUDE.md — worth a one-line addition near the #[op] attribute
section so the next agent/contributor doesn't lose the same hour rediscovering it.

PR #879, opened 2026-08-18
Add pairwise combinator

### LINKED ISSUE #800: Add `pairwise()` combinator (zip-with-previous)
Part of #459. **Good first issue** — a near-clone of an existing op, ~20 lines plus tests.

**What:** `pairwise()` emits `(previous, current)` for every value after the first.
**Why:** `difference()` already does exactly this shape but immediately subtracts, so it is confined to `T: Sub`. `pairwise` is the general form — it lets a caller compute *any* function of consecutive values (ratios, direction changes, gap detection) on types that have no arithmetic at all.

### Semantics

First value: `Tick::Quiet` (no previous exists). Every value after: `Tick::Value((prev, current))`.

| value | emits |
|---|---|
| 10 | `Quiet` |
| 12 | `Value((10, 12))` |
| 7 | `Value((12, 7))` |

### Shape

| | |
|---|---|
| `Cfg` | `()` |
| `State` | `Option<T>` — the previous value |
| `In<'a>` | `(&'a T,)` |
| `Out` | `(T, T)` |
| Attribute | `#[op(build = pairwise, fluent)]` |

**Reference op:** `Difference`, `crates/wingfoil/src/ops.rs:265`. It is structurally identical — same `Option<T>` state, same quiet-on-first rule. `pairwise` tuples where `Difference` subtracts, and drops the `Sub` bound.

### Sketch

```rust
fn cycle(_cfg: &mut (), state: &mut Option<T>, input: (&T,), _ctx: &mut Ctx<'_>) -> Result<Tick<(T, T)>> {
    let value = input.0.clone();
    let out = match state.take() {
        Some(prev) => Tick::Value((prev, value.clone())),
        None => Tick::Quiet,
    };
    *state = Some(value);
    Ok(out)
}
```

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = pairwise, fluent)]` — bound only `T: Clone + 'static`, no `Sub`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_pairwise!(T)`)
- [ ] Test asserting exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`, covering the quiet first tick
- [ ] A test on a **non-arithmetic type** (e.g. `String` or an enum) — that is the capability `difference` cannot offer and the reason this op exists
- [ ] Rustdoc cross-referencing `difference` as the arithmetic shorthand

### Getting started

Run `/new-op pairwise` (`.claude/commands/new-op.md`). Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md). Branch from `main`, PR base `main`.

aryemans:
## Summary

- add `pairwise()` for emitting `(previous, current)` values after the first tick
- support non-arithmetic types without requiring `Sub`
- add fluent, compiled, and exact-timestamp coverage
- fix generated fluent bounds for generic tuple outputs such as `(T, T)`

## Testing

- `cargo test -p wingfoil pairwise`
- `cargo test -p wingfoil --test op_fluent_shapes`
- `cargo test -p wingfoil --test op_completeness`
- `cargo test -p wingfoil --test compiled_stateful_ops pairwise`
- `git diff --check`

`cargo test -p wingfoil --all-features` was also attempted locally; Kafka integration reached the Redpanda test but the container could not start because the host's Linux AIO capacity was insufficient.

Closes #800

### DIFF
--- a/crates/wingfoil-derive/src/lib.rs
+++ b/crates/wingfoil-derive/src/lib.rs
@@ -2677,7 +2677,16 @@ fn expand_fluent(args: &OpArgs, b: &BuilderShape<'_>) -> syn::Result<TokenStream
         let is_generic = matches!(ty, Type::Path(p)
             if p.path.get_ident().is_some_and(|id|
                 Some(id) == self_param.as_ref() || method_generics.contains(&id)));
-        if is_generic {
+
+        // tuples of generics
+        let is_generic_tuple = matches!(ty, Type::Tuple(tuple)
+        if !tuple.elems.is_empty()
+            && tuple.elems.iter().all(|elem| matches!(elem, Type::Path(p)
+                if p.path.get_ident().is_some_and(|id|
+                    Some(id) == self_param.as_ref()
+                        || method_generics.contains(&id)))));
+
+        if is_generic || is_generic_tuple {
             preds.push(sub(quote! { #ty: #bounds }));
         }
     };
--- a/crates/wingfoil/src/fluent.rs
+++ b/crates/wingfoil/src/fluent.rs
@@ -1156,6 +1156,13 @@ pub trait StreamOps<T>: Sized {
     where
         T: Clone + Default + Sub<Output = T> + 'static;
 
+    /// Emit pairs of successive values `(previous, current)`.
+    /// Quiet on the first value.
+    fn pairwise(&self) -> Stream<(T, T)>
+    where
+        T: Clone + 'static,
+        (T, T): Default + 'static;
+
     /// Negate each value (`!value`) — sugar over `map`.
     fn not(&self) -> Stream<T>
     where
@@ -1400,6 +1407,8 @@ impl<T: 'static> StreamOps<T> for Stream<T> {
     __wf_fluent_drop_small_change!(T);
 
     __wf_fluent_difference!(T);
+
+    __wf_fluent_pairwise!(T);
 
     __wf_fluent_not!(T);
 
--- a/crates/wingfoil/src/ops.rs
+++ b/crates/wingfoil/src/ops.rs
@@ -298,6 +298,40 @@ where
     }
 }
 
+/// Emits the successive pairs `(previous, value)`. Quiet on the first
+/// value.
+///
+/// See [`difference`](crate::fluent::StreamOps::difference) for the
+/// arithmetic shorthand.
+pub struct Pairwise<T>(PhantomData<T>);
+
+#[op(build = pairwise, fluent)]
+impl<T> Op for Pairwise<T>
+where
+    T: Clone + 'static,
+{
+    type Cfg = ();
+    type State = Option<T>;
+    type In<'a> = (&'a T,);
+    type Out = (T, T);
+    const ACTIVATION: Activation = Activation::NONE;
+
+    fn cycle(
+        _cfg: &mut (),
+        state: &mut Option<T>,
+        input: (&T,),
+        _ctx: &mut Ctx<'_>,
+    ) -> Result<Tick<(T, T)>> {
+        let value = input.0.clone();
+        let out = match state.take() {
+            Some(prev) => Tick::Value((prev, value.clone())),
+            None => Tick::Quiet,
+        };
+        *state = Some(value);
+        Ok(out)
+    }
+}
+
 /// Negates each value (`!value`).
 ///
 /// Was fluent-only sugar over [`Map`] (`map(|v| !v.clone())`) until it became
--- a/crates/wingfoil/tests/catalog.rs
+++ b/crates/wingfoil/tests/catalog.rs
@@ -127,6 +127,39 @@ fn difference_emits_deltas_after_first() {
     assert_eq!(vec![1, 1, 1], r.value(&acc));
 }
 
+/// Pairwise emits pairs of consecutive values, the first tick is quiet.
+#[test]
+fn pairwise_emits_pairs() {
+    let g = GraphBuilder::new();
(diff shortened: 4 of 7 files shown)


0-jake-0 (commented):
Reviewed. The op itself is right — `Option<T>` state, `Activation::NONE`, `Tick::Quiet` on the first value, no `Sub` bound — and it hits every acceptance checkbox in #800, including the non-arithmetic `String` case that is the whole reason the op exists. The `.map(|(p, n)| *p + *n)` in the `op_completeness` block is a nice touch: it keeps `surface_u64` typed at `u64` so `pairwise` gets merged into the same parity assertion as everything else.

## Verified locally (at `8a89224`)

- `cargo test -p wingfoil --test catalog --test op_completeness --test op_fluent_shapes --test compiled_stateful_ops` — all pass, including `pairwise_agrees_across_engines` and `generic_receiver_supports_tuple_output`
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p wingfoil-derive` — pass (the macro crate is not reached by `-p wingfoil`, and this PR edits it)

CI agrees: `Test (wingfoil)`, `Lint (fmt & clippy)` and `Python Interop` are all green.

## 1. `src/signal.rs` is missing its one line

`#[op(build = pairwise, fluent)]` emits `__wf_signal_pairwise!`, and the `Signal` facade has to opt in or it silently falls behind the catalog — that is step 4b of `/new-op`, and it is exactly what the last new op did (`skip`, 0daeb5f, touches `signal.rs` in the same commit). Nothing else in the file changes:

```rust
    __wf_signal_difference!(T);
    __wf_signal_pairwise!(T);
```

I applied that locally and `cargo build -p wingfoil` is clean, so it really is the one line.

This is the finding I would call blocking — the facade drifted 15 methods behind once already, and it drifts one op at a time.

## 2. `docs/planning/port-plan.md` row (step 8)

`skip` updated the "Trivial state/closure" row at `port-plan.md:494`; `pairwise` belongs in the same list next to `difference`, ideally with the note that it is new surface with no legacy twin and has three-engine parity in `tests/compiled_stateful_ops.rs` (the phrasing `skip` used).

## 3. Python binding — do it, or say why not in the PR body

Step 7 treats the binding as part of adding an op, not an extra. Both of this op's nearest mirrors have one: `difference` (`graph.rs:449`, `python.rs:205`, plus the README table row) and `skip` (`graph.rs:431`).

`#[pyop]` cannot help here, since the erased edge type is `PyElement` and `Out` would be `(PyElement, PyElement)`. But a hand-written `PyStream` method is short and idiomatic for this codebase — keep `Option<PyElement>` state and emit a Python tuple as the element, the way `split`/`item` already treat values as indexable objects:

```rust
// pairs compose straight back into the existing `.split()`
Ok(Tick::Value(PyElement::new(PyTuple::new(py, [prev, cur])?.unbind().into())))
```

which makes `stream.pairwise().split()` work on the Python side and matches what `split_decomposes_tuples` already assumes about tuple-shaped elements. If you would rather leave it for a follow-up that is a defensible call, but step 7 asks for the reason to be stated in the PR description.

## 4. Nit — the catalog tests assert values but not tick times

`CLAUDE.md` and #800's acceptance box both ask for values *and* tick times. `pairwise_agrees_across_engines` does carry `.with_time()`, so the timing is pinned somewhere; adding it to `pairwise_emits_pairs` would make `tests/catalog.rs` self-sufficient and state explicitly that the first pair lands at t=20, not t=10 — which is the quiet-first-tick contract.

## 5. Unrelated churn in `compiled_stateful_ops.rs`

The `throttle_values` body was reflowed onto five lines. rustfmt does not format `nitro!` bodies, so that is a hand edit unrelated to `pairwise` — worth reverting to keep the diff to the op.

## 6. `is_generic_tuple` is narrower than it reads (not blocking)

The new predicate matches only tuples whose elements are *all* bare generic idents, so `(T, usize)`, `(T, Vec<U>)` and `((T, T), T)` still miss the bound. That is enough for `pairwise` and I would not expand it here, but a recursive "does this type mention a method-or-self generic" helper would subsume both arms and stop the next tuple-shaped op re-opening the file.

Related: the coverage you added is in `tests/op_fluent_shapes.rs`, which exercises the in-crate expansion. The out-of-crate path (`tests/trybuild/pass/out_of_crate_op.rs`, the only place `::wingfoil::` paths are provably exercised) has no tuple-`Out` case. Not required by this PR, worth knowing the gap exists.

---

1 is the one I would want before merge; 2 and 3 are the skill's bookkeeping and are cheap; 4–6 are optional. Nice, tight implementation otherwise.

---

0-jake-0:
Merged as `112ff15` — thanks, this was a clean piece of work. The op is exactly the shape #800 described, and the `String` test is the case that justifies the op existing at all.

Rather than hold it open for the bookkeeping, I've taken the three follow-ups on:

- **#881** — the `__wf_signal_pairwise!(T)` line on the `Signal` facade (with a forwarder test over `String`, mirroring `skip`'s), and the port-plan inventory row.
- **#882** — the Python binding, filed as its own `good first issue`. It can't go through `#[pyop]` (the erased edge is `PyElement`, and `Out` is a tuple), so it needs a hand-written `PyStream` method emitting a `PyTuple` — enough work to deserve its own PR, and yours if you want it.

The two nits from the review — the unrelated `throttle_values` reflow, and `tests/catalog.rs` asserting values without tick times — I left alone; neither is worth churn now that the timing is pinned in `compiled_stateful_ops.rs`.

---

Issue #459, opened 2026-07-11
API polish tracker: missing combinators, run builder, `etcd_pub` signature


0-jake-0:
**Tracker.** The combinator gaps below are now one issue each — self-contained ops with the shape, the reference op to copy, and acceptance criteria spelled out. Each links `/new-op` as the recipe.

> **Rescoped 2026-08-14.** Split into per-op issues so they can be picked up independently. The `latency_report_if(false)` item is fixed and struck. Two non-combinator items remain here.
>
> **Updated 2026-08-20.** Merged items ticked; two new combinators added (#889, #890). Every op issue now carries its Python binding as an acceptance item rather than a follow-up.
>
> **Updated 2026-08-23** (9.0 release-readiness audit, verified against `main` @ `f794b07`). `take_while` and `enumerate` have merged since the last pass and are ticked. The `etcd_pub` item was **already fixed** and is struck — the body had it as &#34;still current&#34;, which made the 9.0 API surface look less settled than it is. Nothing on this tracker is an API break: every remaining item is additive, so none of it blocks 9.0.
>
> **Updated 2026-08-23** (tracker audit, verified against `main` @ `a6269d6`). The last three &#34;PR in flight&#34; items have all merged — `skip_while` (#902), `step_by` (#894) and `filter_map` (#873). **The combinator section of this tracker is now complete**: every box below is ticked, and the only combinators still outstanding are #802 and #803, which are tracked on their own issues because each needs a decision rather than a copy-the-neighbouring-op edit. That leaves **one live item on this tracker** — the `RunMode`/`RunFor` builder.

### Missing combinators — one issue each

Presence re-verified against `crates/wingfoil/src/fluent.rs` on `main` @ `a6269d6`.

Good first issues — each is ~15–20 lines plus tests, with an existing op to copy. **All merged; nothing left to pick up here:**

- [x] #797 — `skip(n)` (mirror of `Limit`) — merged in #846
- [x] #798 — `skip_while(pred)` (latching, unlike `filter`) — merged in #902
- [x] #799 — `take_while(pred)` (predicate form of `limit`) — merged in #886
- [x] #800 — `pairwise()` (`difference` without the `Sub` bound) — merged in #879
- [x] #801 — `try_map_filter(f)` (the missing corner of the map/filter × fallible 2×2) — merged in #880
- [x] #889 — `enumerate()` (index-with-value; `pairwise`&#39;s tuple shape, `count`&#39;s state) — merged in #892
[...]

PR #880, opened 2026-08-18
Add try_filter_map combinator

### LINKED ISSUE #801: Add `try_map_filter(f)` combinator
> **Resolved.** Implemented in #880 and renamed in #884. **This issue originally specified the name `try_filter_map`, which was wrong** — the catalog's convention is `try_` plus the *exact* base name (`map` → `try_map`, `join` → `try_join`, `join_passive` → `try_join_passive`), the base here is `map_filter`, and `Signal` already has a `filter_map` with a different, `Option`-shaped signature. The text below has been corrected to `try_map_filter` so it isn't reused as a template for the wrong name. See #884 for the reasoning.

Part of #459. **Good first issue** — the fallible twin of an existing op, ~15 lines plus tests.

**What:** `try_map_filter(f)` where `f: Fn(&A) -> Result<(B, bool)>` — maps and filters in one pass, propagating an error to abort the run.
**Why:** the catalog has `map_filter` (infallible, maps + filters) and `try_map` (fallible, maps). The fallible map-and-filter is the missing corner of that 2×2, and it is the natural shape for parsing/decoding a stream where some inputs are skippable and others are genuinely malformed.

|  | infallible | fallible |
|---|---|---|
| map | `map` | `try_map` |
| map + filter | `map_filter` | **missing — this issue** |

### Naming

`try_` + the exact base name. The base is `map_filter`, so this is **`try_map_filter`**, not `try_filter_map`. The two are easy to transpose and the wrong one collides with `Signal::filter_map`, which takes `Fn(&T) -> Option<B>` and is *not* this op's infallible counterpart.

### Semantics

Follow the two existing ops exactly:
- The closure returns `Result<(B, bool)>` — the `bool` is the emit decision, as in `map_filter`.
- `Ok((v, true))` → `Tick::Value(v)`; `Ok((_, false))` → `Tick::Quiet`.
- `Err(e)` → propagate with `?`, which aborts the run with context, as in `try_map`.

The distinction to make explicit in the rustdoc: **`false` means "no value this tick", `Err` means "the run is broken"** — they are not interchangeable.

### Shape

| | |
|---|---|
| `Cfg` | `F: Fn(&A) -> Result<(B, bool)> + 'static` |
| `State` | `()` |
| `In<'a>` | `(&'a A,)` |
| `Out` | `B` |
| Attribute | `#[op(build = try_map_filter, fluent)]` |

**Reference ops:** `MapFilter` (`crates/wingfoil/src/ops.rs:161`) for the `(value, emit)` convention, `TryMap` (`ops.rs:138`) for the error propagation. This op is the two of them composed.

> ⚠️ **Closure configs are `Fn`, not `FnMut`** — a documented contract in the op recipe.

### Sketch

```rust
fn cycle(cfg: &mut F, _state: &mut (), input: (&A,), _ctx: &mut Ctx<'_>) -> Result<Tick<B>> {
    let (value, emit) = cfg(input.0)?;
    Ok(if emit { Tick::Value(value) } else { Tick::Quiet })
}
```

### Acceptance

- [x] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = try_map_filter, fluent)]`
- [x] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_try_map_filter!(A, B)`)
- [x] Exercised in a `nitro!` block in `crates/wingfoil/tests/op_completeness.rs` (two-sided fluent + compiled guard, asserts `interpreted() == compiled()`)
- [x] Test asserting exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`
- [x] Test that an `Err` **aborts the run** and that the context reaches the caller — the behaviour that distinguishes this from `map_filter`
- [x] Rustdoc drawing the `false` vs `Err` distinction
- [x] A test for the `Signal` facade forward in `tests/signal.rs` — **this was missing from the original list**; the forward shipped untested in #880 and was added in #884. Any op with a `Signal` facade line needs one, or deleting that line leaves the whole suite green.

### Getting started

Run `/new-op try_map_filter` (`.claude/commands/new-op.md`). Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md). Branch from `main`, PR base `main`.

zaydmulani09:
## Summary

Fixes #801 (part of #459).

`map_filter` (infallible, maps+filters) and `try_map` (fallible, maps) already exist — `try_filter_map` is the missing fallible map+filter corner. `f: Fn(&A) -> Result<(B, bool)>`: `Ok((v, true))` ticks `v`, `Ok((_, false))` means no value this tick and the run continues, `Err(e)` aborts the run with `e` as context. The rustdoc calls out that `false` and `Err` are not interchangeable, per the issue.

Composed from `MapFilter` and `TryMap` in `ops.rs`, following their exact shape.

## Changes

- `crates/wingfoil/src/ops.rs` — `TryFilterMap<A, B, F>` op, `#[op(build = try_filter_map, fluent)]`
- `crates/wingfoil/src/fluent.rs` — fluent method on `StreamOps`
- `crates/wingfoil/src/signal.rs` — `Signal` facade forward
- `crates/wingfoil/tests/op_completeness.rs` — exercised in the `surface_fallible` `nitro!` block (`interpreted() == compiled()`)
- `crates/wingfoil/tests/catalog.rs` — values + tick times parity test
- `crates/wingfoil/tests/fallibility.rs` — `Err` aborts the run with node context, mirroring the existing `try_map` test

No Python binding in this PR — issue #801's acceptance criteria list only the Rust op, fluent method, and the three test obligations, with no Python binding mentioned. Happy to add it as a follow-up if wanted.

## Test plan

- [x] `cargo fmt --all` — no diff
- [x] `cargo test -p wingfoil` — full suite passes
- [x] `cargo test -p wingfoil-derive` — passes
- [x] `cargo clippy -p wingfoil --all-targets -- -D warnings` — clean (two pre-existing warnings unrelated to this change, present on `main` too)
- [ ] `cargo lint-all` / `--all-features` — not run; this environment lacks `protoc`/`libclang` needed for the proto/iceoryx2 adapters, unrelated to this change

### DIFF
--- a/crates/wingfoil/src/fluent.rs
+++ b/crates/wingfoil/src/fluent.rs
@@ -882,6 +882,16 @@ pub trait StreamOps<T>: Sized {
         B: Clone + Default + 'static,
         F: Fn(&T) -> (B, bool) + 'static;
 
+    /// Fallible [`map_filter`](StreamOps::map_filter): `f` returns
+    /// `Result<(value, emit?)>`, mapping and filtering in one pass. `Ok((v,
+    /// true))` ticks `v`; `Ok((_, false))` means no value this tick and the
+    /// run continues; `Err(e)` aborts the run with `e` as context — `false`
+    /// and `Err` are not interchangeable.
+    fn try_filter_map<B, F>(&self, f: F) -> Stream<B>
+    where
+        B: Clone + Default + 'static,
+        F: Fn(&T) -> Result<(B, bool)> + 'static;
+
     /// Pair each value with the current engine time: `(time, value)`.
     fn with_time(&self) -> Stream<(NanoTime, T)>
     where
@@ -1284,6 +1294,8 @@ impl<T: 'static> StreamOps<T> for Stream<T> {
     __wf_fluent_try_map!(T);
 
     __wf_fluent_map_filter!(T);
+
+    __wf_fluent_try_filter_map!(T);
 
     fn with_time(&self) -> Stream<(NanoTime, T)>
     where
--- a/crates/wingfoil/src/ops.rs
+++ b/crates/wingfoil/src/ops.rs
@@ -190,6 +190,39 @@ where
     }
 }
 
+/// The fallible [`MapFilter`]: the closure returns `Result<(value, emit?)>`,
+/// mapping and filtering in one pass, with an `Err` aborting the run — the
+/// `try_` counterpart to `map_filter`, composed the same way [`TryMap`]
+/// composes with [`Map`]. `Fn`, like [`Map`].
+///
+/// `false` and `Err` are not interchangeable: `Ok((_, false))` means "no
+/// value this tick" and the run continues; `Err(e)` means the run is broken
+/// and aborts with `e` as context.
+pub struct TryFilterMap<A, B, F>(PhantomData<(A, B, F)>);
+
+#[op(build = try_filter_map, fluent)]
+impl<A, B, F> Op for TryFilterMap<A, B, F>
+where
+    A: 'static,
+    B: Clone + 'static,
+    F: Fn(&A) -> Result<(B, bool)> + 'static,
+{
+    type Cfg = F;
+    type State = ();
+    type In<'a> = (&'a A,);
+    type Out = B;
+    const ACTIVATION: Activation = Activation::NONE;
+
+    fn cycle(cfg: &mut F, _state: &mut (), input: (&A,), _ctx: &mut Ctx<'_>) -> Result<Tick<B>> {
+        let (value, emit) = cfg(input.0)?;
+        Ok(if emit {
+            Tick::Value(value)
+        } else {
+            Tick::Quiet
+        })
+    }
+}
+
 /// Suppresses consecutive duplicate values: emits the first value, then only
 /// when it changes. State is `Option<T>` (not the default-initialised output)
 /// so a genuine first value equal to `T::default()` still ticks.
--- a/crates/wingfoil/src/signal.rs
+++ b/crates/wingfoil/src/signal.rs
@@ -139,6 +139,7 @@ impl<T: 'static> Signal<T> {
     __wf_signal_map!(T);
     __wf_signal_try_map!(T);
     __wf_signal_map_filter!(T);
+    __wf_signal_try_filter_map!(T);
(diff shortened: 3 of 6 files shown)


0-jake-0:
Merged — thanks @zaydmulani09, this was a pleasure to review.

A few things you did that go beyond what #801 asked for and are worth calling out:

- **The tests are stronger than the acceptance criteria.** `try_filter_map_err_aborts_run` doesn't just check that an `Err` aborts — it asserts `vec[10]` at the end, which pins that the *filtered-out* value never landed while the *erroring* one killed the run. That is precisely the `false` vs `Err` distinction the issue asked to document, turned into something a future refactor can't quietly break. A weaker version of that test would have passed too.
- **The `op_completeness.rs` rewiring was the right call.** Moving `try_join`'s left edge from `tried` to `filtered` means the join now sees a `Quiet` left input on alternate cycles instead of a value every time, so the interpreted-vs-compiled guard covers activation behaviour it wasn't covering before. Easy to have appended a parallel chain instead and tested less.
- **Reading the `Fn`-not-`FnMut` contract and the touch-point table correctly**, first time, including the `Signal` facade line that is easy to miss.

One heads-up so it isn't a surprise: I've opened #884 to rename this to `try_map_filter`. **That is a correction to the issue, not to your PR.** #801 specified the name and the `#[op(build = try_filter_map, fluent)]` attribute verbatim and you implemented exactly that. The problem is on our side: the catalog's convention is `try_` plus the *exact* base name (`map` → `try_map`, `join` → `try_join`), the base here is `map_filter`, and `Signal` already has a `filter_map` with a different (`Option`-shaped) signature — so the name landed next to a method it reads as the twin of and isn't. I'm editing #801 so the next person doesn't inherit it.

#884 also adds the `Signal` facade test that neither the issue nor the review checklist asked for; the forward had no coverage, which is on our recipe rather than on you.

On the Python binding — you read #801 right, it wasn't in scope. If you'd like it, `/bind-adapter` isn't the one you want (that's for adapters); the `#[pyop]` / `pyop_fn!` steps are in `/new-op`. Happy to open an issue for it if you'd rather pick it up as a separate piece of work.

---

PR #886, opened 2026-08-20
feat(core): add take_while combinator

### LINKED ISSUE #799: Add `take_while(pred)` combinator

**What:** `take_while(pred)` emits values while `pred` returns true; at the first value where it returns false, the op goes permanently quiet.
**Why:** the counted form (`limit(n)` = `take`) exists; the predicate form does not.

### Semantics

Once the predicate fails, the op latches off and never emits again — even if later values would satisfy it. The failing value itself is **not** emitted.

| value | `pred` | latched off? | emits |
|---|---|---|---|
| 1 | true | no | `Value(1)` |
| 2 | true | no | `Value(2)` |
| 9 | false | → yes | `Quiet` |
| 1 | true | yes | `Quiet` |

**It goes quiet; it does not end the run.** That matches `Limit`, which suppresses rather than terminating — keep the two consistent.

### Shape

| | |
|---|---|
| `Cfg` | `F: Fn(&T) -> bool + 'static` |
| `State` | `bool` — whether it has stopped |
| `In<'a>` | `(&'a T,)` |
| `Out` | `T` |
| Attribute | `#[op(build = take_while, fluent)]` |

**Reference ops:** `Limit` (`crates/wingfoil/src/ops.rs:389`) for the shape and the quiet-not-terminate precedent, `MapFilter` (`ops.rs:161`) for the closure-in-`Cfg` pattern.

> ⚠️ **Closure configs are `Fn`, not `FnMut`** — a documented contract in the op recipe.

### Sketch

```rust
fn cycle(cfg: &mut F, state: &mut bool, input: (&T,), _ctx: &mut Ctx<'_>) -> Result<Tick<T>> {
    if *state || !cfg(input.0) {
        *state = true;
        Ok(Tick::Quiet)
    } else {
    }
}
```

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = take_while, fluent)]`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_take_while!(T)`)
- [ ] Test asserting exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`, **including that a later satisfying value stays suppressed**
- [ ] Rustdoc stating that it suppresses rather than terminating, cross-referencing `limit`

### Getting started

Run `/new-op take_while` (`.claude/commands/new-op.md`). Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md). Branch from `main`, PR base `main`.

maxi-maxima:
## What this changes

Adds a latching `take_while(predicate)` combinator across the Rust Op, fluent Stream, Signal, Nitro, and Python surfaces. The first rejected value and every later value are quiet; the run itself is not terminated.

The Python binding follows Python truthiness and preserves predicate and `__bool__` exceptions as run errors.

## Why

Closes #799.

This fills a maintainer-requested stateful gating primitive while keeping exact interpreted/compiled/nested parity and documenting the Python-facing surface.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all` — attempted locally, but this Windows host has no MSVC `link.exe`/Windows SDK, so dependency build scripts fail before project code is compiled
- [ ] `cargo test -p wingfoil --all-features` — blocked at the same local host linker boundary
- [x] New behaviour is covered by tests asserting **values and tick times**

Additional passing local checks:

- `cargo metadata --no-deps --format-version 1`
- `git diff --check`

The upstream Linux rerun at `e992703` passed Wingfoil tests, Python Interop, fmt+Clippy, Cargo/pnpm audits, and dependency review. The regression suite includes interpreted/compiled/nested execution, fluent and Signal surfaces, post-rejection latching, Python non-`bool` truthiness, and predicate/truthiness exceptions.

## Notes for the reviewer

`TakeWhile` uses a per-run `bool` state initialized to accepting. The Python callable is wired directly instead of through the infallible Rust `Op` config so Python exceptions can propagate without panicking or being swallowed.

The follow-up commit binds the Nitro test's tail stream as required by the macro and matches the preserved Python exception type in the `__bool__` regression.


0-jake-0:
Reviewed — we're happy with this as it stands. It hits every `/new-op` touch point (op + fluent + Signal surfaces, three-engine parity, op-completeness, the Python binding with truthiness/exception coverage, and the docs), and the latching-quiet semantics are the right call and clearly documented. CI is fully green on `e992703`.

Since it's still marked draft: is there anything you'd like to add or change before we merge? If not, just mark it ready for review and we'll take it from there.

---

maxi-maxima:
Marked ready. Current \main\ removed the builder-less \Signal\ façade in #887, so I merged \main\ and resolved the two delete/modify conflicts by keeping that removal; the take-while implementation, fluent API, Nitro coverage, Python binding, and docs remain intact. Fresh CI on \6227cc1\ is fully green.

0-jake-0 (approved):
Follow-up verified — approved. The merge of `main` resolved the delete/modify conflicts the right way: the Signal facade stays removed per #887, with no `signal.rs` resurrection, and the take_while op, fluent method, Nitro coverage and Python binding all survived intact (confirmed on the head, including `take_while` coverage in `catalog.rs`, `compiled_stateful_ops.rs` and `op_completeness.rs`).

Worth calling out: inverting the state to `stopped` so that `bool::default() = false` *is* the accepting state — with the rustdoc line saying so — sidesteps the exact latch-direction pitfall our friction audit flagged for this op, without needing a `start` hook.

CI (`test.rust`, Python Interop) is green on 6227cc1. This is ready to merge; only sequencing against #892/#894/#896 (same insertion points in `ops.rs`/`fluent.rs`/`op_completeness.rs`/`port-plan.md`) remains, and that's on us.

---

PR #892, opened 2026-08-21
feat(core): add enumerate combinator

### LINKED ISSUE #889: Add `enumerate()` combinator (index-with-value)
Part of #459. **Good first issue** — a self-contained op, ~20 lines plus tests, and a Python binding with an exact reference to copy.

**What:** `enumerate()` pairs each value with its zero-based index — `(0, a)`, `(1, b)`, `(2, c)`, …
**Why:** the index is the one piece of per-stream context no other combinator recovers. Today you reach for `count()` and a `join`, which costs a node and gets the base wrong — `count` ticks *with* the value, so it reads `1` on the first one, not `0`.

### Semantics

Ticks `Tick::Value((i, v))` on **every** input value, `i` starting at `0` and incrementing per *value seen*, not per engine cycle. Never quiet — the first value emits `(0, v)`. (Contrast `pairwise`, which must stay quiet until a previous value exists; `enumerate` has no such warm-up.)

### Shape

| | |
|---|---|
| `Cfg` | `()` |
| `State` | `u64` — the next index to emit |
| `In<'a>` | `(&'a T,)` |
| `Out` | `(u64, T)` |
| Attribute | `#[op(build = enumerate, fluent)]` |

**Reference ops — copy their structure:**
- **`Pairwise`, `crates/wingfoil/src/ops.rs:339`** — the tuple `Out`, its fluent method and its `Signal` facade line. This is the closest twin.
- **`Count`, `crates/wingfoil/src/ops.rs:3027`** — the counter `State`.

### Sketch

```rust
fn cycle(
    _cfg: &mut (),
    state: &mut u64,
    input: (&T,),
    _ctx: &mut Ctx<'_>,
) -> Result<Tick<(u64, T)>> {
    let index = *state;
    *state += 1;
    Ok(Tick::Value((index, input.0.clone())))
}
```

`State` is `u64` and `Default`-seeded to `0`, so a re-run starts from `0` again — which is the correct behaviour and worth a test.

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = enumerate, fluent)]`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_enumerate!(T)`), next to `pairwise` (`fluent.rs:1171`)
- [ ] `__wf_signal_enumerate!(T)` on the `Signal` facade in `src/signal.rs`, next to `__wf_signal_pairwise!` (`signal.rs:166`)
- [ ] Exercised in a `nitro!` block in `crates/wingfoil/tests/op_completeness.rs` — the two-sided guard proving the op has *both* a fluent method and a compiled forwarder, asserting `interpreted() == compiled()`
- [ ] Test that a second `run()` restarts the index at `0`
- [ ] **Python binding** — see below
- [ ] Rustdoc on both the op struct and the fluent method

### Python binding — part of this ticket, not a follow-up

`Out` is a tuple, so this **cannot** be a one-line forward like `skip` (`graph.rs:432`), and it cannot go through `#[pyop]` — the erased edge is `PyElement`. It needs a hand-written `PyStream` method emitting a `PyTuple`, exactly as `pairwise` does:

- [ ] **`PyStream::enumerate` in `crates/wingfoil-python/src/graph.rs`, modelled on `PyStream::pairwise` (`graph.rs:458`)** — `register_op1` + `Python::attach` + `PyTuple::new`. Yours is simpler: the first tuple element is the index rather than a held previous value, and there is no quiet branch. Emit the index as a Python int.
- [ ] Facade line on `Stream` in `crates/wingfoil-python/src/python.rs`, next to `pairwise` (`python.rs:210`)
- [ ] Test in `crates/wingfoil-python/tests/test_interop.py` — `enumerate().split()` should give an index stream and a value stream, mirroring `test_pairwise_splits_previous_and_current_strings` (`test_interop.py:481`)

The tuple output composing with `.split()` is the reason to prefer this shape over returning two streams.

### Getting started

The recipe is `/new-op enumerate` (`.claude/commands/new-op.md`) — op shape, `#[op]` flags, the fluent method, `nitro!`/compiled coverage, tests, and §7 for the bindings. Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md).

maxi-maxima:
## What

Add an `enumerate()` combinator that emits each stream value as `(zero_based_index, value)`.

- add the core stateful op and fluent Rust API
- expose the same behavior in Python as a real tuple, compatible with `split()`
- cover exact values and tick times, interpreted/compiled/nested parity, op completeness, and rerun reset
- document the Rust/Python surface and update the port plan

## Why

Consumers currently cannot attach a zero-based per-stream index with a single combinator. Reconstructing it from `count()` requires an additional stream/join and correcting its one-based counter semantics. This implements the maintainer-filed #889 directly while preserving each input tick time.

## Verification

Passed locally:

- `git diff --check`
- `cargo metadata --locked --format-version 1 --no-deps`
- independent static review of the `/new-op` surfaces and coverage

The complete upstream matrix is green on `73c2c75`:

- Wingfoil tests
- Python Interop
- fmt + Clippy
- Cargo and pnpm audits
- dependency review

Local Rust execution remains unavailable because this Windows host has no MSVC `link.exe`; `maturin` is also not installed locally. The upstream Linux jobs provide the authoritative compile/test signal.

## Notes

The builder-less `Signal` facade was removed from current `main` in #887, so this targets the current `Stream` and Python surfaces only.

Closes #889


0-jake-0 (approved):
Reviewed against `/new-op` and #889 — approved. The op matches the ticket's sketch exactly (never quiet, index per value seen, `(0, v)` on the first value), and every acceptance item is covered: values-and-tick-times catalog test, three-engine parity, the `surface_u64` completeness wiring, and the rerun-restarts-at-zero test the ticket specifically asked for.

The Python binding is the right shape — hand-written over `register_op1` with a `PyTuple` and the index as a Python int, modelled on `pairwise` as directed, and the `enumerate().split()` test mirrors the pairwise one so the tuple/`split` composition is pinned on both sides of the boundary. Nice catch adding `pairwise` to api.rst's transform list while you were there — it had been missed when that binding landed.

Skipping the Signal-facade item is correct post-#887. CI (`test.rust`, Python Interop) is green on the head (73c2c75) and the PR is cleanly mergeable.

Same sequencing note as its siblings: #886, #894 and #896 touch the same insertion points in `ops.rs`/`fluent.rs`/`op_completeness.rs`/`port-plan.md`, so merge order will force trivial rebases — we'll sequence that.

---

0-jake-0:
@maxi-maxima this PR went un-mergeable when #886 (`take_while`) landed on `main` ahead of it. Could you merge `origin/main` into your branch? It's first in the merge queue — #894 and #896 will rebase after it lands.

I ran the merge locally to save you the archaeology: all the code files auto-merge (the `surface_u64` graph body correctly ends up exercising both new ops), and only three prose sites conflict:

1. **`crates/wingfoil/tests/op_completeness.rs`** — the `surface_u64` doc comment. Keep `main`'s phrasing and name both ops:
   ```
   // The single-input `u64` surface: `count`, `map`, `map_filter`,
   // `distinct`, `drop_small_change`, `difference`, `pairwise`, `enumerate`,
   // `limit`, `skip`, `take_while`,
   ```
2. **`docs/python-interop.md`** — the "Built-in combinator surface" table row. Take `main`'s version (it has "plus wingfoil-only additions" and `skip`/`take_while`) and insert `pairwise`/`enumerate` after `difference`.
3. **`docs/planning/port-plan.md`** — the "Trivial state/closure" row. Union of both sides: node list gains `enumerate` (after `pairwise`) while keeping `take_while` (after `skip`); in the notes keep `main`'s `take_while` sentence and this PR's `enumerate` sentence side by side.

With exactly that resolution, `cargo test -p wingfoil --test op_completeness --test compiled_stateful_ops --test catalog --test rerun` (62 tests) and `cargo test -p wingfoil-python` (107 tests) both pass on the merged tree here. Once your CI is green we'll merge straight away.

---

maxi-maxima:
Addressed in 4dab931. I merged current main and resolved the three conflicts exactly as suggested: the surface_u64 note names both numerate and 	ake_while, the Python surface row keeps the current wording plus both additions, and the port-plan row unions both nodes and their semantics. Local fmt, locked metadata, and targeted diff checks passed; the refreshed Wingfoil tests, Python Interop, fmt+Clippy, and security matrix are all green.

Issue #830, opened 2026-08-14
Add #[must_use] to transform/source combinators (fluent traits + derive-generated methods)


0-jake-0:
From the fable-review-2 codebase review (finding adversarially verified against the code).

Dropping the result of a pure combinator — `s.map(f);` with no binding — is not a no-op in this engine: `Stream::wire` registers the node with the shared `Builder` (`crates/wingfoil/src/fluent.rs:739-752`) and there is no reachability pruning anywhere in `src/`, so the dropped node **stays wired and cycles every tick for the whole run**, producing an unread value. That's a silent logic bug (the user thinks they transformed the stream) plus a permanent per-cycle cost — and nothing warns.

Meanwhile `latency.rs:211-226` and ~50 adapter-trait methods already carry `#[must_use]`; `fluent.rs` has zero (grep-verified), and the derive-generated `Signal`/fluent methods are bare (`crates/wingfoil-derive/src/lib.rs:2848-2869`). MSRV 1.88 is far past the point where `#[must_use]` on trait method declarations fires for callers.

**Scope — transforms and sources only, not sinks.** Sink-shaped methods are idiomatically called as discarding statements in this very tree (`examples/adapters/fix/main.rs:73`, `:81` call `.logged(...)` bare; kdb examples likewise) and their side effects happen regardless of the returned handle. Annotate the value-returning `StreamOps`/`SourceOps`/`StatisticsOps` declarations where a discarded result means a useless live node; exclude `for_each`/`for_each_mut`/`print`/`logged` and other sink returns; add the attribute to the derive's `expand_fluent`/`expand_signal` quotes so generated methods match.

PR #894, opened 2026-08-22
feat(core): add step_by combinator

### LINKED ISSUE #890: Add `step_by(n)` combinator (emit every nth value)
Part of #459. **Good first issue** — this is #797 (`skip`) with `%` in place of `<`, plus a one-line Python binding.

**What:** `step_by(n)` emits the first value, then every nth value after it — indices `0, n, 2n, 3n, …`
**Why:** decimation is the standard way to thin a fast stream for a slow consumer (a chart, a log, a checkpoint) when you want a fixed *count* ratio. `throttle` is the time-domain answer to the same problem and is already present; this is the count-domain one, and there is currently no way to express it without a `fold` and a `filter`.

Not to be confused with `sample`, which resamples off another stream's ticks.

### Semantics

Counts **values seen**, not engine cycles. Value at index `i` emits iff `i % n == 0`, so:

- `step_by(1)` is the identity — every value passes
- `step_by(3)` over `a b c d e f g` emits `a`, `d`, `g`

Ticks `Tick::Value(v)` when the value passes and `Tick::Quiet` otherwise.

### Shape

| | |
|---|---|
| `Cfg` | `usize` — the step |
| `State` | `usize` — values seen so far |
| `In<'a>` | `(&'a T,)` |
| `Out` | `T` |
| Attribute | `#[op(build = step_by, fluent)]` |

**Reference op — copy its structure:** **`Skip`, `crates/wingfoil/src/ops.rs:515`.** Same `Cfg`/`State`/`In`/`Out`, same activation; only the branch differs. Read `Limit` (`ops.rs:487`) alongside it — `Skip` and `Limit` are each other's mirror and this is the third member of that family.

### Sketch

```rust
fn cycle(
    cfg: &mut usize,
    state: &mut usize,
    input: (&T,),
    _ctx: &mut Ctx<'_>,
) -> Result<Tick<T>> {
    let index = *state;
    *state += 1;
    if index % *cfg == 0 {
    } else {
        Ok(Tick::Quiet)
    }
}
```

### The one judgement call: `step_by(0)`

`index % 0` panics, and **panicking on bad user config is against the repo rule** (`CLAUDE.md`, and §3 of `/new-op`: *"Never panic for bad user config"*). Pick one and say which in the rustdoc:

1. `anyhow::bail!` in `cycle` with a clear message — aborts the run with context. Simplest, and consistent with how the skill says to handle config that can't be validated at wiring.
2. Validate at wiring time in the fluent method.

Option 1 is the lower-friction choice here and matches existing practice; option 2 is better if you can do it without changing the fluent signature to return `Result`. Either way, **add a test for it** — this is the interesting case in the whole ticket.

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = step_by, fluent)]`, placed next to `Skip`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_step_by!(T)`), next to `skip` (`fluent.rs:1114`)
- [ ] `__wf_signal_step_by!(T)` on the `Signal` facade in `src/signal.rs`, next to `__wf_signal_skip!` (`signal.rs:154`)
- [ ] Exercised in a `nitro!` block in `crates/wingfoil/tests/op_completeness.rs` — the two-sided guard proving the op has *both* a fluent method and a compiled forwarder, asserting `interpreted() == compiled()`
- [ ] Test for `step_by(1)` (identity) and for the `step_by(0)` behaviour you chose
- [ ] **Python binding** — see below
- [ ] Rustdoc on both the op struct and the fluent method, stating the `n = 0` behaviour

### Python binding — part of this ticket, not a follow-up

`Out = T`, so this works over the erased `PyElement` edge with **no hand-written op**. It is a one-line forward, exactly like `skip`:

- [ ] `PyStream::step_by` in `crates/wingfoil-python/src/graph.rs`, modelled on `PyStream::skip` (`graph.rs:432` — `self.wrap(self.stream.skip(n))`)
- [ ] Facade line on `Stream` in `crates/wingfoil-python/src/python.rs`, next to `skip`
- [ ] Rust-side unit test mirroring `skip_suppresses_the_first_n` (`graph.rs:1430`)
- [ ] pytest in `crates/wingfoil-python/tests/test_interop.py`

### Getting started

The recipe is `/new-op step_by` (`.claude/commands/new-op.md`) — op shape, `#[op]` flags, the fluent method, `nitro!`/compiled coverage, tests, and §7 for the bindings. Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md).

maxi-maxima:
## What this changes

Adds a count-domain `step_by(n)` combinator that emits input indices 0, n, 2n, and so on across Rust, Nitro, and Python. It validates zero once during graph start and returns a contextual run error rather than panicking.

## Why

Fast streams currently need a fold plus filter to decimate by a fixed value ratio. This adds the maintainer-specified counterpart to time-domain `throttle`.

Closes #890.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo lint` and `cargo lint-all` — upstream fmt + Clippy job passed on `37099c1`
- [x] `cargo test -p wingfoil --all-features` — upstream Wingfoil test job passed on `37099c1`
- [x] New behaviour is covered by tests asserting values and tick times

The complete upstream matrix is green on `37099c1`:

- Wingfoil tests
- Python Interop
- fmt + Clippy
- Cargo and pnpm audits
- dependency review

Additional local checks passed:

- `cargo metadata --locked --format-version 1 --no-deps`
- `git diff --check`
- clean-context review against the repository `/new-op` checklist

The new tests cover step sizes 1 and 3, rerun reset, Python parity, and the zero-step error with an upstream that never ticks. A dedicated Nitro regression observes that start-hook error in interpreted, compiled, and nested execution.

Local Rust execution was unavailable because this Windows host lacks MSVC `link.exe`; the upstream Linux jobs provide the authoritative compile, test, lint, and Python-extension signal.

## Notes for the reviewer

Current `main` removed the builder-less `Signal` facade in #887, so this follows the current Stream and Python surfaces rather than adding the issue's now-retired Signal forwarding line.


0-jake-0 (approved):
Reviewed against `/new-op` and #890 — approved. This hits every acceptance item, and the one judgement call the ticket flagged (`step_by(0)`) got the strongest available answer: validation in `start`, so a zero step is rejected even when the upstream never ticks, with that error observed on all three engines. The `PhantomData` anchor in `StepByState` is exactly the ThrottleState pattern `/new-op` §3 prescribes for an op that gains a `start` hook, and the `checked_add` on the counter keeps the no-panic rule airtight.

Coverage is complete: values-and-tick-times catalog test, `step_by(1)` identity, three-engine parity, the zero-step error across interpreted/compiled/nested, rerun reset, the `surface_u64` completeness wiring, and the one-line Python forward with Rust-side and pytest coverage. Docs (README table, api.rst, port-plan) all updated. Skipping the issue's Signal-facade item is correct — #887 removed it from `main`.

CI (`test.rust`, Python Interop) is green on the head (37099c1) and the PR is cleanly mergeable.

One sequencing note, on us not you: #886, #892 and #896 insert at the same spots in `ops.rs`/`fluent.rs`/`op_completeness.rs`/`port-plan.md`, so whichever merges first leaves the others needing a trivial rebase.

---

0-jake-0:
@maxi-maxima you're next in the queue: #886 (`take_while`) and #892 (`enumerate`) have landed on `main`, so this PR now conflicts. Could you merge `origin/main` into your branch? #896 rebases after you.

I ran the merge here first — ten files conflict, but every conflict is a both-sides-added-adjacent one, so the resolution is "keep both" throughout, with these specifics:

- **`ops.rs`** — keep both complete op blocks after `Skip`: the whole `StepBy` block (`StepBy` + `StepByState` + its `Default` impl + the `#[op]` impl with `start`), then the whole `TakeWhile` block from `main`. Watch that neither impl's tail gets truncated — the conflict markers split mid-impl.
- **`fluent.rs`** — keep both trait declarations and both macro lines, `__wf_fluent_step_by!(T);` then `__wf_fluent_take_while!(T);`.
- **`tests/op_completeness.rs`** — comment list union: ``…`pairwise`, `enumerate`, `limit`, `skip`, `step_by`, `take_while`,``. In `surface_u64`, chain the two new ops instead of keeping two `let seen` bindings:
  ```rust
  let stepped = skipped.step_by(2);
  let taken = stepped.take_while(|i| *i < u64::MAX);
  let seen = taken.inspect(|_| ());
  ```
  (keep your decimation comment above it).
- **`tests/compiled_stateful_ops.rs`** — keep both test sections whole: your `step_by` section (including `step_by_zero`) then `main`'s `take_while` section; union the header comment.
- **`graph.rs` / `python.rs` / `test_interop.py`** — keep both methods / both test groups (`step_by` then `take_while`), minding closing braces at the seams.
- **`README.md`** — both table rows. **`api.rst`** — one Gate list: ``…``limit``, ``skip``, ``step_by``, ``take_while``, ``throttle``…``.
- **`port-plan.md`** — a single "Trivial state/closure" row: node list gains `step_by` between `skip` and `take_while` (keeping `enumerate` after `pairwise`), and the notes keep all three sentences — your `step_by` one plus `main`'s `take_while` and `enumerate` ones.

With exactly that resolution, `cargo fmt` is clean and `cargo test -p wingfoil --test op_completeness --test compiled_stateful_ops --test catalog --test rerun` (68 tests) plus `cargo test -p wingfoil-python` (85 tests) all pass here. Once your CI is green we'll merge immediately.

---

0-jake-0 (commented):
Re-reviewed after the `main` merge (37c21c6 — the approval above was against 37099c1). The merge resolved cleanly: `take_while` and `enumerate` are both present alongside `step_by` in `ops.rs`, `fluent.rs`, `op_completeness.rs` and the `port-plan.md` row, and `surface_u64` still reads coherently with `step_by(2)` sitting between `skip(1)` and `take_while`. CI is green on this head, and the PR is mergeable. The approval stands.

Two small things I missed the first time, both documentation-shaped:

**1. `docs/python-interop.md` wasn't updated.** You updated `crates/wingfoil-python/README.md` and `docs/api.rst`, but the built-in combinator surface row in `docs/python-interop.md` (line ~328) also enumerates the bound combinators — `…/limit/skip/take_while/throttle/…` — and both #886 and #892 updated it. `step_by` belongs in that list.

**2. `#[must_use]` is missing on the `StreamOps::step_by` declaration.** Per `/new-op` §4c and #830, a dropped combinator handle leaves the node wired and cycling every tick for the whole run, so transforms carry

```rust
#[must_use = "a dropped stream stays wired and cycles every tick, producing an unread value"]
```

on the hand-written declaration in `fluent.rs`. `limit` and `skip` have it; `take_while` (#886) and `enumerate` (#892) both merged without it, so there was no neighbouring precedent to copy — that's on us, and we'll fix those two separately. Worth adding here while the PR is open.

Neither blocks; if you'd rather not touch the branch again we can fold both in at merge — say which you prefer. Everything I checked in the original review still holds: the `start`-time zero-step rejection observed on all three engines, the `StepByState<T>` `PhantomData` anchor, `checked_add`, the rerun reset, and the Python forward with both Rust-side and pytest coverage.

---

PR #896, opened 2026-08-22
Add skip_while combinator

### LINKED ISSUE #798: Add `skip_while(pred)` combinator

**What:** `skip_while(pred)` suppresses values while `pred` returns true; once it returns false, everything from then on passes through — including values that would satisfy the predicate again.
**Why:** the latching behaviour is what distinguishes it from `filter`, and it is one of the expected members of the skip/take family (only `limit` = `take` exists today).

### Semantics

`State` is a latch, **not** a re-evaluated condition:

| value | `pred` | latched? | emits |
|---|---|---|---|
| 1 | true | no | `Quiet` |
| 2 | true | no | `Quiet` |
| 5 | false | → yes | `Value(5)` |
| 1 | true | yes | `Value(1)` ← passes, because the latch is set |

That last row is the whole point, and the test must cover it. `filter` is the non-latching alternative and already exists.

### Shape

| | |
|---|---|
| `Cfg` | `F: Fn(&T) -> bool + 'static` |
| `State` | `bool` — whether skipping has finished |
| `In<'a>` | `(&'a T,)` |
| `Out` | `T` |
| Attribute | `#[op(build = skip_while, fluent)]` |

**Reference ops:** `Limit` (`crates/wingfoil/src/ops.rs:389`) for the state-machine shape, `MapFilter` (`ops.rs:161`) for how a closure lives in `Cfg`.

> ⚠️ **Closure configs are `Fn`, not `FnMut`** — a documented contract in the op recipe. Bound it as `Fn(&T) -> bool + 'static`.

### Sketch

```rust
fn cycle(cfg: &mut F, state: &mut bool, input: (&T,), _ctx: &mut Ctx<'_>) -> Result<Tick<T>> {
    if !*state && cfg(input.0) {
        Ok(Tick::Quiet)
    } else {
        *state = true;
    }
}
```

### Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = skip_while, fluent)]`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs` (body via `__wf_fluent_skip_while!(T)`)
- [ ] Exercised in a `nitro!` block in `crates/wingfoil/tests/op_completeness.rs` (the two-sided fluent + compiled guard, which also asserts `interpreted() == compiled()`)
- [ ] Test asserting exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`, **including the post-latch re-satisfying case** from the table above
- [ ] Rustdoc making the latching semantics explicit, and cross-referencing `filter` as the non-latching alternative

### Getting started

Run `/new-op skip_while` (`.claude/commands/new-op.md`) for the full recipe. Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md). Branch from `main`, PR base `main`.

pttydou:
## What this changes

Adds a latching `skip_while(predicate)` combinator to Rust `StreamOps` and Python `Stream`. The predicate stops being called after its first false result, and the op works across interpreted, compiled, and nested execution.

## Why

Closes #798. Wingfoil has non-latching filters and count-based `skip`, but no predicate-based combinator for suppressing only the initial prefix.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all`
- [ ] `cargo test -p wingfoil --all-features`

`cargo lint`, `cargo test -p wingfoil`, `cargo test -p wingfoil-python`, and Python `pytest crates/wingfoil-python/tests/test_interop.py -q` passed locally.

`cargo lint-all` and the all-features test remain CI/Linux checks: on this Windows host, the all-feature build reaches `iceoryx2-pal-posix` and stops because `libclang.dll` is unavailable.

## Notes for the reviewer

The Python binding is hand-written because Python predicates may raise; it mirrors the Rust latch and propagates callback and truthiness errors with context.


0-jake-0 (approved):
Reviewed against the `/new-op` recipe and #798 — this is a clean, complete implementation and we're happy to take it.

What stands out positively:

- The op matches the issue's spec exactly, including the post-latch re-satisfying case, and the latch direction is right: `bool::default() = false` is the still-skipping state, so no `start` hook is needed.
- The predicate-call-count assertion (`3 * 3` across the three engines) pins the "predicate is never called after the latch opens" contract, which the issue didn't even ask for.
- The Python binding correctly goes through `register_op1` rather than the infallible Rust op (the closure-`Cfg` gotcha in `/new-op` §7), mirrors the `filter_value` house pattern, and covers truthiness, predicate exceptions, and `__bool__` exceptions.

Since your Windows host couldn't run the lint/test legs, I ran them here on your head commit (c45a6a3): `cargo lint` (clippy `-D warnings`, workspace) is clean, and the `compiled_stateful_ops` + `op_completeness` suites pass, including the new three-engine tests. CI (`test.rust`, Python Interop) is also green on the head.

Two non-blocking notes:

1. Optional: a rerun-reset test (`tests/rerun.rs`) asserting the latch re-closes on a second `run()` would pin the one piece of state this op has — #892/#894 added the equivalent for theirs. Happy to take that as a follow-up or fold it in at merge; not holding the PR for it.
2. Sequencing: #886 (`take_while`), #894 (`step_by`) and #892 (`enumerate`) insert at the same spots in `ops.rs`/`fluent.rs`/`op_completeness.rs`/`port-plan.md`, so whichever merges first will leave the others needing a trivial rebase. That's on us to sequence, nothing for you to do now.

---

0-jake-0 (commented):
Second pass over this head (c45a6a3), on top of the approval above. The op itself still looks right to me — the latch inversion (`false` = still skipping, so no `start` hook), the short-circuit that stops calling the predicate once open, and the `register_op1` binding are all as reviewed, and the binding is now structurally identical to the `take_while` one that landed in #886, which is the outcome we want.

Three things have changed or turned up since, none of them about the op's semantics.

**1. It no longer merges cleanly — please rebase onto `main`.** #886 (`take_while`), #892 (`enumerate`) and #894 (`step_by`) have all landed at the same insertion points since you branched off 9c50fe8. A trial merge conflicts in six files:

```
crates/wingfoil-python/src/graph.rs
crates/wingfoil-python/src/python.rs
crates/wingfoil/src/fluent.rs
crates/wingfoil/src/ops.rs
crates/wingfoil/tests/compiled_stateful_ops.rs
docs/planning/port-plan.md
```

All of them are adjacent-insertion conflicts — same region, different new item — so the resolution is mechanical. The one that needs a moment's thought is the `port-plan.md` row: `main`'s version now describes `skip`, `step_by` and `take_while` together, so fold `skip_while` into that sentence rather than restoring your version of it.

**2. Four documentation touch-points are missing, all of which #886 updated for the mirror op.** The Rust and Python surfaces are complete, but a Python user goes looking in these:

- `crates/wingfoil-python/README.md` — the combinator table (it lists `.skip(n)` and `.take_while(pred)`; `.skip_while(pred)` belongs between them).
- `crates/wingfoil-python/docs/api.rst` — the *Gate* line, `…, skip, step_by, take_while, throttle, …`.
- `docs/python-interop.md` — the built-in combinator surface row (line ~328), which enumerates the bound combinators.
- `crates/wingfoil/tests/catalog.rs` — `take_while_latches_quiet_after_the_first_rejection` sits there; the values-and-tick-times case for `skip_while` currently only exists in `compiled_stateful_ops.rs` and `op_completeness.rs`. Not a coverage gap in substance, but the catalog is where a reader looks for the single-op contract.

`git show <the #886 merge> --stat` is a good template for the full set — it touched 12 files to this PR's 8, and the difference is exactly the list above.

**3. `#[must_use]` is missing on the `StreamOps::skip_while` declaration.** `/new-op` §4c and #830: a dropped combinator handle is not a no-op — the node stays wired and cycles every tick producing an unread value — so transforms carry

```rust
#[must_use = "a dropped stream stays wired and cycles every tick, producing an unread value"]
```

on the hand-written declaration in `fluent.rs` (never inside the `__wf_fluent_*` expansion, where it is inert). `limit` and `skip` — the ops `skip_while` sits between — both have it. In fairness this is *our* inconsistency as much as yours: `take_while` and `enumerate` both merged without it too, so there is no local precedent in the file to copy from. We'll fix those two separately; please add it here so this one lands right.

None of this is a change to the op. Rebase + the four doc touch-points + the attribute, and it's ready to merge.

---

0-jake-0:
Heads-up on how this one is landing, @pttydou.

The approval above stands — the op, the latch direction, the tests and the Python binding are all as reviewed. What blocks the merge button is purely mechanical: this branched from 9c50fe8, and `take_while` (#886), `enumerate` (#892) and `step_by` (#894) have since landed at the same insertion points, so it now conflicts in six files. Because the head branch is on your fork, we can't push the rebase to it from here (GitHub returns "user doesn't have permission to update head repository" on the update-branch call).

So rather than leave a reviewed, approved contribution stranded, I've carried your commit onto #902 — **your authorship preserved**, conflicts resolved, no change to the op's semantics, tests or binding. It closes #798 the same way. The two resolutions that took a judgement call are written up in that PR's description: the `port-plan.md` row is woven into `main`'s combined sentence rather than restored from here, and `ops.rs` writes `SkipWhile` and `TakeWhile` out in full because the merge had collapsed them onto each other (identical `Cfg`/`State`/`In`/`Out` shape).

I'll close this PR once #902 merges. Thank you for the contribution — the predicate-call-count assertion in particular is a nice touch that pins a contract the issue didn't even ask for, and it's what makes the "never called after the latch opens" claim more than a comment.

The `#[must_use]` point from my earlier review isn't in #902 — it applies equally to `take_while` and `enumerate`, which are already on `main` without it, so all four are being fixed together in a follow-up rather than singling yours out. Nothing needed from you.

---

0-jake-0:
Landed on `main` as 945618d — your commit, rebased, via #902. Closing this one as superseded rather than rejected: the work merged, the credit is yours (`Co-authored-by` on the squash, and the cherry-pick kept you as the commit author), and #798 is closed by it.

Thanks again, @pttydou.

---

PR #900, opened 2026-08-23
docs(examples): demonstrate pooled channel backpressure

### LINKED ISSUE #893: Add examples/core/pooled_channel — worked example of the loan-based pooled channel
## What

`SourceOps::pooled_channel` / `pooled_channel_with` (the loan-based, zero-steady-state-allocation channel in `src/pool.rs`) has no example under `crates/wingfoil/examples/`. Its only worked usage today lives in tests and benches:

- `tests/pooled_channel.rs` — recycling proof, the flat-out-producer wedge regression, timestamped replay
- `tests/steady_state_allocs.rs` — zero-payload-allocation pin
- `benches/pooled_channel.rs` — throughput/latency vs the plain channel

A user reaching for the loan API (`loan()` → write in place → `send`, `Pooled<T>` handles graph-side, bounded pool as backpressure) currently has to reverse-engineer it from tests.

## Proposed shape

`examples/core/pooled_channel/` (it is an engine concept, not an adapter — no external services, plain `cargo run`):

- A producer thread loaning recycled buffers for a heap-owning payload (e.g. a small `Vec`-backed book built with `pooled_channel_with(capacity, || Book::with_depth(..))` to show pre-sizing), writing in place, `send`-ing.
- Graph side: `Stream<Burst<Pooled<Book>>>` through a routing op or two (showing handle clones are refcount bumps, and the wrap-don't-clone enrichment pattern from the `pool` module docs).
- Emit as the run progresses (`print()` / `for_each`), not `accumulate` — per the working conventions, and doubly so here since parked handles occupy the loan budget.
- Demonstrate backpressure: capacity small enough that `loan()` visibly blocks/`try_loan()` fails while the graph holds buffers.

## Checklist (per the example rules in CLAUDE.md, enforced by `scripts/check-example-docs.sh`)

- [ ] `main.rs` + `README.md` in its own directory
- [ ] Explicit target with `path` under `# Examples` in `Cargo.toml` (`autoexamples = false`)
- [ ] Linked from `examples/core/README.md` (house style: `## Sentence-case title`, prose, snippet, output)
- [ ] Sample output in the README pasted from a real run
- [ ] README points at the `pool` module docs for the loan-budget rules

maxi-maxima:
## What this changes

Adds a runnable `pooled_channel` core example that demonstrates a pre-sized payload pool, wrap-not-clone routing, bounded loan backpressure, and buffer recycling with preserved `Vec` capacity. It also registers and indexes the example and includes its deterministic output.

## Why

Closes #893. The loan lifecycle currently has module docs and tests, but no end-to-end worked example showing how producer-side exhaustion relates to graph-held handles or how a heap-owning payload is reused without deep clones.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all` — `cargo lint` passes; `lint-all` reaches the optional `iceoryx2-pal-posix` bindgen build and stops because this Windows host has no `libclang.dll`
- [ ] `cargo test -p wingfoil --all-features` — same host-only `libclang.dll` prerequisite blocks the all-features build before the changed target is compiled
- [ ] New behaviour is covered by a test asserting **values and tick times** — not applicable to this documentation example; its phase gate makes the teaching output deterministic

Additional checks:

- `cargo check -p wingfoil --example pooled_channel`
- `cargo clippy -p wingfoil --example pooled_channel -- -D warnings`
- `cargo run -p wingfoil --example pooled_channel` (output matched the README in 3/3 runs)
- `cargo test -p wingfoil --test pooled_channel` (7 passed)
- `cargo test -p wingfoil` (full default-feature suite passed)
- `scripts/check-example-docs.sh` (`47` targets documented and indexed)

## Notes for the reviewer

`Book` deliberately does not implement `Clone`. A small atomic phase gate is used only to make the backpressure and pointer-reuse steps reproducible while the graph runs in realtime; the example does not retain pooled handles in `accumulate()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ that move that line. All four are open: see [Get Involved](#get-involved).
 
 ## Examples
 
-44 runnable examples, each in its own directory with a README covering what it
+45 runnable examples, each in its own directory with a README covering what it
 teaches, the wiring, and its expected output. Full index:
 [`examples/README.md`](crates/wingfoil/examples/README.md).
 
@@ -262,6 +262,7 @@ No services, no feature flags — these run with a plain `cargo run`.
 | [`async`](crates/wingfoil/examples/core/async/) | Tokio async/await at the graph's edges, with the core graph staying synchronous. |
 | [`async_source`](crates/wingfoil/examples/core/async_source/) | An async quote feed driving the graph through an `external` source. |
 | [`threading`](crates/wingfoil/examples/core/threading/) | Distribute graph execution across worker threads, with no locks on the execution path. |
+| [`pooled_channel`](crates/wingfoil/examples/core/pooled_channel/) | Recycle pre-sized payload buffers through a loan-based channel with a bounded backpressure budget. |
 | [`spawn`](crates/wingfoil/examples/core/spawn/) | Offload slow work off the graph thread with `spawn` / `spawn_map`. |
 | [`tracing`](crates/wingfoil/examples/core/tracing/) | Observability: the `logged` debug tap and the engine's own spans. |
 | [`introspect`](crates/wingfoil/examples/core/introspect/) | Read back the graph you wired — text, Mermaid, DOT, JSON or GML. |
--- a/crates/wingfoil/Cargo.toml
+++ b/crates/wingfoil/Cargo.toml
@@ -703,6 +703,12 @@ path = "examples/core/tracing/main.rs"
 name = "threading"
 path = "examples/core/threading/main.rs"
 
+# Loan-based ingress with recycled, pre-sized payload buffers and a bounded
+# pool as backpressure. Uses only default features.
+[[example]]
+name = "pooled_channel"
+path = "examples/core/pooled_channel/main.rs"
+
 # The same thread offload as `threading`, but through the `spawn` / `spawn_map`
 # combinators that wrap the channel plumbing into one call each.
 [[example]]
--- a/crates/wingfoil/examples/README.md
+++ b/crates/wingfoil/examples/README.md
@@ -40,7 +40,7 @@ Then pick a direction:
 
 **Tiers (`nitro!`)**: [`dual_mode`](core/dual_mode/)
 
-**Concurrency**: [`threading`](core/threading/) · [`spawn`](core/spawn/) · [`async`](core/async/) · [`async_source`](core/async_source/)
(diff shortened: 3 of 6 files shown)


0-jake-0 (approved):
Reviewed against #893 and the example rules in `CLAUDE.md` — approved. Every checklist item on the issue is met, and the two rules that are easiest to get wrong here are both handled properly.

I built and ran the example on this head (0d867db) rather than taking the README's word for it: **the pinned output reproduces byte-for-byte, 3/3 runs**, and `scripts/check-example-docs.sh` passes (47 targets documented and indexed). CI is green.

What stands out:

- **The realtime/determinism collision is resolved the right way.** `CLAUDE.md` says a README that pins output must run historical, *except* where the example exists to show realtime behaviour — which this one does, since the loan budget is a live producer/consumer property. The atomic phase gate makes the ordering an invariant rather than a hope, and it fails loudly (a 5 s deadline → `bail!`) instead of printing a different interleaving. That is the one shape that satisfies both rules at once.
- **`Book` deliberately has no `Clone`.** That turns "routing clones the handle, not the payload" from a claim in prose into a compile-time proof, which is stronger than the `Rc::strong_count` assertion the issue's "handle clones are refcount bumps" bullet would have got you.
- **The example asserts its own teaching points** (`available() == 1`, `try_loan().is_none()`, pointer identity of the recycled buffer, surviving `Vec` capacity) via `ensure!` + `send_error`, so a regression in the pool breaks the example loudly instead of quietly printing something plausible. Emitting from `for_each` rather than parking handles in `accumulate()` is exactly right, and doubly so here — the README says why.

One non-blocking suggestion:

- **`collapse()` is the house spelling for "one item out of the arrival burst".** The hand-rolled

  ```rust
  .map(|burst: &Burst<Pooled<Book>>| {
      burst.last().expect("invariant: a channel tick carries a non-empty burst").clone()
  })
  ```

  is `ops::Collapse` written out: `Collapse::cycle` is `into_iter().last()`, iterates the burst **by reference** (only the emitted item is cloned, same as here), and returns `Tick::Quiet` on an empty burst instead of needing the `expect`. Swapping it for `.collapse::<Pooled<Book>>()` is one line, drops the invariant comment, and points readers at the documented lossy-collapse semantics — which is a thing worth teaching in an example about burst ingress. I tried it locally: it compiles and the output is unchanged. Happy either way; the current form is correct, just not the vocabulary the rest of the tree uses.

Nothing else to change. Thanks for the phase gate — it is the detail that makes this example maintainable.

---

Issue #437, opened 2026-07-11
web: historical streaming is lossy — a fast graph drops frames (no back-pressure / pacing)


0-jake-0:
## Background

#429 makes `web_pub` stream a `RunMode::HistoricalFrom` replay to the browser.
That works for a **slow / compute-bound** historical run, but the delivery path
is **lossy** and never back-pressures the graph, so a **fast** historical graph
(a real backtest that runs at CPU speed) silently drops frames — the browser
sees a corrupted, incomplete replay.

## Why it's lossy today

- `web_pub` pushes into an **unbounded** `consume_async` channel; a tokio task
  drains it into a per-topic `tokio::sync::broadcast` (`PUBLISH_BROADCAST_CAPACITY
  = 1024`). `broadcast` **cannot** block its sender — a lagging receiver gets
  `RecvError::Lagged` and skips ahead.
- `forward_broadcast` uses `out.try_send(...)` into the per-connection outbound
  mpsc and **drops on `Full`**.

This is by design for **real time** — "a frozen browser tab can never
back-pressure the graph" (`adapters/web/CLAUDE.md`). But a historical replay has
no live clock to fall behind, so dropping frames there just corrupts the output.

## Evidence

Probe during #429: a historical graph emitting 5000 values to a single loopback
subscriber delivered **4586 / 5000** frames (with gaps). A real browser over a
network would drop far more.

## Options

1. **Wall-clock replay throttle (opt-in).** `web_pub` paces historical frames to
   wall-clock time (optionally a speed multiplier) so a human can watch. Simple,
   self-contained, and does **not** reverse the graph-never-blocks rule. Cost:
   buffers in memory if the graph badly outruns the throttle, and it's still
   lossy past the buffer.
2. **Lossless back-pressure (historical only).** Bounded, blocking outbound path
   so the graph paces to the (slowest) client and the browser sees every frame in
   order. Most correct for a faithful replay, but reverses "never back-pressure
   the graph", couples graph progress to one client, and needs a policy for
   "zero clients connected" (must drain freely, or the run hangs — same class as
   the `web_sub` hang fixed in #429) and "multiple clients" (pace to slowest).
3. **Status quo (lossy).** Keep as-is; document that faithful historical
   streaming requires a compute-bound graph. Already documented in #429.

A clean framing is a `Delivery { Auto, Lossy, Lossless }` (or `Throttled`) knob on
the server builder, defaulting to `Auto` = lossy in real time, paced/lossless in
[...]

PR #908, opened 2026-08-23
Add Delivery policy for historical replay losslessness (#437)

0-jake-0:
## What this changes

Introduces a `Delivery` policy to the web adapter that resolves per run mode: historical replays now deliver frames losslessly (pacing the graph to the slowest subscriber), while real-time runs remain lossy (never back-pressuring the graph). The default `Delivery::Auto` picks the sensible behaviour for each mode; `Delivery::Lossy` and `Delivery::Lossless` allow explicit opt-out or opt-in.

## Why

Historical replays running at CPU speed were corrupting output by dropping frames to a slow client — a probe on #437 delivered only 4586 of 5000 frames. Real-time graphs cannot afford to wait on a slow client (a frozen browser tab must never stall a live trading system), but a replay has no live clock to fall behind, so dropping frames does not "keep up" — it just corrupts what the browser draws. The fix paces the publisher to the slowest subscribed client in historical mode, while keeping real-time lossy and non-blocking.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo test -p wingfoil --all-features`, `cargo test --doc -p wingfoil --all-features`, `cargo test -p wingfoil-derive`
- [x] `pytest tests/test_web.py` against a `maturin develop --features web` build
- [x] New behaviour is covered by tests asserting values and completion:
  - `historical_streaming_is_lossless_for_a_fast_graph` — all 12000 frames reach a stalled client in order
  - `historical_lossless_does_not_hang_without_subscribers` — unwatched replays run at full speed
  - `historical_lossless_paces_to_the_slowest_of_several_subscribers` — multi-client fairness
  - `lossless_withdraws_a_subscriber_that_stops_accepting_frames` — a half-open peer cannot park the graph
  - `a_withdrawn_subscriber_gets_its_connection_closed` — a withdrawn client sees the close (and leaves `subscriber_count`), so it can reconnect
  - `concurrent_runs_on_one_server_resolve_delivery_independently` — one server, two overlapping runs, opposite modes
  - `lossy_opt_out_never_lets_a_wedged_client_stall_the_graph` — `Delivery::Lossy` opt-out works
  - `realtime_auto_stays_lossy_under_a_wedged_client` — real-time remains non-blocking
  - `delivery_defaults_to_auto_and_round_trips` / `the_lossless_stall_timeout_has_a_default_and_round_trips` — API surface and defaults
  - `historical_noop_server_ignores_delivery` — no-op servers ignore the policy
  - Python: delivery selection, the stall-timeout round trip, and its rejection of a non-positive value

**One known local failure, not from this branch:** `tests/trybuild/must_use_combinators.rs` mismatches under rustc 1.94.1, which emits a fully-qualified path where the captured `.stderr` has a bare one. It reproduces on `main` at 8fc36fd and is green on CI, so the `.stderr` was deliberately left alone.

## Notes for the reviewer

**Implementation strategy.** The lossless path awaits a send into each subscribed connection's bounded outbound queue, so the consumer task stalls behind the slowest client. That stall reaches the graph through a semaphore (`LOSSLESS_INFLIGHT_INSTANTS = 64`): the graph thread takes a permit per instant and the consumer returns it once that instant is delivered, so the graph runs a bounded distance ahead and then waits. Lossy is byte-for-byte the previous path — unbounded sink channel, 1024-slot `broadcast`, `try_send` dropping on a full outbound queue.

**Key design decisions:**
- `Delivery::Auto` resolves at graph `start`, not at construction, because the run mode does not exist when the server is built.
- The resolved flag is stored **per publish** — an `Arc<AtomicBool>` created alongside that publish's semaphore, not on `WebServerInner`. A server-wide cell would let two graphs sharing one `WebServer` in opposite run modes overwrite each other's policy, and would let a flip land between the graph thread's read and the consumer's read for the same instant, leaking or inventing a permit.
- The lossless path snapshots its subscribers **once per instant**, not once per frame: it is the path that sets a replay's throughput ceiling, and the set can only usefully change between instants.
- `lossless_stall_timeout` (default 30 s, settable on the builder and from Python) bounds the wait. Pacing to the slowest subscriber has to mean the slowest *live* one — a half-open peer never closes its socket, so an unbounded await would park the graph until TCP keepalive, and since the end-of-run `Complete` rides the same path, `run()` itself could hang after the last cycle. It cannot catch a merely slow client: the wait is for one slot in a 1024-deep queue, so anything draining at all resets it.
- **Withdrawing a stalled subscriber closes its whole connection.** Registry-only withdrawal would leave the socket open and silent — no data, and no `Complete` either, since `finally` snapshots a registry the client has left — and `@wingfoil/client` keys both `onComplete` and its stop-reconnecting logic on that marker. The clients that would strand are the *recoverable* ones (a laptop that suspends, trips the bound, wakes); a dead peer is indifferent. Teardown on that path aborts the writer rather than awaiting its drain, which would park on the same wedged socket. Closing the connection rather than the subscription is right because every topic on a connection shares one outbound queue. It also drops the broadcast receivers, so `subscriber_count` stops advertising an unreachable client.
- A connection registers on the lossless registry *before* subscribing to the broadcast, because it is the broadcast's `receiver_count` that `subscriber_count` reports — so the count moving means both paths can reach that client.
- With zero subscribers the lossless path never waits, so an unwatched replay runs at full speed and does not hang.
- Real-time behaviour is unchanged; the wire format is untouched, so `wingfoil-wasm` and `@wingfoil/client` need no change.

**Known limitation, documented not fixed.** A client that is withdrawn and reconnects rejoins a replay already in progress and has missed whatever went out while it was away. Losslessness is a property of a subscription, not of a topic; resume-from-cursor would need a wire change. Noted in `js/README.md`.

**Test probe sizing.** The loss tests replay 12000 frames of ~2 KiB each to a client that stalls 750 ms before reading. Both numbers are load-bearing: the lossy path buffers a frame three times over (1024-slot broadcast, 1024-slot outbound queue, and the kernel's auto-tuned loopback socket buffer), so at a few dozen bytes a frame a stalled client still receives every one and the probe silently stops probing. At this size a pre-`Delivery` run measured 11036 of 12000 delivered. Each new test was checked against the bug it pins — raise `lossless_stall_timeout` to 25 s and the stall test fails at 27 s; make the policy flag process-wide and the concurrency test fails on "the concurrent replay must stay lossless"; drop the `close.notify_one()` and the withdrawal test fails on "must see its connection close". The concurrency test can only *miss*, never flake red: if the live run finishes before the replay resolves, there is no overlap.

**Docs.** `cutover-plan.md` and `cutover-runbook.md` both listed #437 as a survivor still to be confirmed against wingfoil's own adapter; it did carry over, and both now record it as discharged. New deviation-register row **D30**, plus the adapter `CLAUDE.md`, module docs, the web example and the `js/` client README. The `/new-adapter` and `/bind-adapter` skills gained the two rules this surfaced: resolve a run-mode-dependent policy at graph `start`, and prove a loss probe against the path it replaces.

https://claude.ai/code/session_01NUKUKsrtrpcy9SeFzs9Mfs


crates/wingfoil/src/adapters/web/server.rs
```diff
@@ -53,14 +130,99 @@ pub(crate) struct WebServerInner {
...
+                .lossless_subs
+                .lock()
+                .expect("lossless_subs lock poisoned");
+            match guard.get(topic) {
+                Some(subs) => subs.iter().map(|s| (s.id, s.tx.clone())).collect(),
+                None => return,
+            }
+        };
+        for (id, tx) in snapshot {
+            if tx.send(bytes.clone()).await.is_err() {
```
0-jake-0: This `send` awaits with no timeout, and the server never originates a WebSocket ping (`Message::Ping(_) | Message::Pong(_) => continue` at line 590 only responds to the peer), so a half-open peer — client machine dies, network partition, laptop lid closed — is undetectable. The socket stays open, the queue stays full, and the graph parks until TCP keepalive fires, which is ~2 h by default.

Since `Delivery::Auto` resolves every historical `web_pub` to lossless, this is now the default behaviour for replays. It also reaches teardown: `finally` calls `handle.block_on(complete_inner.deliver_lossless(..))` on the graph thread (write.rs:344), behind `flush(&())?`, so `run()` — and `SinkShared::flush`/`Drop` after it — can hang after the last cycle. These graphs always terminated before.

Pacing to the slowest subscriber is the right trade for a replay; the gap is that "slowest" currently includes "gone and never coming back". A `send_timeout` that withdraws the sub via `remove_lossless_sub`, or a server-side ping/pong liveness check, would bound it without giving up losslessness for a client that is merely slow.

---

crates/wingfoil/src/adapters/web/server.rs
```diff
@@ -53,14 +130,99 @@ pub(crate) struct WebServerInner {
...
     }
 
+    /// Turn the configured [`Delivery`] into a concrete one for this run.
+    ///
+    /// Called from `web_pub`'s graph `start` hook, which is the earliest point
+    /// the run mode exists — the server is built before the graph runs, and a
+    /// `WebServer` can serve more than one run, so this is a per-run decision
+    /// rather than a construction-time one. It runs before the first cycle, so
+    /// no frame is ever published against an unresolved policy.
+    pub(crate) fn resolve_delivery(&self, run_mode: RunMode) {
```
0-jake-0: `resolved` is one atomic for the whole server, but `resolve_delivery`'s own doc comment (line 148) justifies resolving at `start` on the grounds that "a `WebServer` can serve more than one run". Those two facts don't sit together if the runs overlap:

- A historical graph starting alongside a live real-time one flips the live graph into pacing on browser sockets — exactly what the `Delivery` doc at line 50-52 says must never happen to a trading system.
- The reverse order silently reverts an in-flight replay to lossy, losing the property this PR adds.
- It also opens the permit imbalance the comment at write.rs:287-291 warns about. `paced_sink` and the consumer closure each read `is_lossless()` independently; if a flip lands between those two reads for the same instant, either the consumer adds a permit the graph never took (semaphore inflated for the rest of the server's life) or the graph takes one nothing returns.

Either scope the resolved policy per publish — it's already per-`publish_frames`, so an `Arc<AtomicU8>` created alongside the semaphore and captured by both closures would make the two reads provably agree — or narrow the doc claim to sequential runs and say so where `Delivery` is documented.

---

crates/wingfoil/src/adapters/web/server.rs
```diff
@@ -418,6 +612,8 @@ async fn handle_socket(socket: WebSocket, inner: Arc<WebServerInner>) {
                         }
                         let sender = inner.get_or_create_pub_topic(&topic);
                         let rx = sender.subscribe();
+                        let id = inner.register_lossless_sub(&topic, outbound_tx.clone());
```
0-jake-0: Minor ordering nit: `sender.subscribe()` on the line above is what `subscriber_count` reads (`receiver_count()`), and it lands before `register_lossless_sub` populates the registry the lossless publisher actually consults. So the doc at line 113-115 — "a connection registers here and on `lossless_subs` in the same step, so the count means the same thing either way" — is very slightly overstated: there is a window where the count says 1 and the lossless registry is empty.

The window is a handful of instructions with no `await`, and `wait_for_subscribers` polls on 5 ms and then builds a whole graph before publishing, so I don't think the new tests are meaningfully racy. But swapping these two lines makes the doc claim true unconditionally, for free.

---

crates/wingfoil/src/adapters/web/write.rs
```diff
@@ -211,41 +260,141 @@ where
...
+{
+    for (time, payload) in items {
+        let env = Envelope {
+            topic: topic.to_string(),
+            time_ns: u64::from(time),
+            payload: codec.encode(&payload)?,
+        };
+        let bytes = Bytes::from(codec.encode(&env)?);
+        if lossless {
+            inner.deliver_lossless(topic, bytes).await;
```
0-jake-0: This is inside the `for (time, payload) in items` loop, so `deliver_lossless` re-locks `lossless_subs` and clones the whole `Vec<(u64, Sender)>` once per **frame**, not once per instant — on the path that now sets the replay's throughput ceiling.

The subscriber set can only change between instants from the publisher's point of view (a mid-instant unsubscribe just means a `send` error, which is already handled), so hoisting the snapshot to the top of `deliver_instant` and passing it down would cut a mutex acquisition and an allocation per frame. Worth doing while the lossless path is the one gating replay speed.

---

0-jake-0 (commented):
Reviewed the delivery path in `server.rs` and `write.rs` against `d718338`. CI is green.

## What holds up

The semaphore bookkeeping balances on every path I traced. The empty-burst skip in `paced_sink` mirrors `consume_async_bursts`' own skip, so no permit is taken for an instant the consumer never sees. Both error paths hand the permit back exactly once — graph-side on `handed_over.is_err()`, consumer-side unconditionally under `if lossless`. `flush()`-before-`Complete` in `finally` keeps the end-of-stream marker behind the last data frame in the same outbound queue, so ordering survives. `deliver_lossless` snapshots the sender `Vec` out of the mutex before any `await`, so no lock crosses a suspension point, and the `try_acquire` fast path keeps the graph thread off the runtime unless the pipeline is genuinely full.

The diagnosis in the description is right too: a replay has no live clock to fall behind, so dropping frames buys nothing and corrupts what the browser draws.

## Findings

Four, all at the edges rather than in the core accounting — inline:

1. **`server.rs:219`** — the lossless `send` has no timeout and the server never originates a ping, so a half-open peer parks the graph until TCP keepalive (~2 h), including through teardown via `finally`'s `block_on`. This is the one I'd want addressed before lossless becomes the historical default.
2. **`server.rs:152`** — one `resolved` atomic per server, against a doc comment that advertises a server outliving several runs. Overlapping runs cross-contaminate, and can inflate or leak a permit.
3. **`server.rs:615`** — `subscribe()` lands before `register_lossless_sub`, so `subscriber_count` briefly overstates lossless readiness. Cheap to make the doc claim true.
4. **`write.rs:388`** — per-frame lock and `Vec` clone on the path that now gates replay throughput; hoist it to per-instant.

Nothing here undermines what the PR fixes. #1 is the substantive one; #3 and #4 are small.

---

crates/wingfoil/src/adapters/web/server.rs
0-jake-0: Agreed, and fixed in 87bc4b1 — this was the right one to insist on. My docs said "a client that stops reading will stall the graph; that is the contract", which quietly folded *dead* into *slow*. Those are not the same trade, and only one of them is worth making by default.

`deliver_lossless_to` now uses `send_timeout`, bounded by a new `WebServerBuilder::lossless_stall_timeout` (default 30 s). On `Timeout` the subscriber is withdrawn from both the target snapshot and the registry, with a `log::warn!`, so it costs one timeout total rather than one per frame.

I went with `send_timeout` over ping/pong deliberately: the wait is for one slot in a 1024-deep queue, so it fires only when a subscriber has drained *nothing* for the whole window — which a live browser doing anything, however slowly, never does. That makes it precise for exactly the failure you describe without giving up losslessness for a merely slow client. Ping/pong is the fuller answer if idle-connection liveness is ever wanted for its own sake, but it needs a per-connection timer task and a close path, and it would not have made this case any more detectable.

You are right that it reaches teardown too, and the fix covers it: `finally`'s `block_on(deliver_lossless(..))` goes through the same bounded path.

The knob is public mainly because a 30 s timeout is otherwise untestable — `lossless_withdraws_a_subscriber_that_stops_accepting_frames` sets it to 300 ms. Verified it is what bounds the run: raise it to 25 s and the test fails at 27 s.

---

crates/wingfoil/src/adapters/web/server.rs
0-jake-0: Correct, and I took the first option in 87bc4b1 — narrowing the doc would have been fixing the comment rather than the bug. My own doc comment argued that a `WebServer` serves several runs, then stored the answer somewhere only one run can own; that is self-contradictory as written.

The resolved flag is now an `Arc<AtomicBool>` created in `publish_frames` alongside the semaphore, captured by `paced_sink`, the consumer closure, and the `finally`. `WebServerInner::resolved` is gone; `resolve_delivery` just answers, and the whole of `Auto`'s meaning now lives in `Delivery::resolves_lossless`.

Your third bullet is the one that made this more than a policy-correctness point — the two reads genuinely could straddle a flip, and the permit imbalance outlives the run that caused it. Per publish, the cell is written once at `start`, before any cycle and with the previous run's consumer already joined by `flush`, so no instant can span a write and the two reads provably agree.

Pinned by `concurrent_runs_on_one_server_resolve_delivery_independently`: one server, a live graph with a wedged subscriber overlapping a replay with a reading one, asserting both properties at once. Verified against the bug — make the flag process-wide and it fails on "the concurrent replay must stay lossless". Worth noting it can only *miss*, never flake red: if the live run finishes before the replay resolves, there is no overlap to catch. I said so next to the test rather than let a pass read as proof of overlap.

---

crates/wingfoil/src/adapters/web/server.rs
0-jake-0: Swapped in 87bc4b1 — lossless registry first, then `sender.subscribe()`.

Your read of the risk is right: the window has no `await` in it and `wait_for_subscribers` polls at 5 ms before a whole graph gets built, so the tests were never meaningfully racy. But a doc claim that is "true apart from a few instructions" is the kind that gets relied on later by something with tighter timing than these tests, and the fix is free. I left a comment on the two lines saying the order is load-bearing and why, so it does not get tidied back.

---

crates/wingfoil/src/adapters/web/write.rs
0-jake-0: Fixed in 87bc4b1. `deliver_lossless` was written for the one-frame `Complete` case and then reused inside the per-frame loop without re-examining what it does per call — so a mutex acquisition and a `Vec` allocation landed on the throughput-gating path.

It is now split: `lossless_targets` snapshots once, `deliver_lossless_to` takes `&mut LosslessTargets` and sends, and `deliver_instant` calls the first once per instant and the second per frame. The one-shot `deliver_lossless` survives as a thin wrapper for the `Complete` marker, which genuinely is a single frame outside any instant.

Your reasoning about the subscriber set is what makes the hoist safe, and it also gives the withdrawal path something better: a target that fails mid-instant is now dropped from the snapshot as well as the registry, so the rest of that instant skips it instead of re-discovering it frame by frame.

---

crates/wingfoil/src/adapters/web/server.rs
```diff
@@ -53,14 +153,129 @@ pub(crate) struct WebServerInner {
...
+        for (id, tx) in targets.iter() {
+            match tx.send_timeout(bytes.clone(), timeout).await {
+                Ok(()) => {}
+                Err(mpsc::error::SendTimeoutError::Closed(_)) => withdrawn.push(*id),
+                Err(mpsc::error::SendTimeoutError::Timeout(_)) => {
+                    log::warn!(
+                        "web_pub: subscriber on '{topic}' accepted nothing for {timeout:?} — \
+                         treating it as gone and dropping it from the lossless path"
+                    );
+                    withdrawn.push(*id);
```
0-jake-0: The bound is the right fix and I agree `send_timeout` beats ping/pong for it. One consequence of withdrawal, though, that I think is worth handling: **the withdrawn client is left as a zombie rather than told anything.**

`remove_lossless_sub` only touches the registry. The socket stays open, the connection task keeps running, and under lossless the `forward_broadcast` task for that topic is idle by design — so from this point the client receives no further data frames, *and* it never receives the end-of-run `Complete`, because `finally`'s `deliver_lossless` snapshots a registry the client is no longer in.

That matters because the JS client treats both as load-bearing: `onComplete` (`js/src/index.ts:349`) is the end-of-stream signal, and `handleComplete` is what suppresses reconnect. A withdrawn subscriber gets neither that nor a socket close, so its auto-reconnect never fires either. Net effect: an open socket, no data, no completion, no reconnect — indefinitely.

The case where this is more than theoretical is the recoverable one your doc calls out: a closed laptop lid. It suspends, trips the 300 ms/30 s bound, resumes, drains its queue — and then waits forever on a connection the server has quietly stopped feeding. A half-open peer never notices either way, so the failure mode belongs entirely to clients that *do* come back.

Closing the connection on withdrawal would resolve it — abort the connection task or drop its `outbound_tx` so the writer ends and the socket closes. The client then sees a clean close and reconnects on its own, which is the outcome you want for the recoverable case and harmless for the truly dead one. A `Complete`-style control frame would also work but only reaches a peer that is still listening, which is exactly the one you have just concluded isn't.

Smaller, same root: withdrawal takes the client off `lossless_subs` but leaves its broadcast receiver, so `subscriber_count` keeps reporting a subscriber the lossless path can no longer reach. That inverts the invariant the ordering swap in this commit just established. Closing the connection fixes this one too, since the receiver drops with it.

---

crates/wingfoil-python/src/adapters/web.rs
```diff
@@ -278,25 +292,37 @@ impl PyWebServer {
...
     ))]
+    #[allow(clippy::too_many_arguments)]
     fn new(
         addr: String,
         codec: String,
         static_dir: Option<String>,
         historical: bool,
         cert_path: Option<String>,
         key_path: Option<String>,
+        delivery: String,
```
0-jake-0, line 343: `PyWebServer.__init__` mirrors every other builder option — `codec`, `static_dir`, `historical`, `cert_path`/`key_path`, and `delivery` as added by the first commit — but `lossless_stall_timeout` from 87bc4b1 didn't follow. A Python user therefore gets the 30 s default with no way to change it, on a knob whose own doc says to shorten or lengthen it depending on how long a legitimate client can go quiet.

It's a small addition: another keyword in the `#[pyo3(signature = ...)]` taking seconds as an `f64`, plus a `lossless_stall_timeout_secs()` getter alongside `delivery_name()` for the tier-1 round-trip test that the Rust side already got in `web_adapter.rs`.

---

0-jake-0 (commented):
Re-reviewed at `87bc4b1`. All four findings are genuinely fixed, not papered over — I checked each against the code rather than the commit message. CI is green.

## The four

1. **Bounded wait** — `deliver_lossless_to` uses `send_timeout` against `WebServerInner::lossless_stall_timeout`, and `Timeout` withdraws from both the snapshot and the registry. Teardown is covered because `finally`'s `block_on` goes through the same path. Agreed on `send_timeout` over ping/pong: waiting for one slot in a 1024-deep queue means only a client draining *nothing* for the whole window trips it, which is the discrimination you want, and ping/pong would have needed a per-connection timer for no extra detection.
2. **Per-publish policy** — `WebServerInner::resolved` is gone; the flag is an `Arc<AtomicBool>` created in `publish_frames` next to the semaphore and captured by `paced_sink`, the consumer closure and the `finally`. That's the version that makes the two reads provably agree, so the permit imbalance is closed rather than made unlikely. `resolve_delivery` reducing to a pure answer and `Delivery::resolves_lossless` holding all of `Auto`'s meaning is a nice tightening on top.
3. **Registry before broadcast** — swapped, with a comment saying the order is load-bearing. Right call to fix rather than reword; agreed the tests were never meaningfully racy.
4. **Per-instant snapshot** — `lossless_targets` / `deliver_lossless_to` split, called once and per-frame respectively. The `Complete` marker keeping a one-shot wrapper is the right shape, and dropping a failed target from the snapshot as well as the registry is a real improvement over what I suggested.

The two tests are both verified against the bug they pin, and flagging in-tree that the concurrency test can only miss and never flake red is the honest way to write it.

## Residual

Two, inline, both consequences of the new withdrawal path rather than regressions:

- **`server.rs:260`** — a withdrawn subscriber is silently orphaned: no more frames, no `Complete`, no socket close, so the JS client neither completes nor reconnects. Recoverable clients (the closed laptop lid your own doc cites) are the ones this strands. Closing the connection on withdrawal would fix it, and would also stop `subscriber_count` reporting a client the lossless path has dropped.
- **`wingfoil-python/src/adapters/web.rs:320`** — `lossless_stall_timeout` didn't get a Python keyword, so Python users are pinned to the 30 s default while every other builder option is mirrored.

Neither blocks the fix this PR is making; the first is the one I'd want a decision on before merge, even if the decision is "document it and leave it".

---

crates/wingfoil/src/adapters/web/server.rs
0-jake-0: Fixed in 4b9d9a7, and this was a genuine hole in my own fix rather than a pre-existing one — I bounded the wait and then left the client with nothing to act on.

The framing that convinced me is yours: the failure mode belongs entirely to clients that *do* come back. A truly half-open peer is indifferent to how we withdraw it; the closed laptop lid is the case that suspends, trips the bound, wakes, drains, and then waits forever on a socket the server has quietly stopped feeding. Withdrawing without telling anyone converts a recoverable client into a permanently stuck one, which is worse than what it replaced for that client.

`deliver_lossless_to` now raises a per-connection `Notify` on timeout; the reader loop selects on it and tears the connection down. Two details worth recording:

- Teardown on that path **aborts** the writer instead of awaiting its drain. Awaiting would park forever on the same wedged socket and never drop the `WebSocket`, so the connection would stay open — the exact thing being fixed. The normal close path still awaits, so a queued `Complete` goes out.
- I close the whole connection, not the one subscription, because every topic on a connection shares one outbound queue. A stall is therefore a connection-level condition and the other topics are equally stuck; closing per-subscription would be the wrong unit.

Agreed on `Complete`-style control frame being the wrong tool here: it only reaches a peer still listening, which is the one we have just concluded isn't.

Your smaller point falls out of the same change — the broadcast receivers drop with the connection, so `subscriber_count` no longer advertises a client the lossless path has given up on, and the invariant the ordering swap established holds again. `a_withdrawn_subscriber_gets_its_connection_closed` asserts both halves, and is verified against the bug: drop the `close.notify_one()` and it fails on "must see its connection close".

One consequence I documented rather than fixed, in `js/README.md`: a client that reconnects rejoins a replay already in progress and has missed whatever went out while it was away. Losslessness is a property of a subscription, not of a topic, and reconnect-and-resume would need a replay cursor on the wire — well outside this PR.

---

crates/wingfoil-python/src/adapters/web.rs
0-jake-0: Added in 4b9d9a7 — `lossless_stall_timeout_secs` (default `30.0`) plus a `lossless_stall_timeout_secs()` getter. You are right that the omission was worse than a plain gap: the knob's own doc tells you to tune it to how long a legitimate client can go quiet, so a Python user was being pointed at a dial that wasn't there.

Followed aeron's `timeout_secs` rather than inventing a shape: seconds as an `f64` with a validating helper that rejects non-finite and non-positive values, so `lossless_stall_timeout_secs=0` raises at construction instead of silently becoming an instant timeout. The getter is what makes the round trip testable from Python, matching `delivery_name()` — `test_the_lossless_stall_timeout_defaults_and_is_selectable` and `test_a_non_positive_stall_timeout_raises`, plus a Rust unit test for the helper itself.

---

PR #911, opened 2026-08-23
fix(js): surface client boot failures

### LINKED ISSUE #457: wingfoil-js: handle boot() rejection, add WingfoilClient tests
Part of the codebase review follow-up (`docs/IMPROVEMENT_PLAN.md`, Phase 4).

- **Unhandled promise rejection**: `WingfoilClient` does `void this.boot()` (`src/index.ts`). If wasm `init()` rejects (bad `wasmUrl`, CSP), the error becomes an unhandled rejection and the client sits silently dead with no `ConnectionState` emitted. Catch and emit `{kind: "error"}`.
- **No client tests**: `WingfoilClient`'s subscribe bookkeeping, reconnect/resubscribe, and control-frame handling have zero tests — only `LatencyTracker` is covered. This logic is testable with a mock WebSocket (no wasm needed).
- `wingfoil-wasm`: a comment claims the error path is covered by `wasm-pack test`, but there are no `#[wasm_bindgen_test]`s and CI runs only native `cargo test`. Add wasm-target tests (the dev-dep is already declared) or fix the comment.

Depends on #449 (getting `pnpm test` into CI) to be meaningful.

### Acceptance
- [ ] `boot()` failure surfaces as an error state, not an unhandled rejection
- [ ] Mock-WebSocket tests for `WingfoilClient` bookkeeping
- [ ] wasm tests added or the stale comment corrected

k4its1t:
### Summary

- catch asynchronous `WingfoilClient.boot()` failures and emit an error connection state instead of leaving an unhandled rejection
- preserve the original `Error`, normalize non-Error rejection values, and avoid opening a WebSocket after initialization fails
- add focused initialization-failure tests and correct the stale native-test coverage comment

Closes #457

### Validation

- `pnpm test` — 42 passed
- `pnpm run lint`
- `pnpm run build:ts`
- `cargo test --manifest-path crates/wingfoil-wasm/Cargo.toml` — 8 passed
- `cargo lint`
- `cargo lint-all`
- `cargo nextest run -p wingfoil --all-features --lib --tests -E "not binary(/_integration$/)"` — 952 passed
- `cargo test --doc -p wingfoil --all-features` — 19 passed, 22 ignored
- `cargo test -p wingfoil-derive` — 3 passed
- repository pre-commit and pre-push hooks

A direct `cargo test -p wingfoil --all-features` also passed the 345 library tests and 20 Aeron adapter tests before entering the service-backed `aeron_integration` binary; that binary requires a Docker socket, which was unavailable locally. The CI-equivalent nextest command above compiles but filters service-backed integration binaries and passed completely.

### AI assistance

Implementation and test orchestration used OpenAI Codex assistance. The resulting patch was validated with the commands listed above.

### DIFF
--- a/crates/wingfoil-wasm/src/lib.rs
+++ b/crates/wingfoil-wasm/src/lib.rs
@@ -266,7 +266,7 @@ mod tests {
     #[test]
     fn parse_codec_accepts_bincode_and_json() {
         // `JsError::new` panics off-wasm, so only verify the success
-        // path here; the error path is covered by `wasm-pack test`.
+        // path here; the error path requires a wasm target.
         assert!(matches!(parse_codec("bincode"), Ok(CodecKind::Bincode)));
         assert!(matches!(parse_codec("json"), Ok(CodecKind::Json)));
     }
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -66,7 +66,7 @@ export type ConnectionState =
   | { kind: "connecting" }
   | { kind: "open"; codec: CodecKind; version: number }
   | { kind: "closed"; code: number; reason: string }
-  | { kind: "error"; error: Event };
+  | { kind: "error"; error: Event | Error };
 
 export interface ClientOptions {
   /** WebSocket URL, e.g. `ws://localhost:8080/ws`. */
@@ -229,7 +229,12 @@ export class WingfoilClient {
       wasmUrl: options.wasmUrl ?? "",
     };
     this.codecKind = this.opts.codec;
-    void this.boot();
+    void this.boot().catch((error: unknown) => {
+      this.emitConn({
+        kind: "error",
+        error: error instanceof Error ? error : new Error(String(error)),
+      });
+    });
   }
 
   /** Close the socket and stop reconnecting. */
--- a/js/tests/client.test.ts
+++ b/js/tests/client.test.ts
@@ -3,11 +3,12 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 const wasm = vi.hoisted(() => ({
   control: {} as { Hello?: { codec: string; version: number } },
   encodePayload: vi.fn(() => new Uint8Array([1, 2, 3])),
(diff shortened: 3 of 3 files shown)


0-jake-0 (commented):
Reviewed the full diff against `main` and the #457 acceptance list. This is a clean, well-scoped fix — recommend merge.

**What checks out:**

- The `.catch` on `boot()` is the right shape: the rejection lands in a microtask, so a listener registered synchronously after construction (the normal pattern) sees the error state; `connect()` is only reached after `await init()` succeeds, so no socket is opened on a failed init — both boot tests assert exactly these two properties.
- Normalizing a non-`Error` rejection with `new Error(String(error))` while preserving a real `Error` is correct, and the widened variant `{ kind: "error"; error: Event | Error }` breaks nothing in-repo — `index.ts` and the tests are the only places that touch that variant.
- The `wasm.init` mock is reset in the pre-existing `publish` and `Hello` describes too, so the `mockRejectedValueOnce` in the boot tests can't leak across suites regardless of test order.
- The `wingfoil-wasm` comment fix takes the "or fix the comment" arm of #457's third checkbox accurately — CI does only run native `cargo test` there.
- CI is green across all nine checks, and the validation list matches what the repo's pre-commit checklist asks for.

**Two non-blocking notes:**

1. Widening `error` from `Event` to `Event | Error` is a (mild) breaking change for TypeScript consumers of `@wingfoil/client` who narrow that field — e.g. code doing `state.error.type` no longer typechecks. Worth one line in the release notes when the next version is cut; nothing to change in this PR.
2. `Closes #457` is fair — the first and third checkboxes are fully done here, and the second (mock-WebSocket client tests) has since been substantially covered by the existing `publish`/`Hello`/`shouldReconnect` suites. The one bookkeeping path still untested is resubscribe-on-reconnect (the socket `open` handler replaying `subscribedTopics()`); fine as a small follow-up if we ever want it.

Thanks for the tight scoping and the disclosed validation, @k4its1t — this is exactly the shape of fix the issue asked for.

---

PR #919, opened 2026-08-24
perf(zmq): avoid publisher hot-path polling and cloning

### LINKED ISSUE #826: zmq publisher: unconditional per-message monitor poll (port regression) and per-message payload clone
From the fable-review-2 codebase review (both findings adversarially verified, including against the legacy tree in git history).

**1. Monitor poll — a verified port regression.** `publish` calls `check_monitor()` (a `zmq::poll` syscall + drain loop) unconditionally per published message for the whole run (`crates/wingfoil/src/adapters/zmq.rs:549`), though its only purpose — noticing the first `ZMQ_EVENT_ACCEPTED` — is spent once `subscriber_connected` flips true (`zmq.rs:558` is the only write; the monitor is filtered to `ZMQ_EVENT_ACCEPTED` only at `zmq.rs:495`, so there is no disconnect-detection logic to break). **Legacy guarded the call** — `if !self.subscriber_connected { self.check_monitor(); … }` (legacy `zmq/write.rs:64-65`) — so the unconditional call is a regression introduced in the port, and the one-line guard restores legacy behavior while removing a per-message syscall from the graph thread.

**2. Payload clone.** Each tick clones `T` solely to build the owned `WireMessage::Value(value.clone())` for serialization (`zmq.rs:666`). Rather than a borrowing twin enum (the variant order is a documented wire contract, `zmq.rs:179-182` — a twin is a drift hazard), use a one-off wrapper implementing `Serialize` via `serializer.serialize_newtype_variant("WireMessage", 3, "Value", value)` — bincode encodes only the variant index, so output is provably byte-identical. Add a case to the existing golden test `wire_format_matches_legacy_message` (`zmq.rs:735-772`) to pin it. Minor severity: payloads here are typically small and `bincode::serialize` already allocates per message.

maxi-maxima:
## What this changes

- Stops polling the publisher monitor after the first subscriber has completed its connection propagation window.
- Serializes `WireMessage::Value` through a borrowed wrapper, removing the per-message payload clone while preserving bincode variant index 3.
- Extends the golden wire-format regression and refreshes the ZMQ maintenance notes that describe the active test and compatibility contract.

## Why

The publisher hot path kept performing a non-blocking `zmq::poll` after the only event it consumes had already been observed, and cloned every payload solely to construct an owned serialization envelope. Restoring the legacy one-shot monitor guard and borrowing the serialized value removes both costs without changing socket, buffering, or wire behavior.

Closes #826.

## How it was verified

- [x] `cargo fmt --all`
- [x] Upstream Linux `Test (wingfoil)`
- [x] Upstream Linux `Lint (fmt & clippy)`
- [x] Upstream Python interop and security checks
- [x] Fork-dispatched [ZMQ integration workflow](https://github.com/maxi-maxima/wingfoil/actions/runs/32683109779): core pub/sub, etcd discovery, Python discovery, and Rust ↔ Python cross-language tests
- [x] New behavior is covered by a golden byte regression
- [x] `git diff --check`

The Windows host lacks the MSVC `link.exe`, so local Cargo commands that compile dependencies stop before project code. The upstream Linux matrix and the adapter-specific workflow provide the build, lint, and runtime evidence instead.

## Notes for the reviewer

`BorrowedValue` deliberately uses `Serializer::serialize_newtype_variant("WireMessage", 3, "Value", value)`. The unit test compares that encoding to the owned enum encoding, which is independently pinned to the longhand legacy bytes.

The ZMQ maintenance guide already contained stale post-cutover references (the retired cross-engine test, resolved C2 deviation, old example path, and old feature list). They are corrected here because this change adds the new hot-path invariants to that living document.

### DIFF
--- a/crates/wingfoil/src/adapters/zmq.rs
+++ b/crates/wingfoil/src/adapters/zmq.rs
@@ -107,7 +107,7 @@ use std::time::{Duration, Instant};
 
 use anyhow::{Context, Result};
 use serde::de::DeserializeOwned;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use wingfoil::RunMode;
 
 use crate::Burst;
@@ -211,6 +211,25 @@ enum WireMessage<T> {
     /// [`run_subscriber`]) to route the error through the same match arm, and
     /// a legacy peer may send it.
     Error(String),
+}
+
+/// Borrowed encoding of [`WireMessage::Value`].
+///
+/// `bincode` serializes enum variants by index, so this wrapper emits index 3
+/// directly while borrowing the stream value instead of cloning it into an
+/// owned [`WireMessage`].
+struct BorrowedValue<'a, T>(&'a T);
+
+impl<T> Serialize for BorrowedValue<'_, T>
+where
+    T: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_newtype_variant("WireMessage", 3, "Value", self.0)
+    }
 }
 
 /// Serialize the payload-free `EndOfStream` frame. The variant carries no `T`,
@@ -546,7 +565,9 @@ impl ZmqPubState {
 
     /// Publish (or buffer) one already-serialized message.
     fn publish(&mut self, data: Vec<u8>) -> Result<()> {
-        self.check_monitor();
+        if !self.subscriber_connected {
+            self.check_monitor();
+        }
         if !self.subscriber_connected
             && let Some(accepted_at) = self.accepted_at
         {
@@ -663,7 +684,7 @@ where
                       _s: &mut (),
                       value: &T,
                       _ctx: &mut Ctx<'_>| {
-                    let data = bincode::serialize(&WireMessage::Value(value.clone()))
+                    let data = bincode::serialize(&BorrowedValue(value))
                         .context("zmq_pub: serializing message")?;
                     state.borrow_mut().publish(data)?;
                     Ok(Tick::Value(()))
@@ -721,12 +742,11 @@ mod tests {
     /// level.
     ///
     /// The expected bytes are written out longhand rather than compared against
-    /// legacy's encoder, because legacy's `Message` is `pub(crate)` and cannot
-    /// be reached from here — and because a golden encoding catches a drift on
-    /// *either* side, where a cross-check against the neighbouring crate would
-    /// silently follow it if both moved together. The behavioural half of this
-    /// (a real legacy socket at the other end) is
-    /// `tests/zmq_cross_engine_integration.rs`.
+    /// legacy's encoder, because that retired type was `pub(crate)` and cannot
+    /// be reached from here — and because a golden encoding catches drift
+    /// without following a neighbouring implementation. The current
+    /// behavioural half is `tests/zmq_cross_lang_integration.rs`, which sends
+    /// these frames over real sockets between Rust and Python peers.
     ///
     /// bincode 1.x encodes an enum as a little-endian u32 variant index
     /// followed by the payload, so these indices *are* legacy's declaration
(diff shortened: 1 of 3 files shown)


0-jake-0 (commented):
Reviewed against #826. Both halves do what the issue asked, and I verified the reasoning rather than taking it on trust. LGTM, with one suggestion and one note that are both non-blocking.

**1. Monitor guard — verified semantics-preserving**

`subscriber_connected` has exactly one write site (`zmq.rs:558`) and is never reset, `check_monitor` has exactly one caller (`publish`), and `accepted_at` — the only state `check_monitor` mutates — is read only under `!subscriber_connected` (`zmq.rs:550`) and in the buffering branch, which is itself unreachable once connected. So after the flip the skipped drain observes nothing the publisher can act on. This is the legacy guard restored, as #826 documents.

*Note, not a change request:* the flip side is that the monitor queue is never drained again, so `ZMQ_EVENT_ACCEPTED` events for every later subscriber and every reconnect accumulate unread for the process lifetime. That property is inherited from legacy rather than introduced here, so it shouldn't hold up this PR — flagging it only in case a long-lived publisher with flapping subscribers makes it worth a follow-up. The structural fix would be to tear the monitor down once it is spent instead of conditionally skipping it, but that needs care: `rust-zmq` 0.10's `Socket::monitor` takes `&str`, so there is no way to reach `zmq_socket_monitor(sock, NULL, 0)` to stop monitoring cleanly, and dropping the `PAIR` socket alone leaves libzmq's internal monitor socket sending to a peerless endpoint. Worth an issue rather than a scope increase here.

**2. `BorrowedValue` — correct, but there may be a simpler form worth considering**

The encoding is right (`CheckPoint`, `EndOfStream`, `HistoricalValue`, `Value` → index 3 ✓) and the new test is ordered the right way round: `assert_eq!(value, vec[3, 0, 0, 0, …])` trips first if anyone reorders the variants, so the borrowed comparison can't quietly follow a drift.

That said, the clone can be removed without hand-writing the variant index at all, by instantiating the existing enum at `&T`:

```rust
let data = bincode::serialize(&WireMessage::<&T>::Value(value))
    .context("zmq_pub: serializing message")?;
```

`#[derive(Serialize)]` gives `impl<T: Serialize> Serialize for WireMessage<T>`, and `&T: Serialize` whenever `T: Serialize`, so this needs no new type, no new impl and no literal `3`. I checked it rather than assuming — a standalone crate with the same variant order and serde/bincode 1.x confirms it is byte-identical to the owned encoding for both a `Copy` payload (`u64` → `[3,0,0,0,7,0,0,0,0,0,0,0]`) and a non-`Copy` one (`String`).

#826 ruled out a *borrowing twin enum* as a drift hazard, and rightly so — but this isn't a twin. It's the same declaration instantiated differently, which means the variant order it emits cannot diverge from `WireMessage`'s at all, where the literal `3` in `serialize_newtype_variant` can (it is only caught today because the golden test happens to pin both). Your call, and the current form is tested and correct, so this is a suggestion rather than a request.

**3. Doc corrections — checked against `main`, all accurate**

Confirmed each claim: `tests/zmq_cross_engine_integration.rs` is gone from the tree, `Cargo.toml:510` reads `zmq = ["dep:zmq", "dep:bincode"]`, the example is declared at `examples/adapters/zmq/main.rs`, and `zmq.rs`'s own module header already says "two deliberate ways" and states the wire envelope is explicitly *not* a deviation — so `CLAUDE.md` was the stale side of that disagreement and this brings it back in line with the canonical list. Folding the two new hot-path invariants into the same living document is the right move.

One leftover, if you want it in scope: `docs/planning/deviation-register.md` C2 still cites `tests/zmq_cross_engine_integration.rs` as one of three live guards. That page is a historical record rather than a description of `main`, so leaving it may well be deliberate — maintainer's call.

**Verification**

CI is green on the head commit, and `Test (wingfoil)` runs `--all-features`, so the new golden-byte assertion genuinely executed rather than being gated off. Thanks for dispatching the ZMQ integration workflow on your fork to cover the socket-level tests the upstream required checks don't reach — that was the right way to close the gap left by the local build failing.

Recommending merge to the maintainer.

---

PR #920, opened 2026-08-24
refactor(core): centralize Tick slot writes

### LINKED ISSUE #829: interp.rs: 13 hand-copied Tick match/slot-write blocks — extract a `store_tick` helper

The 10-line `match Tick { Value / Silent / Quiet }` → write-slot block appears 13 times in `crates/wingfoil/src/interp.rs` (grep on `Tick::Silent(v) =>`: lines 1365, 1797, 1873, 1951, 2020, 2089, 2134, 2187, 2254, 2406, 2509, 3845, 3888 — `with_time`, `trimap`, `try_trimap`, `bimap`, `try_bimap`, `poll`, `latency_report`, `latency_report_each`, `composite`, `custom_node`, `Extension::map`, `Extension::fold`, …). The copies are mechanically identical modulo which input borrows get dropped first — and that drop-borrow-before-write ordering is subtly load-bearing, re-implemented by hand in each copy.

**Fix:** one helper — `fn store_tick<T>(tick: Tick<T>, out: &SlotRef<T>) -> bool` — called after input borrows are dropped. `Tick` owns the value at match time, so `let t = Op::cycle(...)?; drop(borrows); Ok(store_tick(t, &out))` preserves the ordering exactly, and a future change to `Silent` semantics or slot-write ordering lands in one place.

**Deliberately out of scope** (considered and rejected during verification): merging the `bimap`/`try_bimap` and `trimap`/`try_trimap` registration twins behind generic inners — those pairs differ in `Op` type and closure signature, and unifying them adds a generics layer that obscures the deliberately plain hand-written `#[op]`-seam pattern for marginal savings.

k4its1t:
## What this changes

Adds a shared `store_tick` helper for interpreted nodes and replaces the 13 hand-copied `Tick` match/slot-write blocks with it.

## Why

This keeps the `Value`, `Silent`, and `Quiet` storage semantics in one place while preserving the explicit drop-before-write ordering for paths that borrow input slots.

Closes #829

## How it was verified

- [x] `cargo nextest run -p wingfoil --all-features --lib --tests -E 'not binary(/_integration$/)'` (955 passed)
- [x] `cargo test --doc -p wingfoil --all-features`
- [x] `cargo nextest run -p wingfoil-derive` (3 passed)
- [x] pre-push `cargo build --workspace --all-targets` and `cargo test --workspace`

## Notes for the reviewer

This is a behaviour-preserving refactor. Borrowed inputs are dropped before `store_tick` mutably borrows the output slot, matching the original ordering.

## AI assistance

Codex assisted with locating the duplicated blocks, applying the mechanical refactor, and running the validation commands. I reviewed the final diff and verification results before submission.

### DIFF
--- a/crates/wingfoil/src/interp.rs
+++ b/crates/wingfoil/src/interp.rs
@@ -436,6 +436,21 @@ impl<T> SlotRef<T> {
     }
 }
 
+/// Store a tick's value and report whether downstream nodes should run.
+fn store_tick<T>(tick: Tick<T>, out: &SlotRef<T>) -> bool {
+    match tick {
+        Tick::Value(value) => {
+            *out.borrow_mut() = value;
+            true
+        }
+        Tick::Silent(value) => {
+            *out.borrow_mut() = value;
+            false
+        }
+        Tick::Quiet => false,
+    }
+}
+
 /// Trim a `type_name` — `wingfoil::ops::Map<u64, …, {{closure}}>` — down
 /// to the bare op name (`Map`) for error context: drop everything from the
 /// first `<`, then keep only the final `::` segment. A plain label with no
@@ -1420,17 +1435,7 @@ impl Builder {
             Box::new(move |k| {
                 let (cfg, state) = &mut *cs_cycle.borrow_mut();
                 let mut ctx = Ctx::new(k, idx);
-                match body(cfg, state, &mut ctx)? {
-                    Tick::Value(v) => {
-                        *out_cycle.borrow_mut() = v;
-                        Ok(true)
-                    }
-                    Tick::Silent(v) => {
-                        *out_cycle.borrow_mut() = v;
-                        Ok(false)
-                    }
-                    Tick::Quiet => Ok(false),
-                }
+                Ok(store_tick(body(cfg, state, &mut ctx)?, &out_cycle))
             }),
             Box::new(|_| Ok(())),
         );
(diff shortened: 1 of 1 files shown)


0-jake-0 (commented):
Reviewed against #829. This is exactly the refactor the issue asked for, and I verified it is behaviour-preserving rather than taking that on trust. LGTM.

**What I checked**

- **Coverage is complete and correctly bounded.** `main` has 14 `Tick::Quiet => Ok(false)` arms in `interp.rs`; 13 are converted here. The fourteenth (`never`, `interp.rs:1716`) is deliberately left alone, and that is the right call — it matches `Tick::Value(()) | Tick::Silent(())` into `Ok(false)` and never writes its slot at all (the slot exists only for index alignment), so `store_tick` would change its meaning. Good that it wasn't swept up mechanically.
- **The load-bearing drop-before-write ordering is preserved.** Every borrowing site now reads `let tick = …cycle(…)?; drop(borrows); Ok(store_tick(tick, &out))`. `Tick<T>` owns its payload, so hoisting the match out past the drop is sound.
- **The one ordering change is on the `Quiet` path**, where input borrows are now released slightly earlier than they were (previously they lived to end of closure). No `borrow_mut()` is taken on that path either way, so no behaviour depends on it.
- `drop((va, vb))` is equivalent to the previous `drop(va); drop(vb);`.
- CI is green on the head commit, including `Test (wingfoil)` (`--all-features`) and both clippy passes.

**One optional nit, no action needed**

`store_tick` is a small generic monomorphised per instantiation, so LLVM will inline it at every call site; I mention it only because these closures sit on the per-cycle hot path and an explicit `#[inline]` would make that guarantee cheap rather than incidental. Fine as-is.

Net −113 lines with the `Value`/`Silent`/`Quiet` storage rule stated once. Recommending merge to the maintainer.

---

PR #921, opened 2026-08-25
perf(core): trim demux hot-path overhead

### LINKED ISSUE #827: demux: double clone per activation; demux_it: per-item RefCell borrows

Two small but genuine costs on the deliberately hot, ungated routing primitive:

1. **`demux` clones the routed value twice per activation** (`crates/wingfoil/src/interp.rs:4279-4284`): `let value = source_slot.borrow().clone();` then `*publish.borrow_mut() = value.clone();`. One clone suffices — route from the borrowed slot, then clone once: `let src = source_slot.borrow(); let slot = route(&src).min(size); *publish.borrow_mut() = src.clone();`. Safe: `route` is `Fn(&T) -> usize` and cannot access the slots.

2. **`demux_it` re-acquires two `RefCell` borrows per item of every burst** (`interp.rs:4405-4414`): `idxs_cycle.borrow()` + `marks.borrow_mut()` inside the per-item loop, though both are loop-invariant. Hoisting both above the loop is safe — the loop body touches only `rows`, the `DemuxMap`, and the user closure, none of which can alias those RefCells.

`T` on a demux parent is often a large payload, so the second clone is pure waste; the borrow churn is the kind of per-item overhead this repo budgets in nanoseconds.

maxi-maxima:
## What this changes

Routes `demux` directly from the borrowed source value so each activation performs only the publication clone, and hoists `demux_it`'s child-index and mark-buffer `RefCell` borrows out of the per-item loop. Adds a clone-count regression that pins the hot-path invariant.

## Why

`demux` previously cloned every routed payload twice in its parent node, while `demux_it` reacquired two loop-invariant borrows for every item in a burst. Both costs sit on the deliberately ungated routing path.

Closes #827.

## How it was verified

- [x] `git diff --check`
- [x] Upstream `Test (wingfoil)` (includes `--all-features`)
- [x] Upstream `Lint (fmt & clippy)`
- [x] Upstream Python Interop, Cargo/pnpm audits, and dependency review

The focused local test attempt was blocked before project code compiled because this Windows host has no MSVC `link.exe`; the upstream Linux matrix compiled and ran the change successfully.

## Notes for the reviewer

The route closure still receives `&T`; the source `Ref` remains live only through routing and the single publication clone. `demux_it` holds the two loop-invariant borrows for the duration of the burst, with no additional aliasing surface inside the loop.

### DIFF
--- a/crates/wingfoil/src/interp.rs
+++ b/crates/wingfoil/src/interp.rs
@@ -4234,9 +4234,9 @@ impl Builder {
         let idxs_cycle = child_indices.clone();
         let publish = parent_out.clone();
         let cycle: CycleFn = Box::new(move |_k| {
-            let value = source_slot.borrow().clone();
-            *publish.borrow_mut() = value.clone();
-            let slot = route(&value).min(size); // `>= size` → overflow (last entry)
+            let source = source_slot.borrow();
+            let slot = route(&source).min(size); // `>= size` → overflow (last entry)
+            *publish.borrow_mut() = source.clone();
             let target = idxs_cycle.borrow()[slot];
             marks.borrow_mut().push(target);
             Ok(false)
@@ -4360,6 +4360,8 @@ impl Builder {
             for row in rows.iter_mut() {
                 row.clear();
             }
+            let child_indices = idxs_cycle.borrow();
+            let mut marks = marks.borrow_mut();
             for item in items {
                 let (key, event) = func(&item);
                 let slot = match event {
@@ -4368,8 +4370,8 @@ impl Builder {
                 }
                 .unwrap_or(capacity); // overflow slot == capacity (last row)
                 rows[slot].push(item);
-                let target = idxs_cycle.borrow()[slot];
-                marks.borrow_mut().push(target);
+                let target = child_indices[slot];
+                marks.push(target);
             }
             Ok(false)
         });
--- a/crates/wingfoil/tests/demux.rs
+++ b/crates/wingfoil/tests/demux.rs
@@ -7,6 +7,8 @@
 //! so it is available (and must stay tested) on default features, exactly as
 //! legacy's ungated `nodes/demux.rs` is.
 
+use std::cell::Cell;
+use std::rc::Rc;
 use std::time::Duration;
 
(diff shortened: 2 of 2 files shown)


0-jake-0 (commented):
Reviewed against #827. This does exactly what the issue specified, and I re-derived the safety argument rather than taking it on trust — both hoists hold up.

**`demux` (interp.rs:4234)** — holding the `source_slot` borrow across `publish.borrow_mut()` cannot panic here: `parent_out` comes from `self.new_slot(T::default())`, so it is a freshly allocated `RefCell` that can never alias `source_slot`. And `route` is `Fn(&T) -> usize`, handed only a `&T` — the slot `Rc`s are private to `Builder`, so a user closure has no path to re-enter either cell. Clone count drops 3 → 2 (parent publish + selected child), which is the point on a large payload.

**`demux_it` (interp.rs:4360)** — the two hoisted borrows are loop-invariant, and the loop body touches only `rows`, the `DemuxMap`, and `func`. `func` is `Fn(&T) -> (K, DemuxEvent)`, so it cannot reach `idxs_cycle` or the graph-wide `marks` buffer; `DemuxMap` holds neither. Taking `marks.borrow_mut()` for the whole burst is fine given node cycles are serialised. Empty-burst case now takes a borrow it doesn't use, which costs nothing.

**The regression test earns its place.** It fails on the old code for two independent reasons — `clones == 3` and `route` observing `1` rather than `0` — so it pins both halves of the invariant, not just the count. Nice touch making the route closure record what it saw.

One subtlety worth a comment in the test, since it took a second read: `CloneCounted::default()` mints a *fresh* `Rc<Cell>`, so clones of the default-initialised slot values are tallied into throwaway counters and never reach `clones`. That is precisely why the assertion can be an exact `2` instead of a range — but it is load-bearing and invisible. A one-line note on the `Default` impl saying the default counter is deliberately unobserved would stop a future reader "simplifying" it into a shared counter and quietly turning the assertion into a moving target.

Verification: full CI green here, including `Test (wingfoil)` with `--all-features` and `Lint (fmt & clippy)`, so the Windows `link.exe` gap in your local run is covered. No further changes needed from my side — the note above is optional.

---

PR #922, opened 2026-08-25
feat(python): expose remaining core stream ops

### LINKED ISSUE #106: Python parity: expose the remaining core ops (ewma, join_passive, finally, for_each, …)
> **Rescoped 2026-08-03** (tracker audit, verified against `next` @ c8cd740).
> This was filed against legacy ("audit all 37 nodes in `wingfoil/src/nodes/`"),
> a directory that does not exist on next. **The adapter half is complete** and
> the op half is close. What remains is a short, concrete checklist.

## Adapters — done

All fifteen next adapters have Python bindings under
`crates/wingfoil-python/src/adapters/`: aeron, augurs, csv, etcd, fix, fluvio,
iceoryx2, kafka, kdb, otlp, postgres, prometheus, redis, web, zmq. Each is
feature-gated with an `all-adapters` roll-up (`Cargo.toml`), and there are 20
adapter pytest modules under `crates/wingfoil-python/tests/`.

Wheel coverage — including aeron and iceoryx2 on Linux — was settled in #367.

The original checklist items map as: KDB+ ✅, ZMQ ✅, Kafka ✅, CSV ✅.
Arrow/Parquet is #104 and SQL is postgres ✅ — neither belongs to this issue.

## Ops — the actual remaining gap

Present on the Python `Stream` today (`src/graph.rs`): `accumulate`, `bimap`,
`buffer`, `collect`, `constant`, `count`, `counter`, `custom_node`, `dataframe`,
`delay`, `difference`, `distinct`, `drop_small_change`, `filter`, `filter_map`,
`filter_none`, `filter_value`, `fold`, `inspect`, `join`, `limit`, `logged`,
`map`, `mean`, `merge`, `merge_all`, `not_`, `print`, `reduce`, `sample`,
`split`, `sum`, `throttle`, `timed`, `values`, `window`, `with_time`, plus the
rolling/cumulative/time-windowed family in `src/statistics.rs` and the
`feedback` / `poll` / `external` / `channel` sources.

Missing, each a small `pyop_fn!` / `#[pyop]` addition:

- [ ] `ewma` on `Stream` — exists as a free fn in `statistics.rs:373` but is not a `Stream` method
- [ ] `join_passive` (and `try_join_passive`)
- [ ] `delay_with_reset`
- [ ] `ticked_at` / `ticked_at_elapsed`
- [ ] `finally` (as `finally_`, matching `not_`)
- [ ] `for_each`

## Acceptance

- [ ] The six ops above exposed on the Python `Stream`
- [ ] A pytest per op, following the existing `test_statistics.py` shape
- [x] All adapters accessible from Python (gated on optional deps)

See `/new-op-next` for the binding step and `/bind-adapter-next` for adapters.

tunglambk:
## What this changes

Exposes `join_passive`, `try_join_passive`, `delay_with_reset`, `ticked_at`, `ticked_at_elapsed`, `for_each`, and `finally_` on Python `Stream`. Adds object-form and Python round-trip regressions plus updates the Python API and migration docs.

## Why

Closes #106. `Stream.ewma` already exists on current `main`; these methods complete the remaining checklist without duplicating it.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all`
- [ ] `cargo test -p wingfoil --all-features`

Also ran:

- `cargo clippy -p wingfoil-python --all-targets -- -D warnings`
- `cargo test -p wingfoil-python` (91 unit tests and 23 plugin-seam tests passed)
- `maturin develop -F extension-module`
- `pytest -q tests/test_interop.py` (110 passed, 1 skipped)
- the pre-push workspace build completed; the workspace test reached two existing trybuild snapshot mismatches caused by changed compiler diagnostic wording
- `cargo lint` reaches an existing `uninlined_format_args` warning in `crates/wingfoil/tests/catalog.rs:149`

## Notes for the reviewer

Both passive-join spellings use the fallible Rust op because Python callables can raise in either form. `finally_` keeps the Python keyword-safe spelling requested by the issue.


docs/python-interop.md
```diff
@@ -325,7 +325,7 @@ only the wiring seams are dynamic.
...
+| Built-in combinator surface on `PyStream` (legacy `test_streams` parity plus wingfoil-only additions) | `map`/`filter`/`merge`/`merge_all`/`delay`/`delay_with_reset`/`distinct`/`count`/`limit`/`skip`/`skip_while`/`step_by`/`take_while`/`throttle`/`sample`/`difference`/`pairwise`/`enumerate`/`neg`/`inspect`/`for_each`/`finally`/`accumulate`/`buffer`/`window`/`with_time`/`ticked_at`/`ticked_at_elapsed`/`collect`/`fold`/`reduce`/`filter_map`/`filter_value`/`filter_none`/`sum`/`mean`/`average`/`ewma`/`bimap`/`join_passive`/`try_join_passive`/`split`/`dataframe` — all with Python-exception propagation; `Vec`→list & `(nanos,value)` tuple edge conversions | ✅ done (`graph.rs`, `python.rs`; PRs #549/#551 + follow-ups) |
```
0-jake-0, line 328: Nit: this row lists `finally`, but the Python-facing name is `finally_` — which is the spelling you (correctly) use in `api.rst`, `migration.rst`, and the crate README in this same PR. Since this table is enumerating the Python surface, it should read `finally_` too, otherwise it is the one place a reader could copy a name that raises `SyntaxError`.

Everything else in the rewritten row checks out against `python.rs` on this branch — including the `not` → `neg` correction, which was already stale on `main`.

---

crates/wingfoil-python/README.md
```diff
@@ -278,27 +280,23 @@ I/O sources are module-level functions taking the graph first — see
...
 
-Cumulative `.sum()` and `.mean()` are `Stream` methods. The full windowed
-moment surface — rolling `variance`, `std`, `median`, `min`/`max`, time- and
-count-weighted windows, and `ewma` — lives in the engine's Rust `adapters::statistics` layer
-and reaches Python through the [plugin seam](#authoring-components-in-rust),
-as a `#[pyop]` exposing exactly the window you want. That keeps the
-parameterisation where it can be type-checked rather than in extra Python
-classes. This is the one place the binding is currently narrower than legacy's;
-the engine already has the whole surface.
+The `Stream` methods `.sum()`, `.min()`, `.max()`, `.mean()`, `.variance()`, `.std()`, `.median()`, and `.ewma()` expose the engine's count-, time-, and window-parameterized statistics surface directly. See the [Python API documentation](https://wingfoil.readthedocs.io/en/latest/api.html#statistics) for the accepted `Window`, `Weighting`, and `EwmaSpan` forms.
```
0-jake-0, line 299: The substance here is right and worth fixing — I checked `python.rs` on `main` and `sum`/`min`/`max`/`mean`/`variance`/`std`/`median`/`ewma` are all `Stream` methods taking the `Window`/`Weighting`/`EwmaSpan` knobs, so the old "this is the one place the binding is currently narrower than legacy" paragraph was stale. Good catch, and the readthedocs link matches the one the crate README already badges.

Formatting nit: this lands as a single 362-character line, where the prose around it wraps at ~80. The long lines elsewhere in this README are all table rows, which can't wrap — this is the only prose paragraph that doesn't. Worth re-wrapping to match the paragraph it replaced.

One judgement call for @0-jake-0 rather than a request: the replaced text also explained that the parameterised window surface reaches Python *through the plugin seam* as a `#[pyop]`, "so the parameterisation stays where it can be type-checked". That rationale disappears here. If it is still true of how you want people to extend the statistics surface, a sentence keeping it would be worth carrying over; if the direct `Stream` methods have superseded it, dropping it is correct as written.

---

0-jake-0 (commented):
Reviewed against #106. This closes the op checklist completely and the implementation follows the existing binding patterns closely. Two doc nits inline; nothing blocking.

**No CI has run on this PR** — `get_status` and `get_check_runs` both return zero, so as a fork PR it is waiting on @0-jake-0 to approve workflow runs. Since you'd left `cargo lint`, `cargo lint-all` and the test gate unchecked in the description, I ran the parts I could locally on `9a27215`:

- `cargo clippy -p wingfoil-python --all-targets -- -D warnings` — clean
- `cargo test -p wingfoil-python` — 91 unit + 23 plugin-seam + doctests, all pass
- `maturin develop -F extension-module` then `pytest -q tests/test_interop.py` — **110 passed, 1 skipped**

The pytest leg matters most: the Rust unit tests exercise `graph.rs::PyStream`, but `python.rs` — where `finally_`, `for_each`, `join_passive` and `delay_with_reset` are actually exposed to Python — is only covered by `test_interop.py`. That surface works.

On the API choices, all three hold up:

- **Routing both passive-join spellings through `try_join_passive`** is the right call. The Rust distinction is infallible-vs-fallible, and a Python callable is always fallible, so an infallible `join_passive` would be a lie. Threading `label` through so the two produce distinguishable error messages is a nice detail, and #106 asked for both names.
- **`finally_`** matches the `not_` precedent the issue explicitly cited.
- **`ticked_at` returning nanos as `u64`** is consistent with `with_time`/`collect` already handing Python ints rather than a `NanoTime` wrapper.

I verified the two non-obvious test expectations by hand rather than trusting them: `join_passive_only_ticks_with_the_active_stream` expecting `[(0, 11), (100, 23), (200, 35)]` is right — at t=100 the 50ns passive stream has ticked three times, so its current value is 3, and the active stream only fires on its own ticks. And `ticked_at_elapsed` returning `[0, 100, 200]` off `start_nanos=1_000` matches `now - start`.

Two optional coverage gaps, both already covered by analogous Rust tests, so take or leave: there's no pytest for `try_join_passive`'s *success* path (only its exception path), and none for `join_passive`'s exception path.

Worth noting when this merges: the `ewma` checkbox on #106 is stale rather than unaddressed — `Stream.ewma` is already on `main` at `python.rs:428`, as your description says. The issue body should get that checked off too so the closure reads cleanly.

---

Issue #916, opened 2026-08-23
Move the PyPI upload into release.yml so attestations can come back


0-jake-0:
The 9.0.0 release publishes to PyPI with `attestations: false`, and it has to, because `pypi-publish.yml` is a reusable workflow. Two claims name the workflow and they disagree:

| | Claim | Names |
|---|---|---|
| Token exchange | `job_workflow_ref` | `pypi-publish.yml` (the workflow that runs the upload) |
| Attestation | Sigstore cert Build Config URI | `release.yml` (the caller) |

With the publisher registered as `pypi-publish.yml`, auth succeeds and the upload is rejected:

```
400 Bad Request
Invalid attestations supplied during upload: Verification failed:
Certificate's Build Config URI (.../release.yml@refs/heads/main)
does not match expected Trusted Publisher (pypi-publish.yml @ wingfoil-io/wingfoil)
```

Registering `release.yml` instead just moves the failure earlier, to `invalid-publisher` at the token exchange — PyPI looks the publisher up by `job_workflow_ref` and never checks `workflow_ref` ([`GitHubPublisher.lookup_by_claims`](https://github.com/pypi/warehouse/blob/main/warehouse/oidc/models/github.py)). No single registration satisfies both, which is what PyPI means by "reusable workflows are not supported".

### The fix

Move the upload job out of `pypi-publish.yml` and into `release.yml`. When the job is defined in the entry-point workflow, `workflow_ref` and `job_workflow_ref` both name `release.yml`, the certificate agrees, and attestations verify. It also makes PyPI match npm, whose publisher is already registered against `release.yml` (#912) — the two stop being opposites.

`pypi-publish.yml` keeps the build matrix and the artifact uploads. A direct dispatch of it stays useful: as the entry point, both claims name `pypi-publish.yml`, so a TestPyPI publisher registered under that name works for hand-run test uploads.

### Steps

- [ ] Move the `upload-pypi` job into `release.yml` (it needs `id-token: write`, which the caller already grants)
- [ ] Re-register the PyPI publisher for `wingfoil` as `release.yml`
- [ ] Drop `attestations: false`
- [ ] Verify with a `dry-run` build plus one real release

Until then the published files carry no PEP 740 provenance. Nothing else about the release is affected.

PR #924, opened 2026-08-26
ci(pypi): restore release attestations

tunglambk:
## What this changes

`pypi-publish.yml` now builds the distributions and stores them as workflow artifacts when called by `release.yml`, while its direct dispatch keeps the TestPyPI upload and dry-run paths. `release.yml` downloads the built distributions and performs the production upload itself, allowing the default PEP 740 attestations to be published.

The release documentation now records separate trusted publishers: `release.yml` on PyPI and `pypi-publish.yml` on TestPyPI.

## Why

PyPI cannot validate attestations from the reusable production upload because the token exchange identifies `pypi-publish.yml` while the Sigstore certificate identifies its caller, `release.yml`. Defining the production upload in the entry-point workflow makes both identities agree. This implements the workflow changes in #916.

## How it was verified

- [x] `cargo fmt --all`
- [x] `cargo build --workspace --all-targets`
- [x] Both changed workflows parse as YAML
- [x] GitHub's `Test (wingfoil)` and `Lint (fmt & clippy)` jobs
- [x] GitHub's cargo audit, pnpm audit, and dependency review jobs

## Notes for the reviewer

Before the next production release, the PyPI trusted publisher must be re-registered as `release.yml` with no environment. TestPyPI should remain registered as `pypi-publish.yml`. After merge, the remaining verification is a direct `dry-run` build followed by one real release.

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -5,24 +5,17 @@ on:
     inputs:
       pypi-target:
         description: >-
-          Where to publish. `dry-run` builds every wheel and the sdist and
-          uploads nothing — what to use when changing how the wheels are BUILT,
-          because a version can only be uploaded once, to TestPyPI as much as
-          to PyPI, and a build fix should not have to spend one to prove
-          itself.
+          Whether to upload to TestPyPI. `dry-run` builds every wheel and the
+          sdist and uploads nothing — what to use when changing how the wheels
+          are BUILT, because a version can only be uploaded once and a build
+          fix should not have to spend one to prove itself.
         required: true
         default: 'test'
         type: choice
         options:
           - test
-          - prod
           - dry-run
   workflow_call:
-    inputs:
-      pypi-target:
-        description: 'Choose PyPI target (test or prod)'
-        required: true
-        type: string
 
 # Read-only by default. The upload job re-declares `id-token: write` for
 # itself; nothing else in this workflow needs a token of any kind.
@@ -287,54 +280,24 @@ jobs:
           name: dist-windows-x64
           path: dist
 
-  upload-pypi:
-    name: Upload to PyPI
+  upload-testpypi:
+    name: Upload to TestPyPI
     runs-on: ubuntu-latest
     timeout-minutes: 30
     needs: [sdist, linux, macos, windows]
-    # `dry-run` stops here: everything above has already run, so the wheels are
-    # proven built (and downloadable as run artifacts) without spending the
-    # version. release.yml always passes `prod`, so a release can never take
-    # this branch by accident.
-    if: ${{ inputs.pypi-target != 'dry-run' }}
+    # `pypi-target` is declared only under `workflow_dispatch`, so a non-empty
+    # value proves this file is the entry point. Under `workflow_call`,
+    # `github.event_name` reflects the caller and cannot distinguish this path.
+    if: ${{ inputs.pypi-target == 'test' }}
     # Trusted publishing (OIDC): the short-lived token is minted from this
-    # job's identity, so there is no long-lived `*_PYPI_API_TOKEN` secret to
-    # leak or rotate. Matches npm-publish.yml. `id-token: write` must ALSO be
-    # granted by the caller when this runs via `workflow_call` — reusable
-    # workflow permissions are capped by the caller — which release.yml does.
-    #
-    # Both PyPI and TestPyPI need a publisher registered for
-    # wingfoil-io/wingfoil + `pypi-publish.yml` (no environment) before this can
-    # succeed — this file's own name, because PyPI matches the workflow that
-    # *runs* the publish. npm does not: it matches the entry point, so its
-    # publisher is registered as `release.yml`. Do not "make them consistent".
+    # job's identity, so there is no long-lived API token to leak or rotate.
+    # TestPyPI needs a publisher registered for wingfoil-io/wingfoil +
+    # `pypi-publish.yml` (no environment). The production PyPI publisher is
+    # `release.yml`, where the production upload job now lives.
     permissions:
       contents: read
       id-token: write
     steps:
-      # Read `inputs.pypi-target`, NOT `github.event.inputs.pypi-target`. The
-      # `inputs` context carries the value for both triggers (workflow_dispatch
-      # *and* workflow_call). `github.event.inputs` is only populated for a
-      # direct workflow_dispatch — on a workflow_call it reflects the *caller's*
-      # triggering event, which for release.yml is a dispatch with no inputs, so
-      # the value was empty and this step silently fell through to TestPyPI while
-      # still reporting success. Fail loudly on anything unexpected instead of
-      # defaulting to test, so a misrouted release can never masquerade as green.
-      - name: Set PyPI target
-        run: |
-          case "${{ inputs.pypi-target }}" in
-            prod)
-              echo "REPO_URL=https://upload.pypi.org/legacy/" >> $GITHUB_ENV
-              ;;
-            test)
-              echo "REPO_URL=https://test.pypi.org/legacy/" >> $GITHUB_ENV
-              ;;
-            *)
-              echo "::error::pypi-target must be 'test' or 'prod', got '${{ inputs.pypi-target }}'"
-              exit 1
-              ;;
-          esac
-
       # Needed by the name/version check below, which reads the version from
       # the manifest rather than trusting the filenames it is checking.
       - name: Checkout repository
@@ -407,26 +370,14 @@ jobs:
           done
           echo "every file is ${expected_name} ${expected_version}"
 
-      # `attestations: false` is forced by this workflow being *reusable*, and
-      # it is the one thing about that structure PyPI genuinely cannot do.
-      # Two different claims name the workflow, and they disagree here:
-      #
-      #   token exchange  job_workflow_ref  -> pypi-publish.yml  (this file)
-      #   attestation     cert Build Config -> release.yml       (the caller)
-      #
-      # The publisher is registered as `pypi-publish.yml`, so auth succeeds and
-      # the upload is then rejected 400 — "Certificate's Build Config URI does
-      # not match expected Trusted Publisher". Registering `release.yml`
-      # instead just moves the failure to the token exchange
-      # (`invalid-publisher`). No single registration satisfies both, so the
-      # files publish without PEP 740 provenance until the upload moves into
-      # release.yml, where both claims name one workflow (#916).
-      - name: Upload to PyPI
+      # This job only runs when pypi-publish.yml is the entry point. Both the
+      # trusted-publisher lookup and the attestation therefore name this file,
+      # which is the publisher registered on TestPyPI.
+      - name: Upload to TestPyPI
         uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
         with:
           packages-dir: dist
-          repository-url: ${{ env.REPO_URL }}
-          attestations: false
+          repository-url: https://test.pypi.org/legacy/
           # An upload is not atomic: PyPI takes the files one at a time, so a
           # rejection partway through leaves the ones before it published and
           # nothing can take them back. 9.0.0's five wheels landed and its sdist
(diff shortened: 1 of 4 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -287,54 +280,25 @@ jobs:
...
-    # `dry-run` stops here: everything above has already run, so the wheels are
-    # proven built (and downloadable as run artifacts) without spending the
-    # version. release.yml always passes `prod`, so a release can never take
-    # this branch by accident.
-    if: ${{ inputs.pypi-target != 'dry-run' }}
+    # The reusable path only builds artifacts; release.yml owns the production
+    # upload so its token and attestation identify the same workflow. A direct
+    # dispatch can still exercise the complete path against TestPyPI. `dry-run`
+    # stops here after proving every artifact builds.
+    if: ${{ github.event_name == 'workflow_dispatch' && inputs.pypi-target == 'test' }}
```
0-jake-0: **Blocking-ish: the first half of this guard doesn't mean what the comment says it means.**

`github.event_name` inside a reusable workflow is the **caller's** event, not this file's trigger. `release.yml` is `workflow_dispatch`, so during a production release `github.event_name == 'workflow_dispatch'` evaluates **true** here. The only thing stopping a TestPyPI upload on every release is that `pypi-target` is no longer declared under `workflow_call`, so `inputs.pypi-target` is empty and the second clause is false.

That is exactly the class of bug the comment block this PR deletes was written to warn about (`inputs` vs `github.event.inputs` under `workflow_call`, silently falling through to TestPyPI while reporting success). Two consequences:

- The comment "This job only runs when `pypi-publish.yml` is the entry point" is not what the condition tests, so the next reader gets false reassurance.
- It's one line away from a real incident: re-add a `pypi-target` input to `workflow_call` (or have `release.yml` pass `with: pypi-target: test`) and every production release also uploads to TestPyPI.

Since `workflow_call` declares no inputs, `inputs.pypi-target` is non-empty *only* under direct dispatch, so the exact guard is just:

```yaml
    # `pypi-target` is declared only under `workflow_dispatch`, so a non-empty
    # value is itself the proof that this file is the entry point. Do NOT add
    # `github.event_name == 'workflow_dispatch'` — under `workflow_call` that
    # reflects the *caller's* event (release.yml dispatches), so it is true
    # during a release.
    if: ${{ inputs.pypi-target == 'test' }}
```

Fewer moving parts, and it documents the trap instead of stepping around it.

---

.github/workflows/release.yml
```diff
@@ -117,20 +117,107 @@ jobs:
...
+      # The name check is the other half, and it is not hypothetical: the 9.0.0
+      # release built `wingfoil_python-9.0.0-*` because `[project].name` had
+      # been set to the *crate* name, and every one of those files would have
+      # gone to a brand-new PyPI project while `pip install wingfoil` — what
+      # the READMEs and this workflow's own release notes tell users to run —
+      # stayed on 8.0.0. PyPI has no rename and no delete-and-reupload, so the
+      # cost of noticing afterwards is a burned version at best and a
+      # permanently forked package at worst. Check it here, where the failure
+      # is free.
+      - name: Inspect what is about to be published
```
0-jake-0, line 164: **~60 lines of safety check now exist twice, and the two copies have already diverged in this diff.**

This step is a near-verbatim copy of `Inspect what is about to be published` in `pypi-publish.yml`'s `upload-testpypi`, down to the PEP 427 name-escaping and the Cargo-vs-PEP-440 prerelease carve-out. The comment blocks are already not identical between the two (the trailing "and preflight still refuses a version whose tag exists" clause survives in one and not the other), which is drift on day one.

This is the check that exists because 9.0.0 published `wingfoil_python-*` to a brand-new PyPI project — the one place in the release pipeline where a mistake is permanently uncorrectable. Having its production copy and its rehearsal copy be separately maintained files is the worst place in the repo for copy-paste.

The repo already has `scripts/` for exactly this. Suggest extracting to something like `scripts/check-dists.sh` (arg: the dist dir) and calling it from both jobs:

```yaml
      - name: Inspect what is about to be published
        run: scripts/check-dists.sh dist
```

Both jobs already `actions/checkout` for precisely this step's benefit, so the plumbing is free. It also means the TestPyPI rehearsal keeps validating the *same* code the production upload will run, which is the whole point of a rehearsal — see my note on `docs/RELEASING.md`.

---

docs/RELEASING.md
```diff
@@ -13,35 +13,13 @@ and the tag is cut only once all three have the artifacts.
...
-   `upload-pypi` job's OIDC mint — without burning a version number on the real
-   index. `test` is the default for a manual dispatch, so this is also what you
-   get if you dispatch that workflow without thinking; only `release.yml` passes
-   `prod`. Do it after any change to that workflow, not just once ever.
+1. **Trusted publishers must be registered on both PyPI *and* TestPyPI.** There is no `*_PYPI_API_TOKEN` secret. For the project `wingfoil`, register owner `wingfoil-io`, repository `wingfoil`, workflow `release.yml`, and **no environment** on PyPI. Register the same values with workflow `pypi-publish.yml` on TestPyPI. The workflows differ because the production upload runs in `release.yml`, while the rehearsal is a direct dispatch of `pypi-publish.yml`; each trusted publisher must name the workflow that contains its upload job.
+2. **Dispatch `pypi-publish.yml` once with `pypi-target: test`.** That is the only way to exercise the whole PyPI path — sdist repair, five wheels, the `upload-testpypi` job's OIDC mint — without burning a version number on the real index. `test` is the default for a manual dispatch. Do it after any change to that workflow, not just once ever.
```
0-jake-0: **This sentence is no longer true after the change it documents, and that's the main risk in the PR.**

"That is the only way to exercise the whole PyPI path" was accurate when prod and test were the *same* `upload-pypi` job differing only by `repository-url`. After this PR, a `pypi-target: test` dispatch exercises `pypi-publish.yml`'s `upload-testpypi` job; the job that actually publishes to PyPI lives in `release.yml` and **is never run by anything except a real release**. Its OIDC mint, its `download-artifact` glob, its inspect step and its attestation emission are all first-executed against a version number that cannot be reclaimed if any of them is wrong.

The PR body says as much ("the remaining verification is a direct `dry-run` build followed by one real release"). Given the failure mode here is a burned version — and 9.0.0 already burned one — I'd rather not merge with that gap open. Two ways to close it, either is fine:

1. Add a `pypi-index` (or `dry-run`) input to `release.yml` so `upload-pypi` can be pointed at `https://test.pypi.org/legacy/` for a rehearsal. It needs `repository-url` to become an expression and the TestPyPI publisher to also be registered for `release.yml`, but then the production job is genuinely exercised before it matters. Note attestations would be exercised too, which is the thing this PR exists to restore.
2. At minimum, reword this step to say plainly what it now covers (build matrix, sdist repair, TestPyPI OIDC) and what it does **not** (the production upload job in `release.yml`), so nobody reads a green rehearsal as proof the release path works.

Two smaller doc notes while here:

- Lines 17, 92, 94 (and 5 lines in `.github/workflows/README.md`) are unwrapped single-line paragraphs. Both files are hard-wrapped at ~80 today — `main` has zero >90-char lines in `workflows/README.md` and 4 in `RELEASING.md`; this PR takes those to 5 and 11. Please re-wrap so the diffs stay line-reviewable.
- Deleting the "Do not 'make them consistent': doing that breaks whichever one you change" warning is right — the replacement rule ("each trusted publisher must name the workflow that contains its upload job") is the correct generalisation and the reasoning behind moving the job checks out. Worth keeping that phrasing prominent, since the next person to see `release.yml` registered on both npm and PyPI will otherwise assume the old asymmetry was a bug.

---

0-jake-0 (commented):
Reviewed the whole diff. **The core diagnosis is right and the fix is the correct one.** Moving the upload into the entry-point workflow does make `job_workflow_ref` (token exchange) and the Sigstore certificate's Build Config URI name one workflow, which is the only way PyPI accepts PEP 740 attestations here. Swapping the registration instead would just move the failure to the token exchange, as the docs you removed already worked out. Splitting the trusted publishers — `release.yml` on PyPI, `pypi-publish.yml` on TestPyPI — follows from that and is stated correctly.

Mechanically it also checks out:

- All artifacts uploaded by the reusable build jobs are `dist-*` (`dist-sdist`, `dist-linux-*`, `dist-macos-*`, `dist-windows-x64`), so the `pattern: dist-*` + `merge-multiple: true` download matches all six. Reusable-workflow jobs share the caller's run, so same-run artifact scope applies and no `actions: read` is needed.
- `workflow_call:` with no inputs is valid, and `inputs.pypi-target` is referenced in exactly one place, so the build jobs are unaffected by its removal.
- `tag`'s `needs` and `!= 'failure'` gates were correctly widened to both `build-pypi` and `upload-pypi`; a `registries: npm` recovery run still skips through to the tag.
- `actions/checkout@v7` matches the other 36 checkouts in the tree.
- No stale `publish-pypi` / `pypi-target: prod` references left in `release.yml`, and the README/RELEASING references to `upload-testpypi` vs `upload-pypi` line up with where the jobs now live.
- Merges cleanly onto `main` today. Heads up that it conflicts with #926 in `.github/workflows/README.md`, so whichever lands second needs a rebase.

Three things I'd want addressed before merge, inline:

1. **`upload-testpypi`'s `if:` guard** — `github.event_name` is the *caller's* event inside a reusable workflow, so that clause is true during a release. Only the undeclared-input clause is actually doing the work, and the comment claims something the condition doesn't test. One-line fix.
2. **The production upload job is now unexercised until a real release.** The TestPyPI rehearsal no longer runs the same job as production, so the release path's OIDC mint, artifact download and attestation emission are all first-run against a version that can't be reclaimed. This is the one I'd actually hold on.
3. **The ~60-line dist inspection is duplicated** across the two workflows and has already drifted in this diff. Extract to `scripts/`.

Plus a wrapping nit: several rewritten paragraphs in `RELEASING.md` and `workflows/README.md` are now single long lines against otherwise hard-wrapped files.

None of these are objections to the approach — it's the right change, and the reasoning in the PR body is sound.

---

0-jake-0:
Softening my earlier position, having thought harder about the actual blast radius. I said I'd rather not merge with the rehearsal gap open. That was overcautious — **nothing in this diff blocks merge.**

The reasoning I skipped: what happens when the untested production upload job fails on the next release?

A trusted-publisher or attestation mismatch is a *configuration-level* failure, so it is uniform across files — it rejects the very first one. `gh-action-pypi-publish` uploads sequentially, so nothing gets published, and `tag` gates on `needs.upload-pypi.result != 'failure'`, so no tag is cut either. That is the documented partial-publish state, recoverable by re-dispatching with `registries: pypi`. Cost: one failed release run. No burned version.

The genuinely unrecoverable case is a *partial* upload, and that needs a per-file failure — 9.0.0's sdist, where `[project].name` produced `wingfoil_python-*`. That class is exactly what the inspect step catches before a single byte goes up, and this PR keeps that check on both paths. So the scenario I was protecting against is already covered by the thing the PR preserves.

So the gate on this PR is operational, not code. Before the next release, in the PyPI console:

1. Register the production publisher for `wingfoil` as `wingfoil-io` / `wingfoil` / **`release.yml`** / no environment.
2. Confirm TestPyPI stays registered as **`pypi-publish.yml`**.
3. **Delete the old production `pypi-publish.yml` publisher.** This PR removes the `prod` option, so that workflow can no longer reach production — leaving the registration live is a publish credential with nothing left to authorise.

If step 1 is missed, the next release fails at the token exchange with `invalid-publisher`, before uploading anything. Fails fast, no tag, recoverable.

Of my three review comments, only one is worth doing in this PR: the `if:` guard on `upload-testpypi`, which is a one-liner. The duplicated inspect step and the markdown re-wrap are both fine as follow-ups — worth filing, not worth another round trip here.

Nice piece of work. The `job_workflow_ref` vs Sigstore cert analysis is exactly right, and moving the job rather than swapping the registration is the only fix that actually resolves it.

---

tunglambk:
Rebased onto current `main` and resolved the workflow conflict using the SHA-pinned publish action now present upstream. I also changed the TestPyPI guard to rely only on `inputs.pypi-target == 'test'`, as requested, and restored the existing Markdown wrapping. The PR is mergeable again and CI is running.

PR #926, opened 2026-08-27
ci: restore current-engine coverage

### LINKED ISSUE #905: Coverage reporting has been dead since the cutover — restore it, scoped to the current engine
## Summary

**No workflow has uploaded a coverage report since 2026-08-14.** There are zero `codecov/codecov-action` steps anywhere in `.github/`. The badge in `README.md:3` is serving the last report Codecov ever received — which measured the *legacy* engine, in a tree that no longer exists.

## How it broke

The only job that ever produced coverage was `test-legacy: "Test (legacy) & Coverage"` in `rust-test.yml`:

```yaml
run: cargo llvm-cov nextest --manifest-path legacy/wingfoil/Cargo.toml --features full --lib --tests --no-capture --lcov --output-path lcov.info
...
- name: Upload coverage to Codecov
  uses: codecov/codecov-action@v5
```

It was deleted in f1a9ffa — "Delete the legacy tree — the final cutover" (#818), step 4, along with the other thirteen `legacy-*` jobs. That was correct as far as it went: the job pointed at `legacy/`, so it could not survive the tree it measured. But no replacement was wired for `crates/wingfoil`, so **coverage was never measured for the current engine at all — not before the cutover, and not since.**

Two artifacts of the old setup are still in the repo and are now misleading:

- `README.md:3` — the badge, showing a stale legacy figure.
- `codecov.yml` — including `flag_management.default_rules.carryforward: true`, machinery for merging per-leg uploads that no leg has ever used (no upload has ever carried a flag).

Original coverage work was #9 (closed).

## Why a naive restore still reports low

Restoring the job as-is against `-p wingfoil` would measure only what the `Test (wingfoil)` leg runs, and that leg is deliberately narrow. Everything below is currently outside any instrumented run:

| Not measured | Scale |
| --- | --- |
| `*_integration` test binaries, excluded by `-E 'not binary(/_integration$/)'` | 15 test files |
| The 13 integration workflows (aeron, etcd, fix, fluvio, iceoryx2, kafka, kdb, otlp, postgres, prometheus, redis, web, zmq) — each runs `cargo test` against a real service | 21 `*-integration-test` features |
| Doctests (`cargo test --doc -p wingfoil --all-features`) — `llvm-cov` excludes doctests by default | 35 fences, 17 executed |
| `wingfoil-derive` tests | own step, never named by a coverage run |
| `wingfoil-python` (350 pytest + rust tests), `wingfoil-wasm`, `js/` | separate workflows, uninstrumented |

This matters most for adapters. `crates/wingfoil/src/adapters/` is **24,535 of 42,461 lines — 58% of the crate's source** — and only 26 of its 42 files carry an inline `#[cfg(test)]` module. The bulk of adapter behaviour is proven by exactly the integration binaries a restored job would exclude, so the ceiling on a naive restore is roughly "the 42% that isn't adapters, minus doctests".

## Proposed work

1. **Base coverage job** in `rust-test.yml`: `cargo llvm-cov nextest -p wingfoil --all-features --lib --tests -E 'not binary(/_integration$/)' --lcov`, uploaded under flag `unit`. Keep the examples exclusion — they run no tests, and linking ~69 example binaries under `-C instrument-coverage` is what exhausted the runner's disk previously (lld SIGBUS). Its own `shared-key` cache: coverage builds into `target/llvm-cov-target` and shares artifacts with nothing else.
2. **Flagged integration uploads**: each of the 13 integration workflows switches its `cargo test` to an instrumented run and uploads under its own flag (`kafka`, `redis`, `kdb`, …). `carryforward: true` is already configured, so a PR that doesn't trigger the kdb leg keeps kdb's last coverage instead of showing a cliff. This is the piece that makes the adapter 58% visible.
3. **Remaining legs** (separable, lower value): `--doc` coverage needs its own invocation; `wingfoil-derive` is cheap to add; `wingfoil-python` would need `pytest-cov` plus an instrumented `maturin develop`.

Keep `informational: true` on both Codecov statuses while the number settles — the point of this issue is to get an honest figure, not to gate merges on one yet.

Items 1 and 2 are the pair worth doing together; 3 can follow.

maxi-maxima:
## What this changes

Adds a dedicated current-engine unit coverage job and instruments the Rust tests in all 13 service-backed integration workflows. Each integration exports one LCOV report under a stable adapter flag; the workflow guide now documents the split and Codecov carryforward behavior.

## Why

Coverage uploads stopped when the legacy tree and its only coverage job were deleted, leaving the README badge pinned to an obsolete engine. The replacement must measure both the current core engine and the adapter behavior that only runs against real services.

Closes #905.

## How it was verified

- [ ] `cargo lint` and `cargo lint-all` — both stop before Wingfoil code because this Windows host has no MSVC `link.exe`
- [ ] `cargo test -p wingfoil --all-features` and `cargo test -p wingfoil-derive` — same pre-crate linker blocker
- [x] actionlint 1.7.12 on all 14 changed workflow files
- [x] Deterministic contract check: unit plus 13 unique adapter flags, each with toolchain support, cleanup, instrumented test invocation, one LCOV export, and one Codecov upload
- [x] `git diff --check`

## Notes for the reviewer

The unit leg retains the existing exclusion for service-backed `*_integration` binaries. Integration workflows use cargo-llvm-cov's documented `--no-report` / final `report` flow, which also combines the three ZMQ Rust invocations into one flag. Codecov project and patch statuses remain informational, and upload errors remain non-blocking while the new baseline settles.

### DIFF
--- a/.github/workflows/kafka-integration.yml
+++ b/.github/workflows/kafka-integration.yml
@@ -35,11 +35,19 @@ jobs:
 
       - name: Install Rust Toolchain
         uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
 
       - name: Cache Rust Build Artifacts
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: integration
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Prepare coverage collection
+        run: cargo llvm-cov clean --workspace
 
       - name: Install protoc
         run: sudo apt-get install -y protobuf-compiler
@@ -58,10 +66,22 @@ jobs:
 
       - name: Run kafka integration tests
         run: |
-          cargo test --features kafka-integration-test -p wingfoil \
+          cargo llvm-cov --no-report --features kafka-integration-test -p wingfoil \
             -- --test-threads=1 --nocapture
         env:
           RUST_LOG: INFO
+
+      - name: Export Kafka coverage
+        run: cargo llvm-cov report --lcov --output-path lcov.info
+
+      - name: Upload Kafka coverage to Codecov
+        uses: codecov/codecov-action@v5
+        continue-on-error: true
+        with:
+          files: lcov.info
+          flags: kafka
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
 
       # --- Python bindings (wingfoil-python) ---
       # The Rust tests above use testcontainers (per-test container); the Python
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -58,10 +58,18 @@ defaults:
 jobs:
   coverage:
     name: Coverage (unit)
-    # Skips the release.bump commit — see the note above `jobs:`.
+    # Coverage runs only on a real merge to `main`, never on a PR.
+    # Instrumentation costs 7.6x on test execution for the same suite, and
+    # both Codecov statuses are informational. Keeping this off PRs preserves
+    # the fast uninstrumented gate while tracking coverage on every main push.
+    #
+    # This is an allow-list because `workflow_call` inherits the caller's
+    # dispatch event and must not add coverage to the release critical path.
+    # Also skips the release.bump commit, as the other legs do.
     if: >-
-      ${{ github.event_name != 'push'
-      || !(startsWith(github.event.head_commit.message, 'bump: ')
+      ${{ github.event_name == 'push'
+      && github.ref == 'refs/heads/main'
+      && !(startsWith(github.event.head_commit.message, 'bump: ')
       && contains(github.event.head_commit.message, ' version to ')) }}
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -139,10 +147,10 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev
 
       - name: Run wingfoil tests
-        # The coverage job is scoped to `-p wingfoil`, so the wingfoil crate's
-        # tests — the adapter parity suites (lines/csv/augurs) and the trybuild
-        # cases — would otherwise never execute in CI. Run them here with all
-        # features so the feature-gated csv/augurs adapters are exercised.
+        # This is the fast uninstrumented PR gate. The main-only coverage job
+        # deliberately runs the same wingfoil selection with LLVM coverage.
+        # Run with all features so the feature-gated csv/augurs adapters are
+        # exercised.
         # `--lib --tests` skips the example binaries (already compile-checked by
         # the workspace clippy job).
         #
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -56,6 +56,60 @@ defaults:
 # workflow, which reaches these legs through test.all.
 
 jobs:
+  coverage:
+    name: Coverage (unit)
+    # Skips the release.bump commit — see the note above `jobs:`.
+    if: >-
+      ${{ github.event_name != 'push'
+      || !(startsWith(github.event.head_commit.message, 'bump: ')
+      && contains(github.event.head_commit.message, ' version to ')) }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install Rust Toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
+
+      - name: Cache Rust Build Artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: rust-coverage
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev
+
+      - name: Run current-engine tests with coverage
+        # Keep the service-backed integration binaries out of this leg: each is
+        # measured by its own workflow and uploaded under an adapter flag.
+        # Examples are compile-checked by clippy and contain no tests, so
+        # instrumenting their wide dependency graph would only consume disk.
+        run: |
+          cargo llvm-cov nextest -p wingfoil --all-features --lib --tests \
+            -E 'not binary(/_integration$/)' \
+            --lcov --output-path lcov.info
+        env:
+          RUST_LOG: INFO
+
+      - name: Upload unit coverage to Codecov
+        uses: codecov/codecov-action@v5
+        continue-on-error: true
+        with:
+          files: lcov.info
+          flags: unit
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
+
   test:
     name: Test (wingfoil)
     # Skips the release.bump commit — see the note above `jobs:`.
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -2,9 +2,10 @@
 
 ## CI (run on push / PR)
 
-* `rust-test.yml` — two parallel jobs: `Test (wingfoil)` and
-  `Lint (fmt & clippy)`. They were one serial job until they were split; the
-  legs share no build artifacts, so serialising them bought nothing.
+* `rust-test.yml` — three parallel jobs: `Coverage (unit)`, `Test (wingfoil)`
+  and `Lint (fmt & clippy)`. The coverage leg measures the current engine with
+  service-backed integration binaries excluded; those binaries are measured
+  by their dedicated workflows below.
 * `python-test.yml` — Python (`wingfoil-python`) build + pytest with coverage.
 * `security-audit.yml` — fails on dependencies with known advisories
   (`cargo audit` for Cargo, `pnpm audit` for `wingfoil-js`, and
@@ -60,7 +61,11 @@ The per-adapter integration workflows above are the *only* place their
 `tests/*_integration.rs` binaries are executed. `rust-test.yml` compiles them
 (so they stay type- and link-checked) but filters them out of its test run:
 without the service each one needs, they only exercise connection-timeout
-paths, and they are slow doing it.
+paths, and they are slow doing it. Each workflow instruments its Rust tests
+with `cargo llvm-cov` and uploads an LCOV report under a stable adapter flag;
+`codecov.yml` carries forward flags for integrations a narrower change does
+not trigger. Codecov project and patch statuses remain informational while the
+current-engine baseline settles.
 
 Every push/PR workflow declares a `concurrency` group so a superseded PR push
 cancels its predecessor. The group name is a literal per file rather than
--- a/.github/workflows/aeron-integration.yml
+++ b/.github/workflows/aeron-integration.yml
@@ -36,11 +36,19 @@ jobs:
 
       - name: Install Rust Toolchain
         uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
 
       - name: Cache Rust Build Artifacts
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: integration
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Prepare coverage collection
+        run: cargo llvm-cov clean --workspace
 
       - name: Install protoc
         run: sudo apt-get install -y protobuf-compiler
@@ -66,10 +74,22 @@ jobs:
 
       - name: Run aeron integration tests
         run: |
-          cargo test --features aeron-integration-test -p wingfoil \
+          cargo llvm-cov --no-report --features aeron-integration-test -p wingfoil \
             -- --test-threads=1 --nocapture
         env:
           RUST_LOG: INFO
+
+      - name: Export Aeron coverage
+        run: cargo llvm-cov report --lcov --output-path lcov.info
+
+      - name: Upload Aeron coverage to Codecov
+        uses: codecov/codecov-action@v5
+        continue-on-error: true
+        with:
+          files: lcov.info
+          flags: aeron
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
 
       # --- Python bindings (wingfoil-python) ---
       # The Rust tests drive the driver container through testcontainers; the
--- a/.github/workflows/etcd-integration.yml
+++ b/.github/workflows/etcd-integration.yml
@@ -35,11 +35,19 @@ jobs:
 
       - name: Install Rust Toolchain
         uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
 
       - name: Cache Rust Build Artifacts
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: integration
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Prepare coverage collection
+        run: cargo llvm-cov clean --workspace
 
       - name: Install protoc
         run: sudo apt-get install -y protobuf-compiler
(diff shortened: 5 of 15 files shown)


.github/workflows/rust-test.yml
```diff
@@ -56,6 +56,60 @@ defaults:
...
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev
+
+      - name: Run current-engine tests with coverage
+        # Keep the service-backed integration binaries out of this leg: each is
+        # measured by its own workflow and uploaded under an adapter flag.
+        # Examples are compile-checked by clippy and contain no tests, so
+        # instrumenting their wide dependency graph would only consume disk.
+        run: |
+          cargo llvm-cov nextest -p wingfoil --all-features --lib --tests \
```
0-jake-0, line 106: **This is byte-for-byte the same test selection as `Test (wingfoil)`, so the suite now runs twice per push.**

```
test:      cargo nextest run       -p wingfoil --all-features --lib --tests -E 'not binary(/_integration$/)'
coverage:  cargo llvm-cov nextest  -p wingfoil --all-features --lib --tests -E 'not binary(/_integration$/)' --lcov ...
```

Identical package, features, target selection and filter. On this PR's own run that cost 10m41s for `Coverage (unit)` against 3m25s for `Test (wingfoil)` — and because the two use different `shared-key`s (`rust-coverage` vs `rust-test`) and different RUSTFLAGS, there is no artifact reuse between them. So `rust-test.yml`'s wall clock goes from ~3m40s (bounded by lint) to ~10m40s on every push and PR, for a second execution of tests that already passed in the job next door.

I don't think that's automatically wrong — instrumented builds are slow, and you may well want the *gating* leg to stay fast and uninstrumented. But it's worth being deliberate about rather than inheriting. Options, roughly in order of how much I'd favour them:

1. **Leave it.** Simplest, honest about the tradeoff: coverage is bought with ~7 minutes of parallel CI. Worth a comment on the job saying so, since the duplication otherwise reads like an oversight.
2. **Narrow the coverage leg.** Coverage of `--all-features` is the point, so features can't go, but the leg doesn't need to be the gate — nothing consumes its pass/fail (Codecov statuses are `informational: true`). `continue-on-error: true` on the job would stop a coverage-tool regression from turning the PR red while keeping the upload.
3. **Fold the two together** — one `cargo llvm-cov nextest` run serving both roles. Cheapest in CI minutes, but it puts `cargo-llvm-cov` on the critical path of the gating test job, which is a real reliability cost for a tool that's only reporting.

Worth noting the comment on `Test (wingfoil)`'s run step (lines 90-91 on `main`) already says *"The coverage job is scoped to `-p wingfoil`, so the wingfoil crate's tests would otherwise never execute in CI"* — written when a coverage job existed and was narrower. With this PR the two are exactly co-extensive, so that comment is now misleading in the other direction and should be updated whichever option you pick.

---

.github/workflows/kafka-integration.yml
```diff
@@ -58,11 +66,23 @@ jobs:
 
       - name: Run kafka integration tests
         run: |
-          cargo test --features kafka-integration-test -p wingfoil \
+          cargo llvm-cov --no-report --features kafka-integration-test -p wingfoil \
             -- --test-threads=1 --nocapture
         env:
           RUST_LOG: INFO
 
+      - name: Export Kafka coverage
```
0-jake-0, line 74: Non-blocking, but it applies to all 13 integration workflows so it's worth deciding once.

The export and upload steps have no `if:`, so a failing test step skips both and the run uploads nothing under this flag. Combined with `carryforward: true` in `codecov.yml` that's benign — the previous report carries forward rather than the flag vanishing — so this is a *coverage-freshness* question, not a correctness one.

The argument for `if: always()` (or `success() || failure()`) is that a red integration run is exactly when you want to see which paths *did* execute before the failure. The argument against is that partial coverage from a half-run suite is noise. Either is defensible; right now it reads as unconsidered rather than chosen. A one-line comment on the export step saying which you meant would settle it.

Separate small note on the caching in these files, also non-blocking: all 13 share `shared-key: integration` and none sets `save-if`, so each job writes that cache — and now they write *instrumented* artifacts, while the `maturin develop` steps later in the same jobs build uninstrumented ones into the same target dir. Expect the `integration` cache hit rate to drop for a while after this lands, and the first run of each workflow to be a cold instrumented build. Not a reason to change anything here, just so the post-merge slowdown isn't mistaken for a regression.

---

0-jake-0 (commented):
Went through all 15 files. **This is careful, consistent work and I think it's close to mergeable** — the one thing I'd want a decision on is the duplicated test run, and that's a tradeoff call rather than a defect.

What I verified rather than assumed:

- **All 13 integration workflows are uniform**: `components: llvm-tools-preview` on the toolchain, `taiki-e/install-action@cargo-llvm-cov`, `cargo llvm-cov clean --workspace` before the tests, `--no-report` on every Rust test invocation, one `cargo llvm-cov report --lcov`, one upload. Thirteen distinct flags (`aeron`, `etcd`, `fix`, `fluvio`, `iceoryx2`, `kafka`, `kdb`, `otlp`, `postgres`, `prometheus`, `redis`, `web`, `zmq`), no collisions with `unit`. The `--no-report` / `report` split is the documented flow, and ZMQ's three invocations correctly collapse into one flag — that's the case that would have been easiest to get wrong.
- **Export/upload is inserted after the Rust tests and before each workflow's Python-bindings section** in every file that has one. That's the right seam: `maturin develop` isn't instrumented and would contribute nothing.
- **The `codecov.yml` claim in the workflow README is accurate as-is** — `flag_management.default_rules.carryforward: true` and both statuses `informational: true` are already committed, so no config change is needed and no PR will report a phantom drop when its path filters skip the adapter legs. Good catch not to touch that file.
- **The upload actually works from a fork.** `codecov/patch` reported success on this PR's head, so the token path (or tokenless) is fine and `continue-on-error` isn't masking a broken setup. That was my main worry going in and it's answered by your own CI run.
- **`--all-features` + apt `cmake` in the new `coverage` job** matches what `Test (wingfoil)` already does and passes, so the CMake >= 3.30 warning in `CLAUDE.md` (which bites `aeron-integration.yml`, `python-test.yml` and `pypi-publish.yml`, all of which install 3.31 by hand) doesn't apply here. Worth having checked, since a copy of that step into a job that *did* build aeron would fail at link time with `-lbsd`.
- Merges cleanly onto `main`. It conflicts with #924 in `.github/workflows/README.md`, so whichever lands second needs a rebase.

Two comments inline: one on the coverage leg duplicating `Test (wingfoil)` exactly (the real discussion), one non-blocking note on missing `if: always()` across the export steps plus a heads-up on `integration` cache churn after merge.

One process note: your PR body flags that `cargo lint`, `cargo lint-all` and the test suite couldn't run because the Windows host has no MSVC `link.exe`. For a workflows-only change that's fine — nothing here touches Rust — and CI covered it: `Test (wingfoil)`, `Lint (fmt & clippy)`, `Coverage (unit)` and `Web Integration Tests` are all green on `abaa14a`. Just worth being explicit that the green CI is what discharges that checklist, not the local run.

---

0-jake-0:
Correcting the timing figure in my review comment on `rust-test.yml` — I compared a cold job against a warm one, so the ~10m40s number overstates the steady-state cost.

Per-step timings from this PR's own run ([job 98388175275](https://github.com/wingfoil-io/wingfoil/actions/runs/33032564532/job/98388175275) vs [98388175304](https://github.com/wingfoil-io/wingfoil/actions/runs/33032564532/job/98388175304)):

| | `Test (wingfoil)` | `Coverage (unit)` |
|---|---|---|
| Cache restore | 31s (**warm hit** on `rust-test`) | 0s (**complete miss** — `rust-coverage` has never existed) |
| Test step | 2m19s | 9m51s |

So the coverage leg did a full cold compile of the crate and its dependency graph under `--all-features` *plus* instrumentation, while the test leg restored main's cache and compiled only the delta. Most of that 9m51s gap is first-run compilation, not instrumentation overhead, and it is transient: `save-if: github.ref == 'refs/heads/main'` means the cache gets populated on the first push to `main` after merge, and PR runs restore it from there.

The duplication point stands — the two jobs still run the identical test selection, and the instrumented build is a genuinely separate artifact set that shares nothing with `rust-test` — but the honest framing is "the steady-state cost is unmeasured and materially lower than 10 minutes", not "PR feedback goes from 4 to 11 minutes". Not worth restructuring the job on the strength of the number I first quoted. Worth one merge-then-watch: if the first few post-merge runs land near the test leg, there is nothing to do here at all.

Two other things the logs settle, both good:

- `Token length: 0` — `secrets.CODECOV_TOKEN` is correctly withheld from this fork PR, and the upload succeeded regardless via tokenless public-repo upload. So the token line is inert for forks rather than a leak, and coverage works from fork PRs.
- `codecov/codecov-action@v5` fetches the CLI from `cli.codecov.io` and GPG-verifies it against Codecov's uploader key plus a signed SHA256SUM before running it. Integrity is checked; the only loose end is `CC_VERSION: latest`, i.e. the CLI version floats. Not something I'd block on given every action in this repo is tag-pinned rather than SHA-pinned, but worth knowing it's a runtime download.

---

0-jake-0:
My previous comment was wrong, and in the direction that matters. I claimed "most of that 9m51s gap is first-run compilation, not instrumentation overhead, and it is transient." The logs say the opposite. Retracting it.

Both jobs ran the **identical** workload — `Starting 958 tests across 73 binaries (15 binaries skipped)`, 958 passed, in both. Splitting each run step:

| | `Test (wingfoil)` | `Coverage (unit)` |
|---|---|---|
| Crates compiled | 4 (warm cache) | 611 (cold cache) |
| Compile | 1m 28s | 3m 43s |
| **Test execution** | **47.4s** | **358.2s** |
| Report | — | 3s |
| Run step total | 2m 19s | 9m 51s |

So of the 9m51s, compilation is 3m43s and **test execution is 5m58s — a 7.6× slowdown on the same 958 tests.** That is LLVM instrumentation overhead at runtime. It is cache-independent and permanent.

Warming the cache only attacks the 3m43s, and not all of it — a warm coverage run still compiles the workspace crates instrumented. Steady state comes out around **9m15s–10m**, against 10m41s cold. Cache warming buys roughly a minute, not six.

Which means my *original* framing was the right one: `rust-test.yml`'s critical path really does go from ~3m40s (bounded by lint at 3m38s) to ~9-10 minutes, on every push and every PR.

This is not a defect in the PR — it is what source-based coverage costs on this codebase, and arguably wingfoil is close to the worst case for it: a tight-loop stream engine where the hot paths that make the engine fast are exactly the regions that get counter increments.

**Suggestion, and I think it resolves the whole thread: run the coverage leg on `push` only, not on `pull_request`.**

#905 asks for an honest coverage number and explicitly says to keep both Codecov statuses `informational: true` "while the number settles" — so coverage is not gating merges and does not need to be on the PR critical path. Making the job `if: github.event_name != 'pull_request'` (or moving it to a schedule) would:

- return PR feedback to ~3m40s,
- still track coverage on every commit that lands on `main`,
- warm the `rust-coverage` cache on main, which is where `save-if` writes it anyway,
- cost nothing in coverage fidelity, since `carryforward: true` is already configured.

The 13 integration legs I have not measured. Their baseline is ~8 minutes (kafka's recent `main` runs: 7m45s, 8m47s, 8m25s) and they are dominated by container startup and network I/O rather than CPU in instrumented wingfoil code, so I would expect a smaller relative hit than 7.6× — but `llvm-cov clean --workspace` does force a workspace rebuild on every run, so the compile side grows for certain. Treat that as unmeasured.

Apologies for the churn on the number.

---

.github/workflows/rust-test.yml
```diff
@@ -56,6 +56,60 @@ defaults:
 # workflow, which reaches these legs through test.all.
 
 jobs:
+  coverage:
+    name: Coverage (unit)
+    # Skips the release.bump commit — see the note above `jobs:`.
+    if: >-
+      ${{ github.event_name != 'push'
+      || !(startsWith(github.event.head_commit.message, 'bump: ')
+      && contains(github.event.head_commit.message, ' version to ')) }}
```
0-jake-0: Concrete fix for the timing thread — this is the only change we'd like before merge, and it's one `if:`.

```suggestion
    # Coverage runs only on a real merge to `main`, never on a PR.
    # Instrumentation costs 7.6x on test *execution* — 47.4s -> 358.2s for the
    # same 958 tests across 73 binaries — and that is runtime overhead, so no
    # amount of cache warming touches it. Left on `pull_request` it makes this
    # workflow's critical path ~10m instead of the ~3m40s the lint leg sets.
    # Nothing gates on the result: both Codecov statuses are `informational`
    # (codecov.yml), and #905 asks for an honest number rather than a merge gate.
    #
    # Spelled as an allow-list on purpose. `github.event_name != 'pull_request'`
    # would still be true under `workflow_call` — a release reaches this file
    # through all-tests.yml, where the event is the *caller's* dispatch — and
    # would put 10 minutes on the release critical path.
    #
    # Also skips the release bump commit, as the other legs do.
    if: >-
      ${{ github.event_name == 'push'
      && github.ref == 'refs/heads/main'
      && !(startsWith(github.event.head_commit.message, 'bump: ')
      && contains(github.event.head_commit.message, ' version to ')) }}
```

Why this is the whole fix: the 13 integration workflows you instrumented have **no `pull_request` trigger at all** — they are `push` + adapter path filters — so they never sit on PR latency and want no change. `web-integration.yml` is the one exception and is path-filtered to web/wasm/js, narrow enough to leave alone. Every second this PR adds to PR feedback comes from this single job.

Two things this also buys, for free:

- With the job on `main` only, `save-if: github.ref == 'refs/heads/main'` is always true, so every run both restores *and* saves. The cold 3m43s compile becomes a warm workspace-only rebuild, and the job settles around 8-9m rather than 10m41s.
- `Test (wingfoil)` stays uninstrumented on the PR path, so the gating leg never depends on `cargo-llvm-cov` behaving.

What it gives up, which is worth a deliberate nod rather than discovering later: **per-PR patch coverage.** `codecov/patch` needs a report on the PR head, so it stops being meaningful on PRs — you get coverage of `main` over time, not "did this PR add uncovered lines". Both statuses are informational today so nothing breaks. If you want it back selectively, appending `|| contains(github.event.pull_request.labels.*.name, 'coverage')` makes it opt-in per PR.

One last thing while you're in this file: the comment on `Test (wingfoil)`'s run step still says *"The coverage job is scoped to `-p wingfoil`, so the wingfoil crate's tests would otherwise never execute in CI"*. That was written for the deleted legacy coverage job. Worth rewording to say the two legs run the same selection deliberately — one as the fast PR gate, one instrumented on `main` — so the next reader doesn't take the duplication for an oversight.

---

0-jake-0 (commented):
Summarising where this lands, since I've iterated on the timing number a couple of times.

**The change is sound and I'd merge it with one adjustment** — the applyable suggestion on `rust-test.yml` above, gating the coverage job to pushes on `main`. Everything else in the PR is right as it stands, including all 13 integration workflows, which need no change: they have no `pull_request` trigger, so their instrumentation never costs PR feedback time.

What I verified, for the record:

- 13 unique adapter flags, no collision with `unit`; `--no-report` on every instrumented invocation with exactly one `report` per workflow; ZMQ's three runs correctly collapsing into a single flag.
- Export/upload placed after the Rust tests and before each workflow's Python leg — the right seam, since `maturin develop` isn't instrumented.
- `codecov.yml` already carries `flag_management.default_rules.carryforward: true` and `informational: true` on both statuses, so the README's claims hold with no config change and no PR will show a phantom cliff.
- Uploads work from a fork: `codecov/patch` succeeded on `abaa14a` with `Token length: 0`, i.e. tokenless. `secrets.CODECOV_TOKEN` is correctly withheld from fork PRs rather than leaked, and `codecov-action@v5` GPG-verifies the CLI it downloads.
- `--all-features` with apt `cmake` matches what `Test (wingfoil)` already does and passes, so the >= 3.30 requirement that bites aeron doesn't apply to this job.

Two non-blocking leftovers from my earlier comments, both fine to merge without:

1. The export/upload steps across the 13 workflows have no `if:`, so a failing test step uploads nothing under that flag. `carryforward` makes that benign; it just reads as unconsidered. A one-line comment saying which behaviour you meant would settle it.
2. Expect the shared `integration` cache to churn for a while post-merge — it now mixes instrumented artifacts with the uninstrumented ones the `maturin develop` steps build into the same target dir. Not a reason to change anything, just so the first slow runs aren't mistaken for a regression.

Thanks for this — the flag-per-adapter structure is exactly what #905 asked for, and getting the ZMQ triple-invocation down to one flag was the fiddly part.

---

.github/workflows/rust-test.yml
maxi-maxima: Applied in `572de17`: `Coverage (unit)` now runs only for non-bump pushes to `refs/heads/main`, so it stays off `pull_request`, `workflow_call`, `workflow_dispatch`, and release paths. I also updated the `Test (wingfoil)` comment to document the deliberate split between the fast PR gate and main-only instrumented run.

Validation: `actionlint 1.7.12 .github/workflows/rust-test.yml` and `git diff --check` both pass.

Issue #940, opened 2026-08-29
`setup-dev.sh`: install the native toolchain an `--all-features` build needs, not just protoc


0-jake-0:
`scripts/setup-dev.sh` installs `protoc` and nothing else. But anyone who runs `cargo lint-all`, `cargo test -p wingfoil --all-features`, or touches the Aeron adapter also needs clang, `libclang-dev`, `uuid-dev`, `libbsd-dev` and **CMake ≥ 3.30**.

Those instructions currently exist only as a copy-paste block in `CLAUDE.md` ("Aeron adapter system dependencies"), which means a contributor typically discovers them the expensive way: a long, apparently successful build that ends at link time with

```
rust-lld: error: unable to find library -lbsd
```

`libbsd-dev` is the one that is easy to miss, because nothing fails until then and only on an `--all-features` target. CMake fails differently and just as late — the vendored `rusteron-media-driver` Aeron sources set a 3.30 floor in their own `cmake_minimum_required`, so the 3.28 that ships on many distros aborts the build script with `CMake 3.30 or higher is required`.

### Where to start

Teach `setup-dev.sh` a second mode — a flag or a positional argument, e.g. `scripts/setup-dev.sh --all-features` — that installs the native toolchain on top of protoc. Keep the script's existing shape:

- **Idempotent**, with the same "already installed → skip and print the version" style the protoc path uses.
- Same four package managers already handled: apt / dnf / pacman / brew.
- Default behaviour **unchanged**: a bare `scripts/setup-dev.sh` still installs protoc only. Nothing about working on the engine itself needs any of this.

CMake needs a *version* check, not a presence check — 3.28 is installed and wrong. Fall back to the Kitware installer script that `CLAUDE.md` documents when the packaged version is too old, and print what it is doing rather than silently overwriting a system CMake.

Then collapse the duplication: `CLAUDE.md`'s block and `CONTRIBUTING.md`'s parenthetical ("a few adapters need more…") should point at the flag instead of restating the package list, so there is one place to update when a dependency changes.

### Acceptance criteria

- [ ] Bare `scripts/setup-dev.sh` behaves exactly as it does today.
- [ ] The new mode is idempotent and safe to re-run.
- [ ] On a clean Ubuntu container, running it is enough for `cargo build -p wingfoil --all-features` to link.
- [ ] The CMake path handles "present but too old", not just "absent".
[...]

Issue #937, opened 2026-08-29
Docs: runnable doctests for the join, merge and sample family


0-jake-0:
Third and last of the doctest issues (see the value-transform and rate/window ones). `crates/wingfoil/src/fluent.rs` documents ~90 public methods and only 7 carry a runnable example.

The multi-input combinators are the ones where prose alone genuinely cannot settle the question — *which* input ticks the output, what the passive variant holds, who wins a same-instant tie. Every one of those is a two-line assert.

### Scope

On `StreamOps`, in `crates/wingfoil/src/fluent.rs`:

- `join`, `join3`, `try_join`, `try_join3`
- `join_passive`, `try_join_passive`
- `merge`, `merge_all`
- `sample`, `filter`

**Partial PRs are welcome — one method is a valid PR.**

### Where to start

Copy the `filter_map` example in the doc block immediately above `fn filter_map` in `fluent.rs`.

What each family should actually demonstrate:

- **`join` vs `join_passive`** — the whole point is the difference. `join` ticks when *either* input ticks; `join_passive` ticks only with the active one and reads the other's held value. One example each, wired over the same two streams, makes that unambiguous in a way the current one-liner cannot.
- **`merge` / `merge_all`** — the tie rule. Same-instant values are not coalesced; the doc block above `merge` already argues the earliest-supplied-wins ordering, so the example should wire two tickers that collide and assert it.
- **`try_*`** — show that `Err` aborts the run with the error as context, rather than silently dropping a tick. `Ok`/`Err` and emit/don't-emit are not interchangeable.
- **`sample`** — the trigger drives; the sampled stream is read, not awaited.

Rules:

1. **Deterministic.** `RunMode::HistoricalFrom(NanoTime::ZERO)` + `RunFor::Cycles(n)`.
2. **Assert exact values and tick times** — `with_time().accumulate()` is the right instrument for anything about *which* input ticked.
3. `.unwrap()` is allowed in doc examples (`CLAUDE.md`, Error Handling).
4. 10–14 lines each.

### Acceptance criteria

- [ ] Each method listed has a fenced, runnable example.
- [ ] The `join` / `join_passive` pair and the `merge` tie rule are each demonstrated by an assert, not only by prose.
- [ ] `cargo test --doc -p wingfoil --all-features` passes.
- [ ] `cargo fmt --all` and `cargo lint` clean.

[...]

Issue #936, opened 2026-08-29
Docs: runnable doctests for the rate-limiting, window and ordering combinators


0-jake-0:
Companion to the value-transform doctest issue. `crates/wingfoil/src/fluent.rs` documents ~90 public methods and only 7 carry a runnable example, so most of the combinator reference on docs.rs is a sentence and a signature.

These particular combinators are the ones most often *misread* from a one-line description — leading vs. trailing edge, whether the value is dropped or held, whether the node ticks or stays silent — so an example is worth more here than anywhere else.

### Scope

On `StreamOps`, in `crates/wingfoil/src/fluent.rs`:

- `audit`, `window`, `buffer`
- `limit`, `skip`, `step_by`
- `skip_while`, `take_while`
- `delay`

**`throttle` is deliberately not on this list** — its contract wording is #929, and the example belongs in that PR alongside the prose.

**Partial PRs are welcome — one method is a valid PR.**

### Where to start

Copy the `filter_map` example in the doc block immediately above `fn filter_map` in `fluent.rs`; it is the house shape.

Two of these need more care than a copy:

- **`delay`** is the one that demonstrates `Tick::Silent` — the value slot updates without ticking downstream. Assert with `with_time().accumulate()` so the example pins *when* the value re-emerges, not just that it does. Note the `T: PartialEq` bound and why it is there: `delay` rides the `TimeQueue`, which deduplicates `(value, time)` pairs on equality (see the `TimeQueue` section of `CLAUDE.md`).
- **`audit` vs `throttle`** — `audit` emits the latest value at the end of each window, `throttle` passes the first and drops the rest. Say so in the example's comment; that distinction is the reason both exist.

Rules for every example here:

1. **Deterministic.** `RunMode::HistoricalFrom(NanoTime::ZERO)` + `RunFor::Cycles(n)`, never `RealTime`. Time-based ops under `HistoricalFrom` are exactly reproducible because engine time is source-driven.
2. **Assert exact values and tick times** — `with_time().accumulate()` for anything whose point is *when* it ticks.
3. `.unwrap()` is allowed in doc examples (`CLAUDE.md`, Error Handling).
4. 10–14 lines.

### Acceptance criteria

- [ ] Each method listed has a fenced, runnable example asserting values (and tick times where timing is the point).
- [ ] `cargo test --doc -p wingfoil --all-features` passes.
- [ ] `cargo fmt --all` and `cargo lint` clean.

[...]

Issue #935, opened 2026-08-29
Docs: runnable doctests for the value-transform combinators


0-jake-0:
`crates/wingfoil/src/fluent.rs` documents ~90 public methods and only **7** of them carry a runnable example — `grep -c '/// ```' crates/wingfoil/src/fluent.rs` returns 14 fence lines. docs.rs is where a new user lands first, and for almost every combinator it currently shows a one-line sentence and a signature.

This is the cheapest possible improvement to the front door, and it is already gated: `cargo test --doc -p wingfoil --all-features` runs every fence, so a wrong example fails CI rather than rotting.

### Scope

The value-transform group on `StreamOps`, in `crates/wingfoil/src/fluent.rs`:

- `map`, `try_map`, `map_filter`
- `with_time`, `ticked_at`, `ticked_at_elapsed`
- `difference`, `pairwise`, `enumerate`

Sibling issues cover the rate/window group and the join/merge group. **Partial PRs are welcome — one method is a valid PR.**

### Where to start

Copy the existing `filter_map` example, in the doc block immediately above `fn filter_map` in `fluent.rs`. It is the house shape and it is 14 lines: build a graph, drive it from `ticker(..).count()`, `accumulate()`, run, assert the whole sequence.

Rules that apply to every one of these:

1. **Deterministic or it does not go in.** `RunMode::HistoricalFrom(NanoTime::ZERO)` with `RunFor::Cycles(n)`. Never `RealTime` — engine time under `HistoricalFrom` consults no clock at all, which is what makes the assert stable.
2. **Assert exact values.** Prefer `accumulate()` plus `r.value(&s)` so the whole tick sequence is pinned in one assert; use `with_time().accumulate()` where the *timing* is the thing being taught (`ticked_at`, `ticked_at_elapsed`).
3. `.unwrap()` is allowed here — doc examples are one of the two exemptions in the Error Handling section of `CLAUDE.md`.
4. Keep them short. 10–14 lines. An example that needs a paragraph of setup is teaching the wrong thing.

### Acceptance criteria

- [ ] Each method listed above has a fenced, runnable example that asserts a value.
- [ ] `cargo test --doc -p wingfoil --all-features` passes, with the doctest count up by the number of methods covered.
- [ ] `cargo fmt --all` and `cargo lint` clean.

[...]

PR #945, opened 2026-08-30
`check-example-docs.sh` should assert the example count the READMEs quote (all three numbers currently disagree)

### LINKED ISSUE #939: `check-example-docs.sh` should assert the example count the READMEs quote (all three numbers currently disagree)
Three places state how many examples this repo has, and no two of them agree:

| Source | Says |
|---|---|
| `README.md:234` | "45 runnable examples, each in its own directory" |
| `crates/wingfoil/README.md:56` | "~40 runnable examples" |
| `ls -d crates/wingfoil/examples/*/*/ \| wc -l` | **38** directories |
| `grep -c '^\[\[example\]\]' crates/wingfoil/Cargo.toml` | **47** targets |

(The last two legitimately differ: `showcase/trading_e2e/` declares two binaries, `trading_e2e_ws_server` and `trading_e2e_fix_gw`.)

Nothing checks any of this, so it drifted once and will drift again on the next example that lands.

### Where to start

`scripts/check-example-docs.sh` already does the hard half: it parses every `[[example]]` target out of `crates/wingfoil/Cargo.toml` and maps each to its directory (the "Collect declared example targets" block), precisely so the check cannot drift from what cargo builds. It runs in CI at `.github/workflows/rust-test.yml:257`.

Add a fourth check to it:

1. Compute the canonical count from what it already parsed.
2. Grep the two READMEs for the number they claim.
3. Fail, printing both numbers and the file, when they differ.

**Decide which number is canonical and write the reason into the script's header comment** — distinct example *directories* (what "each in its own directory" claims) or cargo *targets* (what a user can `cargo run --example`). Either is defensible; the point is that the script says which, so the next person does not have to re-derive it. Then correct both READMEs to whatever you chose, and make `crates/wingfoil/README.md` state the number rather than "~40" — an approximation is not checkable.

### Acceptance criteria

- [ ] `scripts/check-example-docs.sh` fails when a README's count is wrong (verify by editing one and running it) and passes on `main` once the READMEs are corrected.
- [ ] Both `README.md` and `crates/wingfoil/README.md` carry the same, correct, exact number.
- [ ] The script's header comment states which count is canonical and why.
- [ ] No CI change needed — the workflow already invokes the script.

No Rust required for this one. Branch from `main`, target base `main`. See [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md).

YuuGR1337:
Fixes #939

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ sources and sinks in a line.
 > [release notes](docs/release-notes/9.0.0.md) and the
 > [migration guide](docs/migration.md).
 
+There are 47 runnable example targets (38 example directories) under
+[`crates/wingfoil/examples/`](crates/wingfoil/examples/) — this count is
+asserted in CI by `scripts/check-example-docs.sh`, so it cannot drift.
+
 
 ## Languages
 


0-jake-0:
@YuuGR1337 — thanks for this, and apologies for the mix-up on our side.

Looking at it again after merging, I think this branch got pushed before it was finished: it has the README sentence but not the change to `scripts/check-example-docs.sh` that #939 was really asking for. That's an easy thing to do, and it should have been caught in review — I merged it without checking the diff against the issue's acceptance criteria. That one's on me, not you.

For the record, so nothing here reads as a mark against your work: the two numbers you wrote were the ones the repo could give you at the time, and one of them was quietly wrong for a reason nobody had noticed — the "38 directories" figure counts `[[bench]]` paths alongside `[[example]]` ones. The manifest actually declares 43. You surfaced that by writing the number down at all.

I've reopened #939 and picked up the script half myself in #947, since #946 landed a 48th example in the meantime and the stale count was already spreading. The check now parses the numbers back out of the README sentence and fails when they disagree with the manifest — the thing this issue wanted. #939 will close when that merges, so there's nothing left here for you to redo.

If you'd like another one, these are all `good first issue` and each has a mechanical way to tell when it's done:

- #935 / #936 / #937 — runnable doctests for the combinator families. `cargo test --doc -p wingfoil --all-features` is the whole acceptance check.
- #802 — add a `start_with(value)` combinator. Small Rust; `/new-op` in the repo walks the touch points.
- #940 — make `setup-dev.sh` install the full native toolchain an `--all-features` build needs.

Happy to answer questions on any of them before you start. Thanks again for taking a swing at this one.

---

PR #951, opened 2026-09-02
feat(core): add debounce combinator

### LINKED ISSUE #803: Add `debounce(quiet_period)` — trailing-edge rate limiting
Part of #459. **Not a first issue — it schedules — but it is no longer blocked:** both decisions it used to carry are settled below. Pick this one up if you want to work with wingfoil's scheduler. If you want the same idea one size down, #928 (`audit`) is the non-re-arming sibling and is a good first issue.

**What:** `debounce(quiet_period)` emits the *latest* value only after `quiet_period` has passed with no new input.
**Why:** `throttle` is **leading-edge only** — it emits the first value of a burst and drops the rest, including the last one. For "the user stopped typing", "the burst settled", "emit the final state after the storm", that is exactly wrong: you want the trailing value. There is currently no way to express it. (That `throttle`'s own docs never say any of this is #929.)

## Carved out of this issue

Three pieces came out of here and are separately actionable, all of them good first issues:

| | | |
|---|---|---|
| #929 | `throttle` rustdoc: state the leading-edge contract | docs only — this issue's motivating paragraph, said where a user will see it |
| #927 | Document the `is_last_cycle` flush contract | settles decision (1) below |
| #928 | Add `audit(window)` — fixed-window trailing edge | the same op without the re-arm; strictly simpler |

None of them blocks this issue, and this issue does not block them.

## Semantics

- Each input value stores itself as pending and **re-arms** the timer for `now + quiet_period`, superseding the previous arming.
- When the timer fires with no intervening input, emit the pending value.
- If input never goes quiet, nothing is ever emitted — that is correct, and the rustdoc should say so. (`audit`, #928, is the operator for people who wanted the other thing.)

### The two decisions, now settled — do not re-open in the PR

1. **End of run with a value still pending: flush it.** Emit the pending value on the last cycle, matching `Window` and `Buffer`, which both already do this via `ctx.is_last_cycle()`. Note the caveat that comes with it: `is_last_cycle` is **not propagated into islands** (`Ctx::nested` hard-codes it `false`), so a `nested()` composite will not flush at the outer run's end. That does not affect the `op_completeness.rs` guard — `compiled()` owns a real `Kernel` and sets the flag — but it does belong in the rustdoc. #927 is writing this convention down properly.
2. **Interaction with `Burst`: nothing special is required.** A single upstream edge ticks at most once per cycle, so "several values at the same instant" is not a case this op has to disambiguate — the pending value is simply whatever the most recent cycle carried. A `Stream<Burst<T>>` debounces the burst as one value, which is right: the burst *is* the instant's value. Someone who wants the newest single value out of it composes with `collapse()`, and should read that op's data-loss warning before doing so.

## Shape

| | |
|---|---|
| `Cfg` | `Duration` |
| `State` | a small struct: pending `Option<T>`, the armed deadline `Option<NanoTime>`, and the period converted to `NanoTime` |
| `In<'a>` | `(&'a T, bool)` — value plus tick flag, so a scheduled wake is distinguishable from a real input |
| `Out` | `T` |
| `ACTIVATION` | `Activation::SCHEDULES` |
| Attribute | `#[op(build = debounce, fluent)]` |

**Reference op — read this first:** `Delay`, `crates/wingfoil/src/ops.rs` (search `pub struct Delay`; the line numbers this issue originally carried have drifted). It is the closest existing shape: `Cfg = Duration` hoisted to `NanoTime` in `start`, `ACTIVATION::SCHEDULES`, `In = (&'a T, bool)`, and `ctx.schedule(..)` from `cycle`. Hoist the `Duration` in `start` as it does — converting per cycle measurably slows dense graphs; `ThrottleState`'s rustdoc explains the `PhantomData` anchor you will need if your state does not otherwise mention `T`. `Throttle` (same file, search `pub struct Throttle`) is the leading-edge sibling this complements.

> 📌 **You probably do not need a `TimeQueue`, and that makes the re-arm easier than it looks.** Keep one armed deadline in `State`, call `ctx.schedule(deadline)`, and on wake compare `ctx.time()` against the stored deadline — a superseded arming fires, finds the deadline has moved, and returns `Tick::Quiet`. That is the standard stale-wake guard, and it sidesteps needing to *cancel* a pending entry. If you do reach for a `TimeQueue`, read its section of [`CLAUDE.md`](../blob/main/CLAUDE.md) first: **it deduplicates `(value, time)` pairs by design**, scoped to the instant being pushed to, and the re-arm has to work with that rather than around it.

## Acceptance

- [ ] `Op` impl in `crates/wingfoil/src/ops.rs` with `#[op(build = debounce, fluent)]`
- [ ] Fluent method declared on `StreamOps` in `src/fluent.rs`
- [ ] Tests asserting exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`, covering: a burst collapsing to its trailing value; re-arming (input arriving before the timer fires, so the earlier arming must not emit); a lone value still emitting after `quiet_period`; a source that never goes quiet emitting nothing; and the end-of-run flush per decision (1)
- [ ] Rustdoc contrasting it with `throttle` — and with `audit` if #928 has landed — explicitly. That contrast is the reason a reader picks one over the others

## Getting started

Run `/new-op debounce` (`.claude/commands/new-op.md`) for the full recipe, and read the scheduling sections of [`CLAUDE.md`](../blob/main/CLAUDE.md). Background: [`docs/adding-an-op.md`](../blob/main/docs/adding-an-op.md). Branch from `main`, PR base `main`.

maxi-maxima:
## What this changes

Adds StreamOps::debounce(quiet_period) as a sliding trailing-edge combinator. Each input replaces the pending value and re-arms one deadline; superseded wakes stay quiet, zero periods emit inline, and finite runs flush the pending value on their last cycle.

The same surface is available from Python as .debounce(quiet_period_nanos), with the public API tables and port inventory updated.

## Why

Closes #803.

throttle emits the first value in a burst and audit emits on fixed trailing windows. Neither expresses “emit the latest value only after the source has gone quiet,” which is needed for settled input and burst-final-state workflows.

## How it was verified

- [x] cargo fmt --all
- [x] Full default/all-features workspace Clippy, binding registration, dependency duplication, and example-doc checks: [Rust workflow](https://github.com/maxi-maxima/wingfoil/actions/runs/33582750828)
- [x] All-features Wingfoil tests, wingfoil-derive tests, and doctests: [Rust workflow](https://github.com/maxi-maxima/wingfoil/actions/runs/33582750828)
- [x] wingfoil-python Rust tests, maturin build, and pytest: [Python workflow](https://github.com/maxi-maxima/wingfoil/actions/runs/33582754738)
- [x] New behaviour is covered by exact historical values and tick times

Regression coverage includes re-arming, stale wakes, input exactly on an old deadline, lone values, continuously busy input plus final flush, zero periods, runner re-use, three execution tiers, must-use, and Python composition.

## Notes for the reviewer

A real input wins when it coincides with an obsolete scheduled wake: it replaces the pending value and moves the deadline before due-wake evaluation. The documented final flush has the existing nested-island boundary because islands do not receive the outer runner's is_last_cycle flag.

## Before you submit


0-jake-0:
Reviewed, and I ran the affected suites on this branch (`catalog_flow`, `compiled_stateful_ops`, `rerun`, `op_completeness`) — all green. The op itself looks right: stale wakes are correctly suppressed by comparing against the live deadline rather than by cancelling, `TimeQueue` dedup absorbs the re-arm, and `debounce_agrees_across_engines` deliberately runs past the t=55 deadline so `is_last_cycle` stays a no-op and the nested island agrees. Coverage is thorough.

Four things before this can land:

**1. It no longer merges cleanly.** `start_with` (#950, #955) landed in the same tables and tests. Conflicts in `crates/wingfoil/tests/catalog_flow.rs`, `crates/wingfoil/tests/op_completeness.rs`, `crates/wingfoil-python/src/graph.rs`, `crates/wingfoil-python/docs/api.rst`, `docs/planning/port-plan.md` and `docs/python-interop.md`. Please merge `main` in.

**2. `crates/wingfoil/src/ops.rs:775`** — `Audit`'s comparison table still points `debounce` at the issue:

```rust
/// | [`debounce`](https://github.com/wingfoil-io/wingfoil/issues/803) | latest value after the burst ends | trailing edge, sliding |
```

It is a real type as of this PR, so that should become an intra-doc link to [`Debounce`], and the `Unlike `debounce`` sentence just below it can link too.

**3. The `compiled_stateful_ops.rs` module doc enumerates the ops it covers** (`skip` / `skip_while` / `step_by` / `take_while` / `throttle` / `start_with` / `audit` / `window`). #955 has just added `start_with` to that list, so it is a maintained inventory — `debounce` should join it in the rebase, noting it is `ACTIVATION::SCHEDULES` and exercises the `start` hook, the same way `audit` is described.

**4. The deadline tie deserves a comment in the code.** A source tick landing exactly on the armed deadline re-arms and discards the pending value — the opposite convention to `Audit`, where a value arriving exactly on the deadline starts the next window and the closing window still emits. That is the right call for a sliding window, but right now only the name of `debounce_input_at_the_deadline_rearms_without_emitting` says so: `Audit::cycle` carries an inline comment at exactly that spot, and the rustdoc on `Debounce` does not state the rule at all. One sentence in the type doc plus a comment above the `src_ticked` branch would make the asymmetry deliberate to the next reader.

Nothing else from me — the Python binding, the must-use `trybuild` refresh and the rerun coverage are all in the right places.

---

maxi-maxima:
Merged current `main` into this branch and resolved the overlapping `start_with` additions without dropping either operator. I also addressed the review points: Audit now uses intra-doc `Debounce` links; the compiled-stateful-op inventory documents debounce scheduling, activation flags, and start-hook forwarding; and the debounce rustdoc plus inline code define source-wins behavior for an exact-deadline tie.

Validation after the merge: `cargo fmt --all -- --check`, focused debounce catalog/compiled/op-completeness/rerun/trybuild tests, `cargo check -p wingfoil-python`, `cargo lint`, and `cargo lint-all` all pass. The broad `cargo test -p wingfoil` run only hit the known coarse-clock latency flake in `stamp_all_takes_one_clock_read_per_stage_when_precise`; its isolated rerun passed. The local container lacks `libpython3.12` for linking Python tests, so upstream CI remains the final Python execution check.

0-jake-0:
Re-reviewed at `bf09dcc`. All four points from the previous round are addressed, and I checked each against the diff rather than the summary:

1. **Merge.** `main` is in, and the `start_with` overlap is resolved without dropping either operator — the port-plan Scheduling row, `catalog_flow.rs`, `op_completeness.rs`, `graph.rs`, `api.rst` and `python-interop.md` all carry both.
2. **`Audit`'s table** now links `[`Debounce`]` in both the row and the `Unlike` sentence, replacing the issue URL.
3. **`compiled_stateful_ops.rs`** module doc lists `debounce` in the op inventory with its `SCHEDULES` activation and `start`-hook forwarding, matching how `audit` is described.
4. **The deadline tie** is stated twice now, in the `Debounce` rustdoc ("A source tick exactly on the armed deadline re-arms first and suppresses the pending value") and as an inline comment above the `src_ticked` branch — the asymmetry against `Audit` is deliberate to the next reader.

Against #803's acceptance list: op with `#[op(build = debounce, fluent)]` ✅, `StreamOps` declaration ✅, `op_completeness.rs` `nitro!` guard ✅, and the five required test cases — burst collapsing to its trailing value, re-arming with stale wakes staying quiet, a lone value emitting after the period, a never-quiet source, and the end-of-run flush — all present with exact values and tick times ✅. The `throttle`/`audit`/`debounce` contrast table is on both types. Both settled decisions are honoured, including the documented nested-island `is_last_cycle` caveat. CI is green across test, lint, Python interop and the audits.

One optional nit, not worth a round on its own: the port-plan Scheduling row now reads `node_flow (node-level delay/filter/limit/throttle/audit/debounce, …)`, but `node_flow` is the legacy-mirror suite and there is no node-level `debounce` case (nor an `audit` one — the row was already loose here). If you touch this file again, moving `audit`/`debounce` out of that parenthetical and leaving them in the leading list would make the row true.

Nothing blocking. This is ready to merge from my side.

---

PR #952, opened 2026-09-03
feat(dev): install all-feature native toolchain

### LINKED ISSUE #940: `setup-dev.sh`: install the native toolchain an `--all-features` build needs, not just protoc

maxi-maxima:
## What this changes

Adds an idempotent `scripts/setup-dev.sh --all-features` mode for the native Aeron and ZeroMQ toolchain while preserving the existing protoc-only default. It supports apt, dnf, pacman, and Homebrew, including a checked CMake >= 3.30 fallback.

## Why

Closes #940. Contributors should not have to discover the all-features native dependencies after a long build fails at CMake or link time.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all` (no Rust changed; clean-container all-features build used instead)
- [ ] `cargo test -p wingfoil --all-features` (no Rust changed; shell regression and build used instead)
- [ ] New behaviour is covered by a test asserting **values and tick times** (not applicable to a setup script)
- `scripts/test-setup-dev.sh`
- ShellCheck (stable) on both shell scripts
- actionlint v1.7.12 on the updated workflow
- Clean Ubuntu 24.04: ran the setup twice, then `cargo build --locked -p wingfoil --all-features`

## Notes for the reviewer

The all-features package set also includes ZeroMQ/pkg-config and OpenSSL headers because the real clean-container build requires them. The pinned Kitware installers are SHA-256 verified before elevation; dnf enables EPEL only when `libbsd-devel` is unavailable.

## Before you submit

### DIFF
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -256,6 +256,9 @@ jobs:
       - name: Install system dependencies
         run: sudo apt-get update && sudo apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev
 
+      - name: Check developer setup script
+        run: scripts/test-setup-dev.sh
+
       - name: Run rustfmt
         run: cargo fmt --all -- --check
 
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -254,7 +254,7 @@ jobs:
           tool: sarif-fmt
 
       - name: Install system dependencies
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev
+        run: scripts/setup-dev.sh --all-features
 
       - name: Check developer setup script
         run: scripts/test-setup-dev.sh
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -306,26 +306,18 @@ files). On Debian/Ubuntu: `sudo apt-get install -y protobuf-compiler`, or run
 `scripts/setup-dev.sh`. Note this is needed for a *plain* `cargo build
 --workspace`, not just `lint-all` — see the feature-unification note below.
 
-### Aeron adapter system dependencies
+### All-features system dependencies
 
-The Aeron adapter requires clang, libuuid, libbsd, and a recent CMake
-(**>=3.30** — the vendored `rusteron-media-driver` Aeron sources set that floor
-in their own `cmake_minimum_required`, so 3.28 fails the build script outright
-with `CMake 3.30 or higher is required`):
+Use the setup script for the native Aeron and ZeroMQ build dependencies:
 
 ```bash
-sudo apt update
-sudo apt install clang libclang-dev uuid-dev libbsd-dev
-
-# CMake 3.31 (apt version is too old on many distros)
-wget https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-linux-x86_64.sh
-sudo ./cmake-3.31.0-linux-x86_64.sh --prefix=/usr/local --skip-license
+scripts/setup-dev.sh --all-features
 ```
 
-`libbsd-dev` is the one that is easy to miss, because nothing fails until
-**link** time and only on an `--all-features` target: the vendored Aeron C
-client emits `-lbsd`, so a long, apparently successful build ends with
-`rust-lld: error: unable to find library -lbsd`.
+This keeps the ordinary protoc-only setup small. The explicit mode also checks
+that CMake is at least 3.30; older Linux packages are replaced with the pinned
+Kitware build required by the vendored Aeron sources. Missing libbsd otherwise
+surfaces only at link time as `rust-lld: error: unable to find library -lbsd`.
 
 ### Disk space
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -315,8 +315,9 @@ scripts/setup-dev.sh --all-features
 ```
 
 This keeps the ordinary protoc-only setup small. The explicit mode also checks
-that CMake is at least 3.30; older Linux packages are replaced with the pinned
-Kitware build required by the vendored Aeron sources. Missing libbsd otherwise
+that CMake is at least 3.30 because the vendored `rusteron-media-driver` Aeron
+sources set that floor in their own `cmake_minimum_required`; older Linux
+packages are replaced with the pinned Kitware build. Missing libbsd otherwise
 surfaces only at link time as `rust-lld: error: unable to find library -lbsd`.
 
 ### Disk space
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,8 @@ You need the Rust toolchain (latest stable, with `rustfmt` and `clippy`) and
 build needs it:
 
 ```bash
-scripts/setup-dev.sh              # installs protoc; Debian/Ubuntu and macOS
+scripts/setup-dev.sh                 # protoc only
+scripts/setup-dev.sh --all-features  # also Aeron/ZeroMQ native dependencies
 ```
 
 Then check everything works end to end:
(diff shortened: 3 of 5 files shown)


0-jake-0:
Reviewed. The script itself reads well — idempotency checks per package manager, SHA-256 verification before elevation, EPEL only when `libbsd-devel` is actually missing. I verified the two pinned digests against Kitware's published `cmake-3.31.0-SHA-256.txt`: both `linux-x86_64` and `linux-aarch64` match. Good.

One real bug and a few smaller things.

**1. `scripts/test-setup-dev.sh` fails when run as root.** Reproduced both ways on this branch — as a normal user it prints `setup-dev tests passed`; as root it fails:

```
FAIL: expected 'sudo sh …/cmake-installer.sh --prefix=/usr/local --skip-license --exclude-subdir' in
'apt-get update
apt-get install -y build-essential …
curl -fsSL https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-linux-x86_64.sh -o …
sha256sum …'
```

`run_as_root` calls the command directly when `EUID == 0`, so the `sudo` mock never runs and never logs. GitHub-hosted runners are non-root, so the new CI step stays green — but the devcontainer and any root container will fail it, which is a bad first experience for a script whose whole point is onboarding. The same cause quietly guts the `bad-cmake-checksum` case: `assert_not_contains "$CALLS" "sudo sh"` passes vacuously under root, so that test stops proving the installer was not executed.

Suggested fix: make the `sh` mock log its arguments, then assert on `sh <installer> --prefix=/usr/local --skip-license --exclude-subdir` (and `assert_not_contains "$CALLS" "sh "` for the checksum case) rather than on the `sudo` prefix. Elevation can be asserted separately, guarded on `EUID != 0`.

**2. The checksum test is circular.** `sha256sum`'s mock echoes the same literal the script compares against, so a wrong or stale pin passes. Worth saying so in a comment next to the constants — the real check is against Kitware's `SHA-256.txt`, done by hand.

**3. CI still duplicates the package list.** `rust-test.yml` keeps its inline `apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev`, so the two lists can drift and the script is never exercised for real anywhere — only against mocks. Replacing that step with `scripts/setup-dev.sh --all-features` would dogfood it and remove the duplication in one move.

**4. CLAUDE.md loses the *why* for the CMake floor.** The replacement keeps the libbsd link-time note but drops the reason 3.30 is the floor (the vendored `rusteron-media-driver` Aeron sources set it in their own `cmake_minimum_required`, so 3.28 fails the build script outright). One sentence is worth keeping — that file is read by people who are diagnosing a failure, not running the script.

**5. Minor.** `--help` prints usage to stderr and exits 2; conventionally it should go to stdout and exit 0. And on macOS `brew llvm` is keg-only, so bindgen may need `LIBCLANG_PATH` — worth a line in the script's final "toolchain ready" output or in CONTRIBUTING.

---

maxi-maxima:
Addressed the review in \c833300\:

- The \sh\ mock now logs installer execution, so the install and bad-checksum assertions work both as root and non-root; sudo elevation is checked separately only when applicable.
- The Rust CI job now runs \scripts/setup-dev.sh --all-features\ instead of duplicating the apt package list, then runs the mocked regression suite.
- Added the trust-boundary comment for the pinned Kitware digests and restored the exact vendored \cmake_minimum_required\ reason for the 3.30 floor.
- \--help\/\-h\ now print to stdout and exit 0; invalid arguments still exit 2.
- Homebrew completion output now gives the concrete \LIBCLANG_PATH\ value from \rew --prefix llvm\.

Validation: the setup regression passed in containers as UID 0 and UID 1000; ShellCheck passed for both scripts; actionlint 1.7.12 passed for \
ust-test.yml\; \git diff --check\ passed. The refreshed upstream CI will exercise the real Ubuntu setup path.

0-jake-0:
Re-reviewed at `c833300`. All five points are addressed, and the fixes are the right ones:

1. **Root-vs-non-root.** The `sh` mock now logs, and the install case asserts on `sh <installer> --prefix=/usr/local --skip-license --exclude-subdir` with the `sudo`-prefixed assertion guarded on `EUID != 0`. The `bad-cmake-checksum` case asserts `assert_not_contains "$CALLS" "sh $MOCK_STATE/cmake-installer.sh"`, so it now genuinely proves the installer was not executed under either UID rather than passing vacuously.
2. **Circularity** is disclosed in a comment next to the pins. I re-verified both digests against Kitware's published `cmake-3.31.0-SHA-256.txt` on this pass: `x86_64` `7cfdf4a4…e368a` and `aarch64` `45bb6140…e965` both match.
3. **Dogfooding** — the lint job now runs `scripts/setup-dev.sh --all-features`, so the real apt path is exercised in CI instead of only against mocks. See the one residual below.
4. **CLAUDE.md** keeps the `cmake_minimum_required` reason for the 3.30 floor and the libbsd link-time note.
5. **`--help`** goes to stdout and exits 0, invalid arguments still exit 2, and the Homebrew completion line prints the concrete `LIBCLANG_PATH` from `brew --prefix llvm`.

**One residual, and it is small.** `rust-test.yml` has three `Install system dependencies` steps — `Coverage (unit)` (~L97), `Test (wingfoil)` (~L155) and `Lint` (~L256). Only the lint one moved to the script; the other two still carry the literal `apt-get install -y protobuf-compiler cmake clang libclang-dev uuid-dev libbsd-dev`, which is the drift this PR is trying to remove. Same list also appears in `python-test.yml` (plus `patchelf`) and `crates-publish.yml`. Switching the two remaining `rust-test.yml` steps in this PR would finish the job cheaply and give the script three real exercises instead of one; the other two workflows can follow separately since they add extras.

Worth confirming for the record: the package list grows by `build-essential ca-certificates curl libzmq3-dev libssl-dev pkg-config`. `libzmq3-dev` is genuinely load-bearing under `--all-features` — the `zmq` feature links the system `libzmq` (`crates/wingfoil/Cargo.toml`), and today only `zmq-integration.yml` installs it, so the previous list was relying on it being present in the runner image. Making it explicit is a correctness improvement, not scope creep.

CI is green. Happy to merge once the two remaining `rust-test.yml` steps are switched over — or as-is if you would rather do that in a follow-up; say which and I will not hold it.

---

PR #956, opened 2026-09-05
docs(core): add limit, skip, and step_by doctests

maxi-maxima:
## What this changes

Adds deterministic runnable doctests for the `limit`, `skip`, and `step_by` `StreamOps` combinators. Each example uses a bounded historical run and asserts the exact emitted sequence. Prose explains value-counted behavior and first-value stride semantics; additional assertions show that `limit` does not end the graph run and that `step_by(0)` returns an error.

## Why

These count-based controls are easy to confuse from signatures alone. The examples make the first-kept value, stride, and error semantics executable and visible on docs.rs.

Refs #936 (focused partial: `limit`, `skip`, and `step_by`).

## How it was verified

- [x] `cargo lint`
- [x] `cargo lint-all`
- [x] `cargo test --locked --doc -p wingfoil --all-features` — 26 passed, 22 intentionally ignored
- [x] `cargo test --locked -p wingfoil --all-features --lib` — 350 passed
- [x] `cargo test --locked -p wingfoil --test catalog --test catalog_filter_value_scan --test engine_semantics` — 52 passed
- [x] `cargo test --locked -p wingfoil-derive` — 3 passed
- [x] New documentation is covered by runnable examples asserting exact values and the zero-stride error

## Notes for the reviewer

This intentionally leaves the other #936 methods for separate focused PRs. Timing is not the point of these three count-based operators, so the examples assert values rather than tick times. Local verification used Linux and Rust 1.98; service-backed integration tests were not run locally.

The refreshed [CI dependency audit](https://github.com/wingfoil-io/wingfoil/actions/runs/35092542958/job/104782029063) is blocked by RUSTSEC-2026-0285 in the existing `rustls 0.23.43` lock entry. Current `main` uses the same version. This docs-only PR does not change dependencies; CI is not fully green.

## Before you submit

### DIFF
--- a/crates/wingfoil/src/fluent.rs
+++ b/crates/wingfoil/src/fluent.rs
@@ -1280,12 +1280,40 @@ pub trait StreamOps<T>: Sized {
         F: Fn(Stream<T>) -> Stream<B>;
 
     /// Pass through the first `limit` values, then stay quiet.
+    ///
+    /// ```
+    /// use std::time::Duration;
+    /// use wingfoil::prelude::*;
+    /// use wingfoil::{NanoTime, RunFor, RunMode};
+    ///
+    /// let g = GraphBuilder::new();
+    /// let values = g.ticker(Duration::from_nanos(10)).count();
+    /// let limited = values.limit(3).accumulate();
+    /// let mut r = g.build();
+    /// r.run(RunMode::HistoricalFrom(NanoTime::ZERO), RunFor::Cycles(5))
+    ///     .unwrap();
+    /// assert_eq!(r.value(&limited), vec![1u64, 2, 3]);
+    /// ```
     #[must_use = "a dropped stream stays wired and cycles every tick, producing an unread value"]
     fn limit(&self, limit: usize) -> Stream<T>
     where
         T: Clone + Default + 'static;
 
     /// Suppress the first `n` values, then pass every later value through.
+    ///
+    /// ```
+    /// use std::time::Duration;
+    /// use wingfoil::prelude::*;
+    /// use wingfoil::{NanoTime, RunFor, RunMode};
+    ///
+    /// let g = GraphBuilder::new();
+    /// let values = g.ticker(Duration::from_nanos(10)).count();
+    /// let skipped = values.skip(2).accumulate();
+    /// let mut r = g.build();
+    /// r.run(RunMode::HistoricalFrom(NanoTime::ZERO), RunFor::Cycles(5))
+    ///     .unwrap();
+    /// assert_eq!(r.value(&skipped), vec![3u64, 4, 5]);
+    /// ```
     #[must_use = "a dropped stream stays wired and cycles every tick, producing an unread value"]
     fn skip(&self, n: usize) -> Stream<T>
     where
@@ -1303,6 +1331,20 @@ pub trait StreamOps<T>: Sized {
 
     /// Emit the first value, then every `n`th value after it. A zero `n`
     /// returns an error when the graph runs instead of panicking.
+    ///
+    /// ```
+    /// use std::time::Duration;
+    /// use wingfoil::prelude::*;
+    /// use wingfoil::{NanoTime, RunFor, RunMode};
+    ///
+    /// let g = GraphBuilder::new();
+    /// let values = g.ticker(Duration::from_nanos(10)).count();
+    /// let stepped = values.step_by(2).accumulate();
+    /// let mut r = g.build();
+    /// r.run(RunMode::HistoricalFrom(NanoTime::ZERO), RunFor::Cycles(5))
+    ///     .unwrap();
+    /// assert_eq!(r.value(&stepped), vec![1u64, 3, 5]);
+    /// ```
     #[must_use = "a dropped stream stays wired and cycles every tick, producing an unread value"]
     fn step_by(&self, n: usize) -> Stream<T>
     where


0-jake-0:
Reviewed at `729d2e3`. The three examples are correct — I traced each against the ops rather than trusting the doctest run: a 10ns ticker under `Cycles(5)` produces counts 1..5 with no scheduling op to insert extra cycles, so `limit(3)` → `[1,2,3]`, `skip(2)` → `[3,4,5]`, and `StepBy`'s `index.is_multiple_of(n)` on a zero-based counter gives `[1,3,5]`. Deterministic `HistoricalFrom` runs, bounded, exact assertions — the right shape. CI is green, including `Lint (fmt & clippy)`, which is `lint-all`, so that unticked box in the description is covered.

Three things before it lands.

**1. The prose lead-in is missing, and it is the house pattern.** Every runnable doctest already in `fluent.rs` is preceded by a sentence naming what to *notice* in the output — `filter`: "The true condition cannot emit before the first source tick. At 30ns the source ticks while the condition is false…"; `join_passive`: "Unlike `join` on these same inputs, the fast ticks at 10ns and 30ns do not emit." Without it the example shows a result but not the point, which is most of what a docs.rs reader comes for. The non-obvious facts here are each one sentence:

- `limit` — the count is over *values*, not cycles, and the stream stays quiet for the rest of the run rather than the run ending (the `Cycles(5)` against `limit(3)` is deliberately proving that; say so).
- `skip` — likewise value-counted, so the first value through is the (n+1)th, `3` here.
- `step_by` — the first value is always emitted and the stride is counted from it, which is why `[1,3,5]` and not `[2,4]`.

**2. `step_by` is one assertion away from documenting its own error contract.** The declaration promises "A zero `n` returns an error when the graph runs instead of panicking" — `StepBy::start` bails with `step_by requires n > 0`. That is the genuinely surprising part of the API and the part a doctest can pin where prose cannot:

```rust
let g = GraphBuilder::new();
let bad = g.ticker(Duration::from_nanos(10)).count().step_by(0);
let mut r = g.build();
assert!(r.run(RunMode::HistoricalFrom(NanoTime::ZERO), RunFor::Cycles(1)).is_err());
```

Three lines, and it makes the documented behaviour executable. Worth adding here rather than deferring.

**3. Cosmetic, to match the neighbours.** The existing doctests keep the `use` block and the body contiguous (no blank line) and put `r.run(…).unwrap();` on one line. Yours inserts a blank line and wraps the `run` call across two. Not rustfmt-driven — doctests are not formatted — so it is a straight divergence from the four examples immediately around it.

On the timing question in your notes: agreed for `skip` and `step_by`. For `limit` a `with_time()` would carry a little more (it makes "kept ticking, stayed quiet" visible rather than inferred), but the `Cycles(5)` already establishes it, so treat that as optional rather than a request.

Scoping this to three of #936's methods is the right call — small doc PRs review faster than one large one.

---

maxi-maxima:
@0-jake-0 Addressed all three points in 8499bd5:

- Added the prose lead-ins for value-counted `limit`/`skip` and first-value stride semantics. The `limit` example also asserts the source reached `5`, making the continued run explicit.
- Added the runnable `step_by(0)` assertion that `run` returns an error.
- Removed the blank lines after the imports and kept each successful `run(...).unwrap()` on one line, matching the neighboring examples.

Revalidated on Linux/Rust 1.98: 26 all-feature doctests passed (22 intentionally ignored), 350 all-feature library tests, 52 focused catalog/engine tests, and 3 derive tests passed. Formatting, `cargo lint`, and `cargo lint-all` also passed. Updated the PR checklist; the scope remains the same three-method partial of #936.

CI follow-up: the refreshed [dependency audit](https://github.com/wingfoil-io/wingfoil/actions/runs/35092542958/job/104782029063) fails on the existing `rustls 0.23.43` lock entry and the newly published RUSTSEC-2026-0285 advisory. Current `main` has the same version; this docs-only change does not modify dependencies. A dependency update is still needed separately, so the PR is not fully green.


Update: the separate security fix is now [#970](https://github.com/wingfoil-io/wingfoil/pull/970), and its Cargo/pnpm audits, dependency review, Rust test/lint, and Python checks have all passed. This documentation branch remains unchanged; its audit will still need a refreshed base and rerun after the dependency fix lands.

0-jake-0:
Re-reviewed at `8499bd5`. All three points from my earlier pass are addressed properly, not minimally:

1. **Prose lead-ins are in and say the right thing** — value-counted rather than cycle-counted for `limit`/`skip`, first-value stride for `step_by`. The extra `assert_eq!(r.value(&values), 5)` on the `limit` example is better than what I asked for: it makes "kept ticking, stayed quiet" executable instead of inferred.
2. **The `step_by(0)` assertion is there**, which is the one genuinely surprising bit of that API now pinned by a doctest rather than a promise in prose.
3. **Formatting matches the neighbours** — contiguous `use` block, one-line `run(..).unwrap()`.

Values still check out against the ops: counts 1..5 under a 10ns ticker with `Cycles(5)`, so `[1,2,3]` / `[3,4,5]` / `[1,3,5]`.

Two things before it lands, neither about the diff:

- **Rebase onto `main` once #970 merges.** Your base is `bc8ff59`; `main` is at `5e509a5`. The red `cargo audit` here is RUSTSEC-2026-0285 against `rustls 0.23.43`, which #970 fixes — thanks for chasing that down and filing it separately rather than folding a dependency bump into a docs PR. That was the right call.
- **#973 touches the same trait.** It adds doctests to `map`/`try_map`/`map_filter` in `StreamOps`, a few hundred lines above `limit`. Different hunks, so I don't expect a textual conflict, but whichever lands second wants a rebase check.

One micro-nit for consistency across both your doc PRs: `n % 2 == 0` in #973's `map_filter` example vs `n.is_multiple_of(2)` here in the tree (`lib.rs:12`, `ops.rs:652`). MSRV is 1.88, so the house idiom is available — worth matching in #973.

---

PR #968, opened 2026-09-16
docs(core): state the throttle leading-edge contract and dropped trailing value

### LINKED ISSUE #929: `throttle` rustdoc: state the leading-edge contract and the dropped trailing value
Split out of #803, which had to say this before it could explain what `debounce` is *for*. It stands on its own, it is unblocked, and it is docs-only.

## The gap

`throttle` is **leading-edge only**: it emits the first value of a burst and suppresses the rest — *including the last one*. Neither doc site says so.

- **The op rustdoc** (`Throttle`, `crates/wingfoil/src/ops.rs` — search `pub struct Throttle`) says "emits the first value, then suppresses until at least `interval` has passed since the last emit". That is the mechanism, stated accurately, but the reader has to derive the consequence themselves.
- **The fluent method** (`StreamOps::throttle`, `crates/wingfoil/src/fluent.rs`) says only *"Rate-limit: emit at most once per `interval`."* That is actively misleading about **which** value you get — and the fluent doc is the one most readers actually see, because it is what shows up at the call site.

Why it matters: for "the user stopped typing", "the burst settled", "emit the final state after the storm", you want the **trailing** value. Someone reaching for a rate limiter picks it by name, gets the leading value, and nothing tells them.

## The work

- [ ] `Throttle`'s rustdoc: name the contract (*leading edge*) and state the consequence — the last value of a burst is dropped unless it happens to fall outside the suppression window.
- [ ] `StreamOps::throttle`'s doc comment: the same, in a line. "Emit at most once per `interval`" alone is not enough to pick correctly.
- [ ] Say what to reach for when the trailing value is the one you want: link #928 (`audit`, fixed window) and #803 (`debounce`, sliding window). If either has landed by the time you write this, link the op itself rather than the issue.

## House model

`Collapse`'s rustdoc, in the same file, is the model for this kind of warning — match its directness rather than hedging. Note in particular its observation that **the loss only appears under load**: a source producing one value per cycle never bursts, so `throttle` looks lossless in testing and starts dropping values exactly when the graph gets busy. The same is true here, and it is the part worth saying out loud.

Docs only — no behaviour change, no test changes expected. Branch from `main`, PR base `main`.

tunglambk:
## What this changes

Documents `throttle` as leading-edge-only at both places it is described: the `Throttle` op rustdoc in `crates/wingfoil/src/ops.rs` and the fluent `StreamOps::throttle` doc in `crates/wingfoil/src/fluent.rs`. Both now state that the last value of a burst is dropped, and point at `Audit` (fixed window, trailing edge) and `Debounce` (sliding window, trailing edge) for when the trailing value is what you want.

## Why

Closes #929. `Throttle` emits the first value of a burst and suppresses the rest — including the last one. The op doc gave the mechanism but left the consequence to the reader, and the fluent doc ("Rate-limit: emit at most once per `interval`") was misleading about which value you get; that is the doc most readers see at the call site. The new op doc follows `Collapse`'s framing, including that the loss only appears under load.

## How it was verified

- [x] `cargo fmt --all -- --check` — clean.
- [ ] `cargo lint` / `cargo lint-all` — ran `cargo clippy -p wingfoil --lib -- -D warnings`, clean. The full workspace `--all-targets` / `--all-features` clippy needs native Aeron/iceoryx2 dependencies this sandbox does not have.
- [ ] `cargo test -p wingfoil --all-features` — ran default-feature `cargo test -p wingfoil`: every target passes except `trybuild`, which fails on two compile-fail expectation mismatches under rustc 1.88 (`must_use_combinators`, `unknown_combinator`). The same two mismatches reproduce on unmodified `main` with this change stashed, so they are a toolchain-version artifact rather than this diff.
- [x] `cargo test -p wingfoil --doc` — 14 passed, 0 failed.
- [x] `cargo doc -p wingfoil --no-deps` — the new intra-doc links resolve; the remaining warnings are pre-existing feature-gated links.
- [ ] New behaviour is covered by a test asserting **values and tick times** — not applicable: docs only, no runtime change.

I also confirmed the trailing drop directly: a 10ns counter through `throttle(25ns)` over 6 cycles emits `[(0, 1), (30, 4)]`; the value at t=50ns (count 6) is suppressed and never emitted, unlike `Audit`/`Debounce` which flush a pending value on the last cycle.

## Notes for the reviewer

Docs only — no runtime, test-expectation, or dependency changes. The `Audit`/`Debounce` comparison table already lists `Throttle` as leading edge, so this makes the prose match it.

## Before you submit


0-jake-0:
Reviewed at `a21b261`. Docs-only, and I checked every claim against `Throttle` in `ops.rs` rather than taking the prose on trust:

- `cycle` returns `Tick::Quiet` on suppression and `ThrottleState` holds `{ interval, last_emit, _marker }` — no pending slot. So "discarded, not deferred" is literally true, and there is no last-cycle flush to qualify it with.
- `Audit` (`ops.rs:827`) and `Debounce` (`ops.rs:930`) both exist, and are the right redirections for the trailing-value case — `[Collapse]` (`ops.rs:482`) resolves too, so the intra-doc links are live.
- The `Collapse` parallel is the best thing in this PR. "Looks lossless in testing because a source that produces one value per cycle never bursts" is exactly the shape of bug this repo keeps re-learning, and pointing at the op that already documents it is the cheapest way to say it.

The fluent declaration and the op doc now agree, which they didn't before — the old one-line "Rate-limit: emit at most once per `interval`" was true and useless.

One wording nit, take it or leave it: "emit the first value of a burst" is slightly stronger than the code, which emits a value only if `interval` has already elapsed since the last emit — a burst arriving 1ns after a previous emit drops its first value too. The next clause ("then at most once per `interval`") recovers it, so it isn't wrong in aggregate, but "emit on the leading edge, then at most once per `interval`" would be exact without costing a word.

Blocked only on #970 — `cargo audit` is red here on the RUSTSEC-2026-0285 `rustls` entry that is red on `main` too, and nothing in this diff touches dependencies.

---

PR #970, opened 2026-09-16
fix(deps): require patched rustls 0.23.45

maxi-maxima:
## What this changes

Raise the optional `rustls` requirement to `0.23.45` and refresh its locked dependency closure (`aws-lc-rs`, `aws-lc-sys`, and `rustls-webpki`). No adapter code, feature flags, or certificate-validation settings change.

## Why

[RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285.html) affects the `0.23.43` version on main: TLS 1.3 handshake messages can be accepted across encryption-level boundaries. `0.23.45` is the first patched release. This also explains the unrelated audit failure on #956.

The manifest floor matters as well as the lockfile: published-library consumers do not use this repository's lockfile. The existing FIX, web TLS, and WSS paths share this optional dependency.

## How it was verified

Linux, Rust 1.98.0, using the committed lockfile:

- [x] `cargo fmt --all -- --check`, `cargo metadata --locked --format-version 1`, and `python3 scripts/check-dep-duplicates.py`
- [x] `cargo audit`: the old lock reproduces RUSTSEC-2026-0285; the updated lock reports zero vulnerability entries, with no advisory ignores. Existing informational warnings remain.
- [x] `cargo test --locked -p wingfoil --features web-tls-integration-test --test web_integration --test web_adapter -- --test-threads=1`: 27 passed, including the certificate-validated TLS/WSS round trip and missing-certificate rejection.
- [x] `ws_adapter` with `ws-tls`: 12 passed; the `ws`-only WSS rejection control also passed. `fix_adapter` with `fix`: 2 passed.
- [x] CI's non-service selection: `cargo nextest run --locked -p wingfoil --all-features --lib --tests --no-fail-fast -E 'not binary(/_integration$/)'`: 979 passed.
- [x] All-feature doctests: 23 passed, 23 intentionally ignored; `wingfoil-derive`: 3 passed.
- [x] Both default-feature and all-feature workspace/all-target Clippy passes with `--locked` and `-D warnings` (`cargo lint` / `cargo lint-all` equivalents).

The first broad test run hit the existing `latency_bursts::stamp_each_all_reads_the_clock_once_per_stage_not_per_value` clock-resolution assertion. It reproduced on unmodified main without TLS features (two clock reads returned equal values); the complete candidate rerun passed all 979 tests. No assertion was changed or excluded.

## Notes for the reviewer

The lock contains one Rustls version. The separate WASM workspace has no Rustls dependency and is untouched. Service-backed adapter suites other than the self-contained web/TLS loopback suite were not run locally; the FIX check is not an external FIX-over-TLS end-to-end test. The protocol-level regression itself is covered upstream; locally the security check is the before/after advisory audit plus the existing legitimate TLS controls.

Per the dependency policy, raising a published floor should accompany a **minor release** when maintainers publish; this PR does not edit release versions.

## Before you submit

- [x] Allow edits by maintainers


0-jake-0:
Reviewed at `025cf5e`. **This should land first** — `cargo audit` is red on every other open PR (#956, #968, #971, #972, #973) and on `main`, and this is the only branch where it's green. Merging it unblocks all five.

The change itself is right and minimal:

- **Patch granularity is justified and commented**, which is exactly the carve-out the dependency policy allows ("add a `.patch` component only when a specific patch release introduced something we use, and say so in a comment"). 0.23.45 being the first patched release is the reason; the comment says so.
- **Manifest floor as well as the lock** is the correct reading of the policy — consumers never see `Cargo.lock`, so a lock-only bump would have fixed CI and left published users on the vulnerable range.
- One `rustls` version in the lock, and `rustls-webpki`/`aws-lc-*` move as its closure rather than as independent bumps.
- You're right not to touch crate versions here; `bump.yml` owns that, and a floor raise being a minor release is the maintainer's call at publish time.

**One follow-up this surfaced, not yours to fix here.** `crates/wingfoil/Cargo.toml:288` declares `rustls = { ..., features = ["ring"] }` with no `default-features = false`, so rustls's default `aws_lc_rs` provider is enabled alongside ring. Confirmed it's really in the graph, not just the lock:

```
$ cargo tree --locked -p wingfoil --features fix -i aws-lc-rs
aws-lc-rs v1.17.0
├── rustls v0.23.43
│   └── wingfoil v9.0.0
└── rustls-webpki v0.103.13
    └── rustls v0.23.43 (*)
```

That's a cmake/C build (`aws-lc-sys`) in the tree for a provider the adapters never install — the `ws-tls` comment at line 423 says the adapter installs ring explicitly. It's also why this PR has to churn `aws-lc-sys` 0.41 → 0.45 at all. Policy rule 6 covers it; separate PR, and a minor bump since it narrows a dependency's default features.

---

PR #971, opened 2026-09-17
feat(ws): render on-connect payloads at connect time

### LINKED ISSUE #969: ws: on-connect payloads should be rendered at connect time, not fixed at wiring
### What kind of addition

An I/O adapter capability — a small extension to the existing `ws` adapter.

### The problem

`ws` already owns the thing hand-rolled reconnect loops most often get wrong: it re-sends `WsConfig::subscriptions` after **every** connect, so a dropped socket comes back with its feed restored instead of silently subscribed to nothing. That set is frozen at wiring time, and two ordinary situations cannot be expressed inside it.

**1. A subscription set that changes while the graph runs.** Venues list and delist instruments during a session. A graph that discovers a new instrument subscribes to it through `WsSender::send` — and the adapter documents what happens next:

> Frames sent while the socket is down are **queued, not dropped**, and go out after the next connect completes its subscriptions. Anything that must survive a reconnect belongs in `WsConfig::subscriptions` instead — the queue replays a message once, not on every reconnect.

So every subscription added after wiring is lost at the next reconnect. The socket comes back, `WsStatus` reads `Connected`, the originally-configured channels resume — and everything discovered since is simply absent. Nothing errors and nothing looks unhealthy; the data just stops for part of the universe. That is the exact failure mode `WsConfig::subscriptions` exists to prevent, reintroduced for anything the graph learns about at runtime.

**2. An authenticated stream.** Venues that gate private channels behind an auth frame need a signed, timestamped payload sent before any subscription, on every connect. A signature over a nonce or a timestamp cannot be a constant frozen at wiring, so the auth frame cannot live in `subscriptions` at all — and routed through `WsSender::send` it hits problem 1, with the reconnect coming back unauthenticated.

Both are the same shape: **what must go out on connect is state as of connect time, not a constant.** Neither asks the transport to understand a payload — it stays "send these bytes after connect, before the frames start", which is what the adapter already does.

Working around it above the adapter is possible, but everyone writes the same subtle code: fold over `WsStatus`, re-derive the whole payload set on each `Connected`, push it through `WsSender`. The part that is easy to miss is that a frame queued while the socket was down must not go out **twice** — once from the queued backlog, once from the re-derivation. The connection loop is the only place that knows the ordering, so it is the cheap place to get that right once.

### How it might look

```rust
// `subscriptions` becomes the degenerate case of this.
let cfg = WsConfig::new(url).on_connect(move || {
    let mut frames = vec[auth_frame_signed_now()];
    frames.extend(live_subscriptions.lock().unwrap().iter().map(subscribe_frame));
    frames
});
```

A `Fn() -> Vec<WsMessage> + Send + Sync`, called by the connection task after each successful connect, its output sent in order ahead of any queued `WsSender::send` backlog — the slot `config.subscriptions` occupies today.

Details worth settling rather than guessing:

- **Fallible?** `Fn() -> Result<Vec<WsMessage>>` would let a signature fail (clock skew, a missing credential). Treating that as a failed connection attempt and backing off seems right for a transient cause, but a permanently bad credential would then hammer the venue forever — aborting the run may be the better default.
- **Ordering vs. the backlog.** Whatever is chosen, it wants spelling out in the module docs, since it is the reason to use this over `WsSender`.
- **Keep `subscriptions`?** I would — most adapters never need more than the static list, and it reads better at the call site.
- **Redaction.** On-connect output can carry credentials, so it must stay out of any frame logging, consistent with the existing `redacted()` discipline around URLs.

Alternative shape considered: a keyed sticky set on `WsSender` (`send_sticky(key, msg)` / `retract_sticky(key)`), with the loop replaying the live set on each connect. That is nicer for the dynamic-subscription case and needs no lock in user code, but it does not address the auth case — sticky frames are values captured at send time, not rendered fresh — and it adds more API surface. The closure subsumes it.

### Backwards compatibility

Fully source- and behaviour-compatible, as far as I can see — this is additive in every direction:

- **`WsConfig` is `#[non_exhaustive]`**, so downstream cannot struct-literal it and a new field is not a breaking change. The builder setters are chainable and additive already.
- **`subscriptions` stays exactly as it is** — same public `Vec<WsMessage>` field, same `subscribe()` setter. Direct field reads and writes keep compiling. The new field sits alongside it: `on_connect: Option<Arc<dyn Fn() -> Vec<WsMessage> + Send + Sync>>`.
- **Default `None` reproduces today's loop byte for byte.** No existing config changes behaviour, so the semantics only move for callers who opt in.
- **`Clone` survives** — `Arc` clones. `WsConfig` derives no `PartialEq`/`Eq` and has no serde impls, so nothing else is disturbed.
- **One snag, and it is internal:** a closure is not `Debug`, so `#[derive(Clone, Debug)]` on `WsConfig` has to become a hand-written `Debug` (or the field gets a newtype whose `Debug` prints `<on_connect>`). `WsConfig: Debug` stays true for downstream — the impl changes, the API does not.
- **`Send + Sync` on the closure** keeps `WsConfig` as `Send + Sync` as it is today, which it needs since it is moved into the spawned connection task.

The change site is a single loop. Today:

```rust
for payload in &config.subscriptions {
    if write.send(to_tungstenite(payload.clone())).await.is_err() { … }
}
```

It becomes "the configured payloads, then the rendered ones", still ahead of the `Connected` yield — so the existing *"subscribe before announcing Connected, so a downstream that reacts to Connected cannot observe a socket that is open but not yet subscribed"* guarantee holds unchanged, and now covers the rendered frames too.

The one way to break compatibility here would be to *replace* `subscriptions` with the closure, or to order the rendered frames before the configured ones. Append, and nothing moves.

### Suggested scope for a first PR

The mechanical change is small and well isolated — one optional config field, one setter, one loop, plus the `Debug` impl — but two of the open questions above are judgement calls a newcomer should not have to make alone. Scoping a first PR to the infallible `Fn() -> Vec<WsMessage>` form, with the backlog ordering documented as-is, leaves a self-contained change; fallibility can follow separately if it is wanted. A maintainer pinning those two before someone starts would make this a cleanly newcomer-sized piece of work.

### What you are doing instead

Folding over the `WsStatus` stream and re-pushing the full subscription set through `WsSender` on each `Connected` transition, handling the double-send hazard by hand in the venue adapter. It works, but it is per-adapter code that is easy to get subtly wrong and awkward to test without forcing real reconnects.

### Would you like to implement it?

- [x] I would be interested in opening a PR for this

tunglambk:
## What this changes

Adds `WsConfig::on_connect`, an optional closure called once per successful connect whose frames are sent in order after `WsConfig::subscriptions` and before the queued `WsSender` backlog — still ahead of the `Connected` transition, so the existing subscribe-before-announce guarantee covers them too.

## Why

Closes #969.

`WsConfig::subscriptions` is frozen at wiring and re-sent after every connect. That covers a static feed, but not the two cases the issue describes: a subscription set the graph updates as instruments are listed and delisted, and an auth frame that has to be signed or timestamped per connect. Sent through `WsSender` instead, both are lost on the next reconnect, because the queue replays a message once, not on every connect.

## How it was verified

- [x] `cargo fmt --all`
- [ ] `cargo lint` and `cargo lint-all` — the workspace `cargo lint` (default features) passes. `lint-all` doesn't build here: it pulls in the native Aeron build, which wants `libclang`/`libbsd` this sandbox doesn't have. `cargo clippy -p wingfoil --all-targets --features ws,statistics -- -D warnings` passes.
- [ ] `cargo test -p wingfoil --all-features` — same native-Aeron blocker. `cargo test -p wingfoil` (default features) passes, and the `ws` suites do too: `--features ws --test ws_adapter` is 18 passed (5 new), `--features ws --lib adapters::ws` is 15 passed (2 new), and the new `on_connect` doctest passes under `--features ws --doc`.
- [x] New behaviour is covered by tests. The `ws` source is realtime-only, so these assert values and wire ordering rather than tick times, as the rest of this suite does. The new cases are: rendered fresh on every connect, sent ahead of the queued backlog, appended after the static list, and not rendered for a connect that never completed.

If you run the narrow `cargo clippy -p wingfoil --all-targets -- -D warnings`, it fails on `tests/op_completeness.rs` (`EwmaDecay` unused, `W` dead) because that file's uses of both sit behind `statistics`. It does the same on unmodified `main` — I reproduced it with my changes stashed — and `--workspace` resolves it through feature unification, which is why CI doesn't see it.

## Notes for the reviewer

- Infallible, per the issue's suggested scope. A `Result` return needs a ruling on whether a failure backs off or aborts the run first, so I left that for a follow-up.
- Ordering is append: static `subscriptions` first, rendered frames after. That's what leaves an existing config's frames byte for byte unchanged. The consequence for the auth case is documented on the setter — a sequence that has to precede the subscriptions goes wholly in the closure, with `subscriptions` left empty.
- `#[derive(Clone, Debug)]` becomes `#[derive(Clone)]` plus a hand-written `Debug`. `WsConfig: Debug` stays public; the closure prints as `<on_connect>` and is never called, since its output can carry credentials.
- On-connect output is never logged (there was no frame logging to begin with), and the module's `# Credentials` section now says so.
- Docs updated in the module header, `WsConfig`, `WsStatus::Connected`, `WsSender`, and `src/adapters/ws/CLAUDE.md`.

## Before you submit


0-jake-0:
Reviewed at `816ef17`. The only substantive change in the current contributor batch, and it's well-shaped: the ordering argument is the whole justification for the feature and the PR makes it explicit rather than leaving it to be discovered.

**What's right:**

- **Append, not prepend.** Rendered frames after `subscriptions` is what leaves an existing config's wire bytes unchanged, and `static_subscriptions_are_unchanged_without_on_connect` pins that rather than asserting it in prose. The consequence — auth-before-subscribe goes wholly in the closure with `subscriptions` empty — is documented at the setter where someone will hit it.
- **Inside the subscribe-before-`Connected` window.** Rendered payloads go out before the `Connected` yield, so the existing "a downstream reacting to `Connected` cannot see an unsubscribed socket" guarantee extends to them for free instead of growing an exception.
- **Ahead of the `WsSender` backlog**, which is the ordering `WsSender` genuinely cannot express, and it's the test that best explains why this isn't just "queue a frame yourself".
- **`on_connect_is_not_rendered_for_a_failed_connect`.** Once the documented use is a signed nonce, "per successful connect" stops being a wording choice and becomes a contract; asserting the counter is still `0` after backoff exhaustion is the right guard.
- **Hand-written `Debug` that prints `<on_connect>` without calling the closure**, with a test asserting `hunter2` never appears. Correct instinct — the module already treats the URL as secret-bearing and rendered output is strictly more sensitive.
- Infallible for now with the reason stated (a `Result` needs a retry-vs-abort ruling first) is the right scope cut, and `src/adapters/ws/CLAUDE.md` is updated in the same commit, which is what that file is for.

**One gap worth a decision before merge:** the Python binding. `crates/wingfoil-python/src/adapters/ws.rs` exposes `url` / `subscriptions` / `realtime`, so `on_connect` is Rust-only. `/bind-adapter` is explicit that where the Rust adapter grows a capability, "expose it too and say so in the docs; the superset objective runs through the bindings as well" — and a per-connect *signed auth frame* is precisely the thing a Python user of a venue feed will want. I don't think it blocks this PR: a `Py<PyAny>` callable invoked from the tokio connection task means taking the GIL on a task documented as "must not block", which deserves its own design pass. But the skill says don't drop it silently, so either a deviation note in the binding module's `//!` header or a follow-up issue linked from here, rather than nothing.

**Two minor things:**

- If `write.send` fails partway through `render()`'s output, `subscribed` goes false and the connection retries — the nonce for that attempt is burned even though the frames didn't all land. That's inherent to an infallible render and probably fine (a venue rejects a stale nonce and you reconnect), but it's the one case the doc's "once per successful connect" doesn't quite cover, and it's a sentence at the setter rather than a code change.
- Cosmetic: `Debug` prints `on_connect` last while it's the second struct field. Field order in `debug_struct` is usually worth matching the declaration.

Blocked on #970 for CI — `cargo audit` is red here on the RUSTSEC-2026-0285 `rustls` entry that's red on `main` too, unrelated to this diff. On the `lint-all` gap in your checklist: the native Aeron blocker is a known sandbox limitation (`scripts/setup-dev.sh --all-features` is what provisions it), and CI's `Lint (fmt & clippy)` is green here, which covers it. The `op_completeness.rs` failure you reproduced on unmodified `main` is diagnosed correctly — it's a narrow `-p wingfoil` resolve losing feature unification, not something your branch introduced.

---

PR #972, opened 2026-09-17
docs(core): document the feedback-send recipe and pin its +1 offset

### LINKED ISSUE #730: Document the feedback-send recipe that replaces legacy's `FeedbackSink::send`
Carried over from the July 2026 review, which deferred this as a contract question rather than a bug. **The contract question is now settled — see the ruling.** What is left is documentation plus a test, and it is a good first issue.

## The gap as filed

Legacy exposes `FeedbackSink::send(value, &mut GraphState)`, which schedules the paired *source* node via `GraphState::add_callback_for_node`. Wingfoil's `FeedbackSink` (`crates/wingfoil/src/interp.rs` — search `pub struct FeedbackSink`) has no public `send`, because sending means scheduling a **different** node than the caller, and wingfoil's op-facing `Ctx` is deliberately self-scheduling-only.

## Ruling: `Ctx` is not being widened

Two use cases were conflated under "send", and they have different answers:

1. **From outside the graph** — a producer thread pushing values in. Already covered, and not by feedback at all: `external` and `channel` sources exist for exactly this, and `ExternalSource::send` is public.

2. **From inside a running op** — the case the deferral was actually about. The wiring-time pair covers it. `GraphBuilder::feedback` returns `(source, sink)`; `stream.feedback(&sink)` forwards `stream` unchanged while pushing each value onto the sink's queue at `time + 1` and scheduling the paired source with direct kernel access. A *conditional* or *transformed* send is `map_filter` (or `filter`, or `map`) in front of that edge — composition, not a missing capability.

No use case has been found that this cannot express, so **`FeedbackSink::send` is not being added and `Ctx` stays narrow.** `Ctx` is the seam interpreted, compiled and nested all share; a widening admitting cross-node scheduling would have to hold across all three tiers, and that is not a price worth paying for an API that already has a spelling.

**If you find a case the wiring-time pair genuinely cannot express, that reopens this** — comment with the concrete wiring you tried and where it fell down. That is the bar. An ergonomics preference is not.

## The work

- [ ] **Update the `FeedbackSink` rustdoc.** It currently explains *why* there is no `send` — correctly, and at length — but never shows what to write instead, and it ends on "deferred until a concrete need arises". Replace that with the ruling and a short worked example along the lines of `let (loop_src, sink) = g.feedback(); ...; some_stream.map_filter(..).feedback(&sink)`.
- [ ] **Add a row to the "What is gone (and what replaced it)" table** in [`docs/migration.md`](../blob/main/docs/migration.md): `FeedbackSink::send(v, &mut state)` → the wiring-time spelling. That section is where a legacy user goes looking, and feedback is not mentioned anywhere in `migration.md` today.
- [ ] **Add a test** in `crates/wingfoil/tests/` demonstrating the equivalence: a conditional feedback send, where only values passing a predicate go round the loop. Assert exact values **and** tick times under `RunMode::HistoricalFrom(NanoTime::ZERO)`. The `+1` cycle offset the sink applies is the thing to pin — it is what a reader will get wrong, and it is the reason the graph stays acyclic.

## Notes

No engine changes. If you find yourself editing `Ctx`, stop and re-read the ruling — that is a different issue with a different argument to make.

Branch from `main`, PR base `main`.

tunglambk:
## What this changes

The `FeedbackSink` rustdoc now shows what to write instead of legacy's `send`, `migration.md` gets the legacy → wingfoil mapping, and a new test pins the conditional-send contract and its `+1` offset.

## Why

Closes #730. The ruling settled that `Ctx` isn't being widened and `FeedbackSink::send` isn't coming back; what was missing was the "if you're doing X, do Y" spelling and a test for it. The rustdoc explained the *why* correctly but ended on "deferred until a concrete need arises", and `migration.md` didn't mention feedback anywhere — which is exactly where a legacy user goes looking.

## How it was verified

- [x] `cargo fmt --all` — clean.
- [x] `cargo lint` (workspace clippy, default features) — clean.
- [ ] `cargo lint-all` — blocked in this sandbox: `etcd-client`'s build script can't find `protoc`, and all-features also wants the native Aeron/iceoryx2 toolchain. Same boundary #968 and #953 reported.
- [ ] `cargo test -p wingfoil --all-features` — ran the default-feature `cargo test -p wingfoil` instead: every target passes, including `trybuild`, all 15 doctests and the new `feedback_conditional_send_maps_and_filters`.
- [x] `cargo doc -p wingfoil --no-deps` — the new intra-doc links resolve (`Builder::feedback`/`external`/`channel`, `StreamOps::map_filter`, `ExternalSource::send`, `ChannelSender::send`).

The test wires ticker → `map_filter` (keep evens, ×10) → `feedback(&sink)` and asserts both halves: the pass-through ticks `[(100ns, 20), (300ns, 40)]`, and the fed-back source ticks `[(101ns, 20), (301ns, 40)]`. That `+1` is the part the issue says a reader will get wrong, so it's pinned rather than described.

## Notes for the reviewer

Docs plus a test, no engine changes — `Ctx` is untouched. I gave feedback its own mapping table under the entry rather than adding a row to the `Graph::export` table above it, because that table is scoped to the export replacement. Happy to fold it into the same table if you'd rather see it there.

## Before you submit


0-jake-0:
Reviewed at `c5f08e8`. This replaces a doc comment that explained a *deferral* with one that explains a *design*, which is the right correction — the old text read as "we haven't built `send` yet", and the actual position is that `send`'s two jobs each already have a spelling. Splitting it by where the value comes from (outside the graph → `external`/`channel`; inside a running op → the wiring-time pair) is the distinction that was missing.

Checked the claims against the source:

- **The `+1` is real.** `interp.rs:3750` is `let time = kernel.time() + 1;`, so "pushes at `time + 1` and schedules the source for that same instant" is accurate, and pinning it is worth doing — a reader's default assumption is one ticker period, not one nanosecond.
- **Qualifying it to `HistoricalFrom` is correct and easy to get wrong.** Under `RealTime` the kernel takes `NanoTime::now().max(time + 1)`, so the `+1` is a floor rather than the observed delta. Both the doctest and the migration table say `HistoricalFrom`; good.
- **The narrow-`Ctx` justification holds** — self-scheduling only is the seam the interpreted, compiled and nested tiers share, so "cannot schedule a different node" is a property of the design rather than a missing feature.
- `map_filter` in front of `.feedback(&sink)` really is the conditional-send replacement; `feedback` forwards the stream unchanged.

The migration.md table is the part that will actually get read — a legacy user greps for `FeedbackSink::send`, and it now lands somewhere that answers them in two rows.

One observation, not a change request: the new test and the doctest are the same graph with the same assertions, plus a `sent` accumulate in the test. That's defensible — the doctest is the thing docs.rs renders and the test is the thing that fails loudly on a regression — but if you'd rather not maintain two copies, the test is the one that earns its keep, since it also asserts what was *sent* at 100ns/300ns and so pins both ends of the `+1`.

Blocked only on #970 — `cargo audit` is red here on the RUSTSEC-2026-0285 `rustls` entry that's red on `main` too, and nothing here touches dependencies.

---

PR #973, opened 2026-09-17
docs(core): add map family doctests

maxi-maxima:
## What this changes

Adds runnable, deterministic doctests for `map`, `try_map`, and `map_filter`. The examples assert complete emitted sequences; the `try_map` example also proves that an error aborts only after earlier successful values have been retained.

## Why

Part of #935. These core value-transform methods currently expose only signatures on docs.rs, so the examples make ordinary mapping, quiet filtering, and fallible-abort behavior executable and visible at the API entry point.

## How it was verified

Linux, Rust 1.98.0, with the committed lockfile:

- [x] `cargo test --locked --doc -p wingfoil --all-features` — 26 passed, 22 intentionally ignored; the runnable count increased by three
- [x] `cargo test --locked -p wingfoil --all-features --lib` — 350 passed
- [x] `cargo test --locked -p wingfoil-derive` — 3 passed; 2 doctests intentionally ignored
- [x] New behavior is covered by exact value assertions, plus the `try_map` error context

## Notes for the reviewer

This is the claimed first slice of #935. It changes one documentation file and no runtime behavior. The remaining time and pairwise/enumerate methods stay available for separate contributions.

The PR's Cargo audit currently fails on the unchanged `rustls 0.23.43` entry and RUSTSEC-2026-0285. The separate dependency fix is #970, where the audit and full test/lint matrix pass; this branch intentionally does not mix in that dependency update.

## Before you submit


0-jake-0:
Reviewed at `14994f1`. Three doctests, each with a prose lead-in naming what to notice, contiguous `use` block, one-line `run(..).unwrap()` — this picked up the house pattern from the #956 round without being asked, which is the right way to run a series of doc PRs.

Traced the values rather than trusting the doctest run: a 10ns ticker with no scheduling op inserts no extra cycles, so counts are 1..N. `map` → `[2,4,6]` over `Cycles(3)`; `map_filter` → `[20,40]` over `Cycles(4)`; `try_map` errors at `n == 3` having emitted `[10,20]`. All correct.

**The `try_map` example is the strongest of the three** and worth calling out: it pins the part of the contract prose can only assert — that the abort preserves values already produced, *and* that the closure's message survives into the run error with context (`{error:#}`). That's two behaviours a signature change would silently break, now executable. Good instinct reaching for `unwrap_err()` here instead of demonstrating only the happy path.

Two nits, neither blocking:

- **`n % 2 == 0` in `map_filter`** — the tree's idiom is `n.is_multiple_of(2)` (`lib.rs:12`, `ops.rs:652`), and MSRV is 1.88 so it's available. Doctests aren't clippy-linted, so nothing will fail; it's purely that the snippet a reader copies out of docs.rs should look like the code in the repo.
- **Argument order in `assert_eq!`** — these are `(expected, actual)`, while #956's are `(actual, expected)`. Across two PRs from the same series in the same file, worth picking one. `(actual, expected)` matches more of `fluent.rs`.

Blocked only on #970 — `cargo audit` is red here on the RUSTSEC-2026-0285 `rustls` entry that's red on `main` too, and this diff touches no dependencies. Rebase after that lands; #956 edits the same trait a few hundred lines below, so whichever goes second wants a conflict check.

---

PR #976, opened 2026-09-18
fix(deps): disable rustls default provider

### LINKED ISSUE #974: deps: set `default-features = false` on rustls to drop the unused aws-lc-rs provider
## What

`crates/wingfoil/Cargo.toml` declares:

```toml
rustls = { version = "0.23.45", features = ["ring"], optional = true }
```

with no `default-features = false`. rustls 0.23's default features include the `aws_lc_rs` provider, so **both** providers are compiled even though every TLS path in the tree explicitly installs ring.

## Evidence

`aws-lc-rs` is genuinely in the enabled graph, not merely present in the lockfile — `cargo tree -i` only walks reachable edges:

```
$ cargo tree --locked -p wingfoil --features fix -i aws-lc-rs
aws-lc-rs v1.18.1
├── rustls v0.23.45
│   └── wingfoil v9.0.0 (crates/wingfoil)
└── rustls-webpki v0.103.15
    └── rustls v0.23.45 (*)
```

And every consumer picks ring by name:

| site | call |
|---|---|
| `src/adapters/fix.rs:2298` | `builder_with_provider(rustls::crypto::ring::default_provider().into())` |
| `src/adapters/web/server.rs:867` | `builder_with_provider(rustls::crypto::ring::default_provider().into())` |
| `src/adapters/ws.rs:931` | `rustls::crypto::ring::default_provider().install_default()` |

So `aws_lc_rs` is dead weight in all three of `fix`, `web-tls` and `ws-tls`.

## Why it matters

`aws-lc-rs` drags in `aws-lc-sys`, which is a **cmake + C build**. That's build-time cost and a native toolchain requirement for a provider nothing selects. It also showed up as churn in #970, which had to move `aws-lc-sys` 0.41 → 0.45 purely as a side effect of the rustls patch bump.

This is `CLAUDE.md` dependency rule 6: *"`default-features = false` unless you have checked what the defaults drag in. A dependency's default features are unmaskable by consumers."* Here the defaults are both unmaskable and unused.

## Suggested change

```toml
rustls = { version = "0.23.45", default-features = false, features = ["ring", "std", "logging", "tls12"], optional = true }
```

The exact feature list needs checking against what the adapters actually use — `std` at minimum, and `tls12` should stay on unless we intend to drop TLS 1.2 support, which would be a behaviour change rather than a build-time one. Don't guess this list; read rustls 0.23's `Cargo.toml` and confirm each one.

## Verification

- `cargo tree --locked -p wingfoil --features fix -i aws-lc-rs` should report nothing, and the same for `web-tls` and `ws-tls`.
- The existing TLS suites must still pass: `web_integration` / `web_adapter` under `web-tls-integration-test`, `ws_adapter` under `ws-tls`, `fix_adapter` under `fix`.
- `cargo audit` and `python3 scripts/check-dep-duplicates.py` stay clean.

## Release note

Per the dependency policy this is a **minor** bump when published: narrowing a dependency's default features can take something away from a consumer receiving it through feature unification with us.

## Provenance

Surfaced while reviewing #970 (RUSTSEC-2026-0285 / rustls 0.23.45), which is merged. Deliberately kept out of that PR to keep a security fix minimal.

maxi-maxima:
## What this changes

Disables rustls default features and explicitly preserves the `ring`, `std`, `logging`, and `tls12` features used by Wingfoil's TLS adapters.

## Why

Closes #974.

rustls defaults enable the unused `aws_lc_rs` provider alongside Wingfoil's explicitly selected ring provider. That adds an unnecessary `aws-lc-sys` C/CMake build to the `fix`, `web-tls`, and `ws-tls` feature graphs.

## How it was verified

- [x] `cargo tree --locked -p wingfoil --features <fix|web-tls|ws-tls> -i aws-lc-rs` finds no reachable package for each feature
- [x] `cargo test --locked -p wingfoil --features web-tls-integration-test --test web_adapter` (8 passed)
- [x] `cargo test --locked -p wingfoil --features web-tls-integration-test --test web_integration -- --test-threads=1` (19 passed, including TLS round trip)
- [x] `cargo test --locked -p wingfoil --features ws-tls --test ws_adapter` (17 passed)
- [x] `cargo test --locked -p wingfoil --features fix --test fix_adapter` (2 passed)
- [x] CI-equivalent non-service nextest suite (986 passed after excluding the known coarse-clock assertion below)
- [x] `cargo test --locked --doc -p wingfoil --all-features` (31 passed, 22 intentionally ignored)
- [x] `cargo test --locked -p wingfoil-derive` (3 passed)
- [x] `cargo audit` (no vulnerabilities; 11 repository-allowed warnings)
- [x] `python3 scripts/check-dep-duplicates.py`

The broad all-feature run's `stamp_each_all_reads_the_clock_once_per_stage_not_per_value` assertion still sees identical nanosecond timestamps in this Docker Desktop environment. The same coarse-clock failure is unrelated to dependency selection; the remaining 986 CI-equivalent non-service tests pass.

Upstream `cargo audit` and dependency review pass. The separate `pnpm audit` job currently fails on the unchanged `svelte > devalue` graph (`devalue <5.9.1`, GHSA-9rgm-9g3h-6x36); this Cargo-only PR does not change `pnpm-lock.yaml`.

## Notes for the reviewer

`Cargo.lock` is unchanged. Per the dependency policy, narrowing default features requires a minor release when published.

## Before you submit

- [x] **Allow edits by maintainers** is enabled


0-jake-0 (commented):
Reviewed and independently verified. The change is correct and matches the dependency policy (rule 6: `default-features = false` unless the defaults have been checked).

**Feature set is exactly right.** `rustls` 0.23.45 declares `default = ["aws_lc_rs", "logging", "prefer-post-quantum", "std", "tls12"]`. This keeps `logging`/`std`/`tls12`, swaps `aws_lc_rs` for `ring`, and correctly drops `prefer-post-quantum` — that feature is itself `= ["aws_lc_rs"]`, so it could not have been kept.

**Reproduced the before/after.** On `main`, `cargo tree --locked -p wingfoil --no-default-features --features <f> -i aws-lc-rs` resolves `aws-lc-rs v1.18.1` for all three of `fix`, `web-tls` and `ws-tls`. With this patch applied, all three report `package ID specification 'aws-lc-rs' did not match any packages`. Closes #974 as claimed.

**No runtime behaviour change.** Every TLS entry point already names the provider explicitly — `fix.rs:2298` and `web/server.rs:867` use `builder_with_provider(ring::default_provider())`, `ws.rs:931` does `ring::default_provider().install_default()`, and `web_integration.rs:1438` matches. Nothing relies on rustls' implicit default provider, which is the only thing this could have moved.

Two non-blocking notes:

1. **`--all-features` still pulls `aws-lc-rs`**, via `fluvio` → `futures-rustls` and `bollard` → `testcontainers`. Out of scope for #974 (which is about the `fix`/`web-tls`/`ws-tls` graphs) and not fixable from our manifest, but worth an issue if the goal is "no aws-lc-sys build anywhere".

2. **The `ws-tls` comment at `crates/wingfoil/Cargo.toml:424`** justifies the explicit ring install with "rustls 0.23 panics at connect time if a process links more than one provider and none was chosen". After this change only one provider is linked in that graph, so the stated reason no longer holds there. The explicit install is still right — a consumer can re-add `aws-lc-rs` through feature unification — but the comment now reads as describing something the manifest just fixed. A one-line amendment would keep it honest.

**CI:** `test.rust` and `test.python` are green; `security.audit` is red on the pre-existing `svelte > devalue` advisory (GHSA-9rgm-9g3h-6x36), unrelated to this diff. #980 fixes that, so this should go green once #980 lands and this branch picks up `main`.

Agreed on the version note — narrowing a dependency's default features is a minor per the policy, but nothing to do in this PR.

---

PR #980, opened 2026-09-20
fix(deps): require patched devalue

maxi-maxima:
## What this changes

Raises the pnpm override for `devalue` from 5.8.1 to the patched 5.x range and refreshes the lockfile to 5.9.4.

## Why

GitHub advisory [GHSA-9rgm-9g3h-6x36](https://github.com/advisories/GHSA-9rgm-9g3h-6x36) reports a denial-of-service issue in the version currently resolved through Svelte. The existing `pnpm audit --audit-level moderate` check now fails on 5.8.1.

Keeping the override at `>=5.9.2 <6` selects the first patched release without allowing an unrelated major upgrade. No other dependency is changed.

## How it was verified

- `pnpm install --frozen-lockfile`
- `vitest run tests/tracing.test.ts` — 15 passed
- `pnpm audit --audit-level=moderate --registry=https://registry.npmjs.org/` — no known vulnerabilities
- `pnpm list devalue --depth 4` — Svelte resolves `devalue` 5.9.4
- `git diff --check origin/main...HEAD`

## Notes for the reviewer

The clean checkout does not contain the CI-generated `js/src/wasm/wingfoil_wasm.js`, so the four WASM-dependent JS suites are left to the normal CI build. This dependency-only change is covered locally by the frozen install, focused non-WASM tests, and the same audit command used by `security-audit.yml`.

## Before you submit

- [x] **Allow edits by maintainers** is enabled


0-jake-0 (commented):
Reviewed. The change does what it says and CI is fully green (`test.rust`, `test.python`, `test.integration.web`, `security.audit` all passing on `6ab4009`). The lockfile diff is minimal and consistent — only the `devalue` resolution and svelte's edge to it move.

One correction to the description, and one thing worth deciding before merge:

1. **The advisory is patched at 5.9.1, not 5.9.2.** Queried the npm advisory endpoint directly: GHSA-9rgm-9g3h-6x36 reports `vulnerable_versions: "<5.9.1"`. So `>=5.9.2` is safe but is not "the first patched release" — `>=5.9.1` is. No behavioural difference here (the lock resolves 5.9.4 either way), just don't want the floor's rationale on record as something it isn't.

2. **The `<6` cap diverges from every other entry in the override block.** `svelte`, `esbuild`, `@babel/core`, `postcss`, `seroval` and `seroval-plugins` are all open-ended `>=`. A pnpm override *forces* the resolution, so the cap means that when a future `svelte` requires `devalue@^6`, the frozen install breaks rather than resolving — and the whole point of these entries is to be a floor that stops mattering once upstream catches up. The stated worry (an unrelated major upgrade) isn't really live: the override can only pick from what the graph asks for. I'd make it `"devalue": ">=5.9.1"` and keep the block uniform, but this is a judgement call and either form is safe today.

**Merge order:** this one first. #976's `security.audit` is currently red on exactly this advisory, so landing #980 clears that leg for #976 too.

---

